Issue #765, opened 2014-04-14
Upgrade of crouton bricks Chromebook due to dev_boot_signed_only=1


wchao:
I have an Acer C720 running ChromeOS 34.0.1847.118. I installed crouton on my Chromebook and it worked great. Because I wanted to use virtualization, I chose to add the disablevmx=off option to the kernel boot command line, which required that I set dev_boot_signed_only=0. All was well. Then I upgraded crouton, which set dev_boot_signed_only=1 again. When I rebooted, my Chromebook was unable to boot the active image, since it requires a kernel signed by Google.

I would like to suggest that when performing updates, crouton first check if the currently active image is signed by Google. If not, it should not set dev_boot_signed_only to 1. Otherwise people will run into the same problem as me. That should help for the future.

Does anyone know how I can either set dev_boot_signed_only=0 or boot using the previous boot image in /dev/sda2 or /dev/sda4? I believe the other boot partition has a boot image that IS signed by Google, so if I can get in, then I can run crossystem dev_boot_signed_only=0 and all should be well. It's just getting in that's the hard part. I'd prefer not to reformat the drive since I spent a couple of days getting the chroot Debian and the Windows VM under VirtualBox just so, and I'm not looking forward to repeating all that work.

rrolsbe:
wchao commented on Apr 16, 2014

I managed to recover the data. I had another Chromebook and realized that if I took the SSD from the other Chromebook, I would be able to boot. Then I set dev_boot_signed_only=0, and installed the original SSD. Problem solved.

I am looking for some clarification on exactly what wchao did and why it worked.  I assume the drive from the other Chromebook must have also been in Developer Mode (or  the Developer Mode status is stored in the r/w firmware) and had the kernel boot bits stored on the drive were signed by Google, else the dev_boot_signed_only=0 command could not have been executed?  Also, The dev_boot_signed_only status must be stored in the r/w firmware area of Chromebooks, and NOT anywhere on the hard drive?

Thanks Much In Advance for any clarification!

Regards, Ron

PS - I had to perform a Recovery from USB after performing a crouton update. After the update and execution of the "sudo startxfce" command, it promptly crashed and rebooted thereafter.  The dev_boot_signed_only must of been set to 1 by the update.

wchao:
[...]

PR #785, opened 2014-04-20
Test if kernels are signed before setting dev_boot_signed_only=1

drinkcat:
This has been discussed extensively in #765.
- The verification has been modified a little: We check all bootable kernels (instead of hardcoding `KERN-A` and `KERN-B`). This should play better with Chrubuntu.
- We start the kernel verification process in parallel with bootstrapping.
- After boostrapping is done, we check the status, and enable `dev_boot_signed_only` if required.
- We inform the user of the modification in a trap, so it's displayed in all cases.
- Warnings regarding USB/legacy booting are only displayed on success (it's less critical to display them).

I also provide a test, and the function is put in `installer/functions`, before `###`, so it's not included in `host|chroot-bin` files.

### DIFF
--- a/installer/functions
+++ b/installer/functions
@@ -35,60 +35,6 @@ installscript() {
     1' "$src" | sed -e "${3:-;}" > "$dst"
 }
 
-# Verify if kernels are signed: We use this to decide if we should set
-# dev_boot_signed_only=1. We are conservative, and only return 0 if all bootable
-# kernels are signed with factory keys.
-# We do not take non-bootable kernels into account, as those will be ignored
-# during the boot process (e.g. KERN-C in a typical setup), or, if no
-# bootable kernel is present, the Chromebook will not boot no matter the value
-# of dev_boot_signed_only.
-# This is only used in main.sh and a test script, so it does not need
-# to be inserted in other scripts.
-verifykernelsig() {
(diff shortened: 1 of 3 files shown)


divx118:
Any news on when or if this will be merged? When just changing the kernel boot command line we also need to set dev_boot_signed_only=0, however we don't really need to set other boot options (usb and/or legacy).  
Current workaround for me seems to set other boot options, to at least prevent  crouton installer setting dev_boot_signed_only=1.

dnschneid:
Haven't heard anyone hitting this in a long time.  @divx118, do you still see value in such a patch?  Are we not seeing any bug reports just because your scripts/instructions anticipate and avoid the issue?

divx118:
@dnschneid I indeed avoid the issue in my script https://github.com/divx118/crouton-packages/blob/master/change-kernel-flags#L221 So it isn't really an urgent problem for me. IIRC It takes some time to check if the kernels signed or not.
IMHO in the long run this should be fixed somehow, because it can leave you with an non bootable system when you just set the flag and don't check the kernels. People are not obligated to use my script, they may as well change kernel flags and don't notice that crouton will  set dev_boot_signed_only.

dnschneid:
Going to close this for now.  We can re-open if this somehow resurges as an issue for people.

PR #832, opened 2014-05-22
remove cinnamon target

### LINKED ISSUE #812: Can't install Cinnamon on any distro
when setting up the chroot after around a hour of it running i get this:
E: Unable to locate package nemo
The chroot setup script may be broken. Your chroot is not fully configured.
Removing the chroot setup script. You may want to update your chroot again
UID 1000 not found in "linux distro name"
Unmounting /user/local/chroots/...
same error occurs on every distro/DE combination I'v tried. deleted and re-downloaded your script, any ideas how i can fix this?

picostove:
RIP Cinnamon target, 2013-2014. Perhaps you'll make a comeback one day, when we can get decent packages for you.

Closes #812


SaikoHaven:
Well, guess I'll bring cinnamon back. I have ways to get MATE etc. on crouton, with sudo startmate working as well from crosh, so I'll give it a shot and make a PR if all goes well. Though it will likely only work as far back as 16.04 and won't have ARM support right away, but I like cinnamon and i have nothing better to do

DennisLfromGA:
Sounds good, can't wait for cinnamon and/or mate.
I dual-boot Linux Mint on my Chromebooks and love both of those desktops.

PR #924, opened 2014-07-11
Downgrade jessie and sid mesa libraries

drinkcat:
This is trickier than precise, as `libegl1-mesa` depends on `libwayland0=0.85.0-2`. We manually fetch and install the packages, then hack the dpkg database to change the dependency (we could modify the `.deb` files but that's a bit more complicated and slower as we need to repack the file). Not pretty, but I couldn't find any better way.

To avoid fetching the `.deb` files on every update, I added a `list_needs_update` function in `ubuntu/prepare`. We could move it to `prepare.sh` if there is a need to use it on other releases.

I'd appreciate if somebody could check if it actually works on affected hardware (@DennisLfromGA @peter-gulka @feeddageek: anyone feeling adventurous?), at least kde/gnome targets on `jessie`. `sid` as a bonus.

Tested in `2014-07-11_10-20-09_drinkcat_chroagh_x11-test.tmp-lts-trusty-fix-jessie-sid_35`, matrix here: http://drinkcat.github.io/chroagh/2014-07-11_10-20-09_drinkcat_chroagh_x11-test.tmp-lts-trusty-fix-jessie-sid_35.html .

### DIFF
--- a/installer/ubuntu/prepare
+++ b/installer/ubuntu/prepare
@@ -166,6 +166,29 @@ install_mirror_package() {
     fi
 }
 
+
+# list_needs_update: For the specified package names, prints out any
+# package that needs an update (packages that do not exist in the
+# database are also printed).
+list_needs_update() {
+    local pkg
+    for pkg in "$@"; do
+        local v v1="" v2="(not found)"
+        for v in `apt-cache policy "$pkg" | \
+                  awk '$1~/(Installed|Candidate):/{print $2}'`; do
+            if [ -z "$v1" ]; then
+                v1="$v"
+            else
+                v2="$v"
+                break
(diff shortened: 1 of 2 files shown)


dnschneid:
Unlikely to happen due to the invasiveness and eventual 3.8 move.

DennisLfromGA:
> eventual 3.8 move.

Can't wait for that to happen ( on my falco ).

divx118:
@DennisLfromGA falco is on 3.8

DennisLfromGA:
@divx118 - Yup, sorry, I got confused with my 'parrot' that I just revived from a TPM error. I'm not used to using it.... :/

PR #1049, opened 2014-09-17
Audio auto update

drinkcat:
So I'm trying to find ways to auto-update the CRAS plugin.

One way we discussed, a while ago, was to download the latest version of crouton, and update with `-U audio`. I'm not too comfortable with that as it will break chroagh (or other fork/branch), and it requires that the checking code is placed in `enter-chroot`, which, really, shouldn't be the place for that kind of checks.

The approach here is to output the generated target files to the chroot into separate files in `/usr/local/crouton/`, and source them from `prepare.sh`. Then, we can update the audio target from inside the chroot by sourcing these files (see `cras-audio-update`).
Finally `enter-chroot` will just need to call `cras-audio-update` just after `/etc/rc.local` (or another better way, to be defined).

We'll probably also want to split `core`/`post-common` to contain only what is really needed by `audio`.

### DIFF
--- a/chroot-bin/cras-audio-update
+++ b/chroot-bin/cras-audio-update
@@ -12,8 +12,8 @@ QUIET=
 
 USAGE="$APPLICATION [-a] [-f]
 
-Update CRAS ALSA plugin in the chroot, so that it is in sync with the version
-provided by Chromium OS.
+Update Chromium OS audio plugin in the chroot, so that it is in sync with the
+version provided by Chromium OS.
 
 Options:
     -a          Ask before updating.
@@ -55,16 +55,16 @@ Chromium OS does not provide CRAS version information. Not updating."
 
     if [ "$hostver" = "$localver" ]; then
         if [ -z "$QUIET" ]; then
-            echo "CRAS ALSA plugin is up to date."
+            echo "Chromium OS audio plugin is up to date."
         fi
         exit 0
     else
         if [ -z "$ASK" ]; then
             echo "\
-CRAS ALSA plugin is out of date." 1>&2
+Chromium OS audio plugin is out of date." 1>&2
         else
             echo -n "\
-CRAS ALSA plugin is out of date, would you like to update it? [Y/n] " 1>&2
+Chromium OS audio plugin is out of date. Would you like to update it? [Y/n] " 1>&2
             response='n'
             if [ -t 0 ]; then
                 read -r response
@@ -77,7 +77,7 @@ CRAS ALSA plugin is out of date, would you like to update it? [Y/n] " 1>&2
     fi
 fi
 
-echo "Updating CRAS ALSA plugin..." 1>&2
+echo "Updating Chromium OS audio plugin..." 1>&2
 
 # Run the installer
 ret=
@@ -92,7 +92,7 @@ PREPAREDIR="/etc/crouton/prepare"
 
 if [ -n "$ret" ]; then
     echo "\
-Failed to update CRAS.
+Failed to update Chromium OS audio plugin.
 Please download a fresh crouton installer and update your chroot." 1>&2
     exit "$ret"
 fi
@@ -100,15 +100,15 @@ fi
 echo "Testing audio..." 1>&2
 cras_test_client --dump_server_info > /dev/null || ret=$?
 timeout 2 sh -ec 'dd if=/dev/zero bs=48000 count=1 2>/dev/null \
-                     | aplay -fdat -Dcras' || ret=$?
+                     | aplay -q -fdat -Dcras' || ret=$?
 
 if [ -n "$ret" ]; then
     echo "\
-CRAS was updated successfully, but audio could not be played.
-Please download a fresh crouton installer and update your chroot." 1>&2
+Chromium OS audio plugin was updated successfully, but audio could not be
+played. Please download a fresh crouton installer and update your chroot." 1>&2
     exit "$ret"
 fi
 
-echo "CRAS ALSA plugin was updated successfully!" 1>&2
+echo "Chromium OS audio plugin was updated successfully!" 1>&2
 
 exit 0
--- a/chroot-bin/cras-audio-update
+++ b/chroot-bin/cras-audio-update
@@ -1,0 +1,113 @@
+#!/bin/sh -e
+# Copyright (c) 2014 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+set -e
+
+APPLICATION="${0##*/}"
+ASK=
+FORCE=
+QUIET=
+
+USAGE="$APPLICATION [-a] [-f]
+
+Update CRAS ALSA plugin in the chroot, so that it is in sync with the version
+provided by Chromium OS.
+
+Options:
+    -a          Ask before updating.
+    -f          Force updating, even if the plugin is up to date.
+    -q          If up to date, don't print anything."
+
+# We need to run as root
+if [ "$USER" != root -a "$UID" != 0 ]; then
+    echo "$0 must be run as root." 1>&2
+    exit 2
+fi
+
+while getopts 'afq' f; do
+    case "$f" in
+    a) ASK='y';;
+    f) FORCE='y';;
+    q) QUIET='y';;
+    \?) echo "$USAGE" 1>&2; exit 2;;
+    esac
+done
+
+if [ -z "$FORCE" ]; then
+    hostver="`cat /var/host/cras-version`"
+    localver="`cras_test_client --version 2>/dev/null || true`"
+    # Strip out anything before the commit id
+    hostver="${hostver##*-}"
+    localver="${localver##*-}"
+
+    if [ -z "$hostver" ]; then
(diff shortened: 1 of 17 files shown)


drinkcat:
Rebased on #1350. Behaviour is already ok: no auto-update if `/var/host/cras-version` is empty, when `cras_test_client --version` in Chromium OS fails (either `cras_test_client` or `/var/run/cras` does not exist)

chroot-bin/cras-audio-update
```diff
@@ -0,0 +1,114 @@
...
+
+    if [ "$hostver" = "$localver" ]; then
+        if [ -z "$QUIET" ]; then
+            echo "CRAS ALSA plugin is up to date."
+        fi
+        exit 0
+    else
+        if [ -z "$ASK" ]; then
+            echo "\
+CRAS ALSA plugin is out of date." 1>&2
```
dnschneid: Can we say "Chromium OS audio plugin" instead?  All users will be seeing these messages and will see "CRAS ALSA" as complete gibberish.

chroot-bin/cras-audio-update
```diff
@@ -0,0 +1,114 @@
...
+        else
+            echo -n "\
+CRAS ALSA plugin is out of date, would you like to update it? [Y/n] " 1>&2
```
dnschneid: , -> .

chroot-bin/cras-audio-update
```diff
@@ -0,0 +1,114 @@
...
+    echo "\
+Failed to update CRAS.
+Please download a fresh crouton installer and update your chroot." 1>&2
+    exit "$ret"
+fi
+
+echo "Testing audio..." 1>&2
+cras_test_client --dump_server_info > /dev/null || ret=$?
+timeout 2 sh -ec 'dd if=/dev/zero bs=48000 count=1 2>/dev/null \
+                     | aplay -fdat -Dcras' || ret=$?
```
dnschneid: aplay -q

chroot-bin/croutonxinitrc-wrapper
```diff
@@ -131,6 +131,9 @@ fi
 if [ "$XMETHOD" = 'xiwi' ]; then
     # The extension sends evdev key codes: fix the keyboard mapping rules
     setxkbmap -rules evdev
+    # Reapply xkb map: This fixes autorepeat mask in "xset q"
+    xkbcomp "$DISPLAY" - | xkbcomp - "$DISPLAY" 2>/dev/null
```
dnschneid: Weird, even though I merged this, the diff doesn't update?

host-bin/enter-chroot
```diff
@@ -656,12 +661,22 @@ if [ -n "$NOLOGIN" ]; then
...
-        chrootcmd 'exec /etc/rc.local >/dev/null 2>/dev/null </dev/null' \
-            || ret=$?
+    if [ -n "$firstrun" ]; then
+        # Run rc.local
+        if [ -x "$CHROOT/etc/rc.local" ]; then
+            chrootcmd 'exec /etc/rc.local >/dev/null 2>/dev/null </dev/null' \
+                || ret=$?
+            if [ ! "$ret" = 0 ]; then
+                echo "WARNING: /etc/rc.local failed with code $ret" 1>&2
+                ret=0
```
dnschneid: sleep 5

host-bin/enter-chroot
```diff
@@ -656,12 +661,22 @@ if [ -n "$NOLOGIN" ]; then
...
+            fi
+        fi
+
+        # Run cras-audio-update
+        chrootcmd 'hash cras-audio-update 2>/dev/null && \
```
dnschneid: This should happen before rc.local.

installer/main.sh
```diff
@@ -652,9 +658,13 @@ if [ -z "$RESTOREBIN" ] && [ -z "$RESTORE" -o -n "$UPDATE" ]; then
     VAREXPAND="${VAREXPAND}s #VERSION# ${VERSION:-"git"} ;"
     VAREXPAND="${VAREXPAND}s #USERNAME# $CROUTON_USERNAME ;"
     VAREXPAND="${VAREXPAND}s/#SETOPTIONS#/$SETOPTIONS/;"
-    installscript "$INSTALLERDIR/prepare.sh" "$PREPARE" "$VAREXPAND"
+    installscript "$INSTALLERDIR/prepare.sh" \
+                  "$CHROOT$PREPAREDIR/prepare.sh" "$VAREXPAND"
     # Append the distro-specific prepare.sh
-    cat "$DISTRODIR/prepare" >> "$PREPARE"
+    cat "$DISTRODIR/prepare" > "$CHROOT$PREPAREDIR/distroprepare"
```
dnschneid: cp?

dnschneid:
I'm not a big fan of the extra targets, nor putting distro-specific prepare stuff into a target.  `core` is actually designed to be "kind of optional" (post-common and the prepare* stuff are the only mandatory things).

Maybe add a "minimal" script to each distro that is sourced after the prepare, and have an environment variable CROUTON_MINIMAL or CROUTON_CONTEXT=update or CROUTON_UPDATE or something that skips stuff in other targets?

test/tests/30-audio
```diff
@@ -45,14 +45,39 @@ if [ "`uname -m`" = "x86_64" ]; then
...
+# At this stage, enter-chroot would try to run the auto-update, but
+# we are not in a tty, so it doesn't.
+fails host enter-chroot -n "$release" cras_test_client --dump_server_info
+
+# Manually update
+host enter-chroot -n "$release" -u 0 cras-audio-update
+
+if [ -z "`cras_test_client --version`" ]; then
+    # This version of Chromium OS does not support audio auto-update:
+    # force update.
```
dnschneid: should confirm that audio still doesn't work.

chroot-bin/cras-audio-update
drinkcat: Makes sense. Fixed (everywhere in the script).

chroot-bin/cras-audio-update
drinkcat: fixed

chroot-bin/cras-audio-update
drinkcat: Sure.

chroot-bin/croutonxinitrc-wrapper
drinkcat: Yeah, maybe I need to rebase...

host-bin/enter-chroot
drinkcat: Good catch...

host-bin/enter-chroot
drinkcat: Sure. And it shouldn't fail if `cras-audio-update` does not exist... Updated.

installer/main.sh
drinkcat: Sure.

test/tests/30-audio
drinkcat: Fixed.

drinkcat:
> Maybe add a "minimal" script to each distro that is sourced after the prepare

Agreed. Need to thing of a better name though (I would not source that `minimal` script from auto update). Maybe `secondstage` or `setup` (most of what it does is running bootstrap 2nd stage, the rest (2 lines) can be moved into `core.minimal` (or `core`).

> core is actually designed to be "kind of optional"

Well, that's the reason why I split it (only setting up the proxy is really necessary for other targets).

> and have an environment variable CROUTON_MINIMAL or CROUTON_CONTEXT=update or CROUTON_UPDATE or something that skips stuff in other targets?

Doing that would mean indenting most of `core` and `post-common`. I'm not sure if that's a great idea in terms of readability... And I'm not a big fan of this kind of "control-flow" variables (I think it's the main reason why `installler/main.sh` is so hard to read and modify)

dnschneid:
OK, that's reasonable.  How about moving things to secondstage and then we can see what's left to deal with?

drinkcat:
Moved functions to an `installer/<distro>/functions` file, and left `installer/<distro>/prepare` with the "active" code (similar to `secondstage` described above).

That way `cras-audio-update` can source only the functions.

targets/core.minimal
```diff
@@ -0,0 +1,46 @@
...
+REQUIRES='core.minimal'
+DESCRIPTION=''
+CHROOTBIN=''
+. "${TARGETSDIR:="$PWD"}/common"
+
+### Append to prepare.sh:
+
+if [ "${DISTROAKA:-"$DISTRO"}" = 'debian' ]; then
+    # Run dpkg in non-interactive mode
+    export DEBIAN_FRONTEND=noninteractive
```
dnschneid: This should probably be in `functions`, since it just impacts the implementation of `install` and whatnot.

dnschneid:
Nice!  I like the functions file.  Still don't like the .minimal targets though.  How about adding a second `###` line?  `common` can check the number of `###` lines (`grep -c`) and generate the .minimal file on demand.

```
DESCRIPTION=blah
. ./common
### Environment setup for dependent targets
x=/etc/crouton/x
### Chroot preparation steps
touch "$x"
```

dnschneid:
Audio seems to be very stable on the Chromium side, so I don't think this patch is going to get the further development it needs.

PR #1144, opened 2014-10-23
Crouton in a tab

drinkcat:
This PR adds support for crouton in a window (kiwi). A dummy Xorg server is started on the crouton side, and a NaCl module communicates with the dummy Xorg server (input events, cursors, frame requests) via Websocket. The framebuffer data is not transfered by Websocket, for performance reasons. Instead, data is written directly into shared memory between NaCl and Chrome.

Known issues (fix needed for merge):
- Close the window, then extension cannot reopen it (https://github.com/dnschneid/crouton/pull/1144#issuecomment-62241254)
- From Dennis's video (http://goo.gl/R5V7GE):
  - Pressing the page flipper resizes the chroot, leading to strange artifacts on LXDE.

Known issues (can be merged without a fix):
- Switching away from the chroot using `croutoncycle` does not restore the window to maximum size
- The extension does not start with focus (you need to click, this might be a Chromium OS bug)
- IME events (place holder is there, needs implementation)
- Caps lock is not synchronized (https://github.com/dnschneid/crouton/pull/1144#discussion_r19914500)
- Waiting for extension to connect before starting WM is potentially racy (the setres 1024 768 trick to provide xfce with a reasonable DPI fails as the extension starts too soon)... (actually, this is more of a xorg+dummy bug: we could get rid of that if dummy supported xrandr 1.2)
- On stable channel (`6158.70.0`), the extension/Chrome sometimes crashes on disconnect (fairly frequently): clearly a Chrome bug. Never seen that problem on canary.

Fixed:
- Pressing on "Unknown" hangs the system? => Do not allow switching to "X11" windows from the extension.
  - Pressing Ctrl+Shift+V is broken (sends SuperL?)
  - Clipboard handling might be broken (`aura`->`cros` change?)
  - Touch events (basic, single point only)
  - Window title should correspond to chroot name
  - A leftover FIXME in the code, we should force croutoncycle instead of removing files
  - Errors when loading NaCl module are not reported
  - Limit number of resolution changes when resizing (it's slow because we send far too many "R" requests)
  - Move the websocket functions into a common file
  - Rename Nacl module/target to Kiwi/xiwi.
  - "Error target STRING not available" sometimes appears during switch (harmless, probably due to old chroots running)
  - Logging: https://github.com/dnschneid/crouton/pull/1144#discussion_r19862772
  - Super_L: https://github.com/dnschneid/crouton/pull/1144#discussion_r19866490
  - Use patched xorg+dummy.
  - Add test for xiwi target.

### DIFF
--- a/chroot-bin/setres
+++ b/chroot-bin/setres
@@ -57,22 +57,32 @@ replacemode() {
 
 # Try to change to arbitrary resolution
 mhz="$((r*x*y/1000000))"
-name="Crouton in a tab"
+name="kiwi_${x}x${y}_${r}"
 
-# Tweak the temp mode so xrandr knows the 2 modes are different, and
-# switches away from the temp mode.
-replacemode "$o" "$name (tmp)" $mhz $x $x $x $((x+1)) $y $y $y $y
+# Try to switch mode, if it already exists.
+if xrandr --output "$o" --mode "$name" 2>/dev/null; then
+    echo "${x}x${y}_${r}"
+    exit 0
+fi
 
-# This fails on non-patched xorg-dummy
-if xrandr --output "$o" --mode "$name (tmp)"; then
-    replacemode "$o" "$name" $mhz $x $x $x $x $y $y $y $y
-    xrandr --output "$o" --mode "$name"
-    replacemode "$o" "$name (tmp)"
+# Add the new mode
+xrandr --newmode "$name" $mhz $x $x $x $x $y $y $y $y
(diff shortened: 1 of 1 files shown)


kaendfinger:
This thread should win the award for longest conversation on GitHub

dnschneid:
There are some viral threads around GitHub that beat this by a longshot...

kaendfinger:
Aww, I was hoping.... :P

PR #1281, opened 2014-12-29
Several small fixes.

dnschneid:
Opening a PR for a bunch of small fixes so that they get reviewed.


build/wrapper.sh
```diff
@@ -62,6 +62,34 @@ if [ -z "$TRAP" ]; then
     exit
 fi
 
+# See if we want to just run a script from the bundle
+if [ "$1" = '-X' -a -z "$2" ]; then
```
drinkcat: Put all this under the same `if [ "$1" = '-X' ];` block. i.e. move `[ -z "$2"]` to the if block below, you could even merge `-z "$2"` and `-f "$script"` and output something that makes sense in both cases.

chroot-bin/crouton-noroot
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh -e
+# Copyright (c) 2014 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# Wrapper for scripts that shouldn't be launched as root.
+# Symlink to automatically launch that application.
```
drinkcat: the application with the same filename under `/usr/bin/`

host-bin/edit-chroot
```diff
@@ -60,29 +62,32 @@ Options:
...
+    d) DELETE='y'; NEEDS_UNMOUNT='y';;
+    e) ENCRYPT='y'; NEEDS_UNMOUNT='y';;
+    f) TARBALL="$getopts_arg";;
+    k) KEYFILE="$getopts_arg";;
     l) LISTDETAILS=$(($LISTDETAILS+1));;
-    m) MOVE="$OPTARG";;
-    r) RESTORE=$(($RESTORE+1));;
+    m) MOVE="$getopts_arg"; NEEDS_UNMOUNT='y';;
+    r) RESTORE=$(($RESTORE+1)); NEEDS_UNMOUNT='y';;
+    s) SPLIT="$OPTARG";;
```
drinkcat: $getopts_arg

host-bin/edit-chroot
```diff
@@ -292,8 +317,11 @@ for NAME in "$@"; do
             error 2 "$CHROOT already exists! Specify a second -r to overwrite it (dangerous)."
         elif [ -n "$RESTORE" ]; then
             EXISTS='y'
-        else
-            sh -e "$BINDIR/unmount-chroot" $YESPARAM -c "$CHROOTS" -- "$NAME"
+        elif ! sh -e "$BINDIR/unmount-chroot" $YESPARAM \
+                    -c "$CHROOTS" -- "$NAME"; then
+            if [ -n "$NEEDS_UNMOUNT" ]; then
+                exit 1
+            fi
```
drinkcat: Why are we trying to unmount when we don't need to unmount?

host-bin/edit-chroot
```diff
@@ -335,15 +363,34 @@ for NAME in "$@"; do
                 dest="$dest.gz"
             fi
         fi
+        # If we're writing to a fat32 filesystem, split the file at 4GB chunks
+        if ! df -T "`getmountpoint "$dest"`" | awk '$2~"^v?fat"{exit 1}'; then
+            SPLIT="${SPLIT:-4GB}"
+        fi
+        if [ -n "$SPLIT" ]; then
+            tardest="`mktemp -u --tmpdir=/tmp 'crouton-backup.XXX'`"
```
drinkcat: unsafe (okay, the likelyhood of something bad happening is low, but, still). Better create a temp directory and put the fifo in it.

host-bin/edit-chroot
```diff
@@ -335,15 +363,34 @@ for NAME in "$@"; do
...
+        addtrap "echo 'Deleting partial archive.' 1>&2; \
+                 kill '$splitpid' 2>/dev/null; rm -f '$dest' '$dest.part'*"
         ret=0
         spinner 1 tar --checkpoint=100 --checkpoint-action=exec=echo \
             --one-file-system -V "crouton:backup.${date%-*}${date#*-}-$NAME" \
-            -caf "$dest" -C "$CHROOTS" "$NAME" || ret=$?
-        if [ "$ret" != 0 ]; then
+            -caf "$tardest" -C "$CHROOTS" "$NAME" || ret="$?"
+        if [ -n "$SPLIT" ]; then
+            wait "$splitpid" || ret="$?$ret"
```
drinkcat: This makes me worry about strange exit values and overflows. I'd rather simply override `ret="$?"` (it's only set on error anyway). Or another ret variable.

host-bin/edit-chroot
```diff
@@ -335,15 +363,34 @@ for NAME in "$@"; do
                 dest="$dest.gz"
             fi
         fi
+        # If we're writing to a fat32 filesystem, split the file at 4GB chunks
+        if ! df -T "`getmountpoint "$dest"`" | awk '$2~"^v?fat"{exit 1}'; then
+            SPLIT="${SPLIT:-4GB}"
```
drinkcat: Please add a comment that `4GB = 4 * 10^9` (and not `4 * 1024^3` (4G), since the fat32 limit is 4G-1 bytes). Actually, I'd almost prefer `4000000000` as GB is not POSIX.

host-bin/edit-chroot
```diff
@@ -360,7 +407,7 @@ for NAME in "$@"; do
             # Search for the alphabetically last tarball with src.
             # Dated tarballs take precedence over undated tarballs.
             for file in "$file."* "$file-"*; do
-                if [ ! -f "$file" ]; then
+                if [ ! -f "$file" -o "${file%part[0-9][0-9]}" != "$file" ]; then
```
drinkcat: What if there are more than 100 parts? (400 GB is not unthinkable...)

host-bin/edit-chroot
```diff
@@ -385,8 +432,22 @@ for NAME in "$@"; do
         fi
         echo -n "  Restoring $src to $CHROOT..." 1>&2
         mkdir -p "$CHROOT"
+        if [ -f "$src.part01" ]; then
+            tarsrc="`mktemp -u --tmpdir=/tmp 'crouton-restore.XXX'`"
```
drinkcat: Same as above. Use `-d` and create the fifo in that directory.

host-bin/edit-chroot
```diff
@@ -385,8 +432,22 @@ for NAME in "$@"; do
...
+            mkfifo -m 600 "$tarsrc"
+            # Pipe through dd to eat the EOFs
```
drinkcat: I don't understand this.

host-bin/edit-chroot
```diff
@@ -385,8 +432,22 @@ for NAME in "$@"; do
...
+            cat "$src" "$src.part"* | dd status=none >> "$tarsrc" &
+            catpid="$!"
+            addtrap "rm -f '$tarsrc'; kill '$catpid' 2>/dev/null"
+        else
+            tarsrc="$src"
+            catpid=''
+        fi
         spinner 1 tar --checkpoint=200 --checkpoint-action=exec=echo \
-            --one-file-system -xaf "$src" -C "$CHROOT" --strip-components=1
+            --one-file-system -xaf "$tarsrc" -C "$CHROOT" --strip-components=1
```
drinkcat: Actually, I'm really unclear how this works. I didn't try your modified `edit-chroot`, but I ran some small test on Arch, and Chrome OS (peppy beta, `6457.36.0`), and somethign like `cat file.tar.gz > fifo`, `tar xavf fifo` does not work (tar hangs). 

Looking at `ltrace` output, `tar` (when `-a` is specified), first reads a block from the file, then closes it, and `gzip` then reopens the file. That won't work with fifos. We need to detect the archive type (`-z`, `-j`).

host-bin/edit-chroot
```diff
@@ -477,7 +538,8 @@ for NAME in "$@"; do
         # safety. We don't do this when encrypting a chroot (see mount-chroot),
         # because that would require 2x the space on one device. When switching
         # filesystems like this, however, that isn't a concern.
-        if [ ! "`getmountpoint "$target"`" = "`getmountpoint "$CHROOT"`" ]; then
+        if [ ! "`getmountpoint "$target"`" = \
```
drinkcat: !=

host-bin/edit-chroot
```diff
@@ -335,15 +363,34 @@ for NAME in "$@"; do
...
         echo -n "  Backing up $CHROOT to $dest..." 1>&2
-        addtrap "echo 'Deleting partial archive.' 1>&2; rm -f '$dest'"
+        addtrap "echo 'Deleting partial archive.' 1>&2; \
+                 kill '$splitpid' 2>/dev/null; rm -f '$dest' '$dest.part'*"
         ret=0
         spinner 1 tar --checkpoint=100 --checkpoint-action=exec=echo \
             --one-file-system -V "crouton:backup.${date%-*}${date#*-}-$NAME" \
-            -caf "$dest" -C "$CHROOTS" "$NAME" || ret=$?
-        if [ "$ret" != 0 ]; then
+            -caf "$tardest" -C "$CHROOTS" "$NAME" || ret="$?"
```
drinkcat: Similar remark as below, `-a` won't work as indented, and all split archives will be uncompressed.

host-bin/enter-chroot
```diff
@@ -385,10 +385,120 @@ if ! mountpoint -q "$destmedia"; then
...
+# Defines directories to map into the chroot from the host.
+# Order is important: be sure to mount directories before you mount
+# subdirectories of those mounts.
+# The only directories that can be mapped from outside of the chroot are
+# subdirectories of the following locations:
+#  downloads: ~/Downloads
+#  shared: /mnt/stateful_partition/crouton/shared
+#  encrypted: ~/crouton/shared
+# Add a line with a single keyword to modify the mount options of all the bind
+# mounts that follow. For example, the following will make all following mounts
```
drinkcat: This feels error-prone, I'd rather specify the options on each line (default to exec is good)

Sorry, again, I'm a bit late in the discussion, but I'd prefer a `fstab`-like syntax, without the arrow, e.g.:
```
downloads ~/Downloads exec
shared/home ~/Share noexec,ro
```

host-bin/enter-chroot
```diff
@@ -385,10 +385,120 @@ if ! mountpoint -q "$destmedia"; then
...
+    localencrypted='/home/chronos/user/crouton/shared'
+    if [ ! -d "/home/chronos/user" ]; then
+        localencrypted=''
+    fi
+    localshare='/mnt/stateful_partition/crouton/shared'
+    # Allow line continuation characters (no -r) and strip leading/trailing
+    # whitespace, remove comments, and remove spaces around ->
+    sed 'y/\t/ /; s/^ *//; s/ *$//; s/ *-> */->/; /^#/d;' "$shares" \
+    | while read line; do
+        src="${line%%->*}"
```
drinkcat: Ah, I see... The arrow makes parsing far easier... It wouldn't be impostible without it, though...

Also, some invalid lines in the file may get set as mount options... Either way, I would like a bit more sanity checking... (mount options are comma-separated, probably no space, etc.)

host-bin/enter-chroot
```diff
@@ -385,10 +385,120 @@ if ! mountpoint -q "$destmedia"; then
...
+            dest="${dest#[\"\']}"
+            dest="${dest%[\"\']}"
+        fi
+        dest="${dest%/}/"
+        # .. path elements are not allowed
+        if [ "${src%/../*}" != "$src" -o "${dest%/../*}" != "$dest" ]; then
+            echo "Invalid share: $line" 1>&2
+            continue
+        fi
+        # Expand src with a little bit of leniency
```
drinkcat: I would do exact matches, to avoid mistakes like a missing slash...

installer/functions
```diff
@@ -153,6 +201,35 @@ validate_name() {
...
 
+# Returns the mountpoint a path is on. The path doesn't need to exist.
+# $1: the path to check
+# outputs on stdout
+getmountpoint() {
+    mp="`readlink -m -- "$1"`"
+    while ! stat -c '%m' "$mp" 2>/dev/null; do
+        mp="${mp%/*}"
+    done
+}
```
drinkcat: `getmountpoint /xyz` will result in an infinite loop (`/xyz` is stripped to an empty string).

Possible workaround: `stat -c '%m' "${mp:-/}"` (`getmountpoint ""` will fail as expected, due to `readlink`)

build/wrapper.sh
dnschneid: Good call.

host-bin/edit-chroot
dnschneid: Because we manually mounted with mount-chroot, so we'd like to put it back the way we found it.  If the chroot is active, unmount-chroot will fail, which is OK if it doesn't really need to be unmounted to continue.

host-bin/edit-chroot
dnschneid: At that point, I might as well just put in `4294967295`, since that's the actual file size limit.

host-bin/edit-chroot
dnschneid: When cat goes from file to file, it seems to generate EOFs between each file that screws up tar.  Piping through dd turns it into a standard stream.  Not really sure why this is.

host-bin/edit-chroot
dnschneid: Argh, yeah.  OK.  I now have it trying --test-label with the different available parameters to see which one works, then using that for the actual extraction.

installer/functions
dnschneid: Nice catch.

host-bin/edit-chroot
dnschneid: FAT32 limits the number of files in a directory to 65534.

So you get five digits now :)

host-bin/enter-chroot
dnschneid: Done!  It's all fstabby now.

host-bin/edit-chroot
dnschneid: Fixed by tweaking the pipe filename

dnschneid:
I think I've addressed everything.

installer/functions
```diff
@@ -68,6 +68,54 @@ undotrap() {
...
+            # Still need to loop through to fix the ordering
+            getopts_dashdash=y
+        else
+            set -- "$@" "$1"
+            getopts_npos="$((getopts_npos+1))"
+        fi
+        shift
+    done
+    [ "$getopts_ret" = 0 ]'
+
```
drinkcat: Finally managed to review this properly. Looks good!

host-bin/edit-chroot
drinkcat: Understood.

host-bin/edit-chroot
drinkcat: Okay, managed to reproduce with small split sizes. `tar` does not like partial reads when reading from what it thinks is a file (and when no compression is used). You should add `-B` to the extract operation below:

```
-B, --read-full-records
              When listing or extracting, accept incomplete input records  after  end-of-
              file marker
```
(I don't think the help text makes too much sense, I was always told that partial reads do not mean EOF (0-length read does); it's clearer when you look at `tar` source)

Note that using `cat fifo | tar xvf -` sets the option by default, as well as when the archive is compressed (since `gunzip` writes irregular blocks).

The `dd` hack only works because more data gets buffered (dd block size is 512, while tar blocks appear to be 10k, so it's not guaranteed to work).

host-bin/edit-chroot
```diff
@@ -385,8 +434,33 @@ for NAME in "$@"; do
...
         echo -n "  Restoring $src to $CHROOT..." 1>&2
         mkdir -p "$CHROOT"
-        spinner 1 tar --checkpoint=200 --checkpoint-action=exec=echo \
-            --one-file-system -xaf "$src" -C "$CHROOT" --strip-components=1
+        if [ -f "$src.part01" ]; then
+            # Detect the type of compression before sending it through a fifo
+            for tarparam in -z -j -J -Z --no-auto-compress fail; do
+                if [ "$tarparam" = 'fail' ]; then
+                    error 2 "Unable to detect compression method of $src"
+                elif tar $tarparam --test-label -f "$src" >/dev/null 2>&1; then
```
drinkcat: Is there a minimum chunk size under which this won't work?  (this might be tricky with different compression algos)

It might be better to use `cat "$src" "$src.part"* | tar $tarparam --test-label`, tar only reads the first block, then closes the pipe, so it should be ok performance-wise.

host-bin/enter-chroot
```diff
@@ -385,10 +385,139 @@ if ! mountpoint -q "$destmedia"; then
...
+
+# Examples:
+#  Share a home directory encrypted by the current user's unsynced profile
+#   encrypted/home ~
+#  Share an unencrypted directory between chroots.
+#   shared/home ~/Share
+#  Absolute directories on the chroot side work too
+#   shared/bin /usr/local/bin exec,suid
+#  Share a noexec path that has spaces, and a specific user home directory
+#   "encrypted/stupid example" "~omg/Why are you even doing this?" noexec
```
drinkcat: Very good example ,-))))

host-bin/enter-chroot
```diff
@@ -385,10 +385,139 @@ if ! mountpoint -q "$destmedia"; then
...
+                fi
+                src="$localencrypted/${src#*/}";;
+            share|shares|shared)
+                src="$localshare/${src#*/}";;
+            *)
+                echo "Invalid share:$line" 1>&2
+                continue;;
+        esac
+        # Expand dest for homedirs
+        if [ "${dest#[~/]}" = "$dest" ]; then
```
drinkcat: You do not document this case, nor provide an example for it, in the description above.

Either do, or fail in that case (i.e. do not allow path that do not start in `/` or `~`).

test/tests/12-edit-chroot
```diff
@@ -88,6 +83,16 @@ host enter-chroot -n "$release" true
...
+# Back up with a split
+host edit-chroot -y -s 50000000 -f "$BACKUPDIR/$release.tar" -b "$release"
+test "`ls "$BACKUPDIR"/* | wc -l`" -ge 3
+du -b "$BACKUPDIR"/* | passes awk '$1>50000000{exit 1}'
+
+# Restore the split archive. Fails first due to existence
+fails host edit-chroot -y -f "$BACKUPDIR/$release.tar" -r "$release"
+host edit-chroot -y -f "$BACKUPDIR/$release.tar" -rr "$release"
+host enter-chroot -n "$release" true
+
```
drinkcat: Might be good to add a test for compressed split archives, since this was broken before.

Using
```
test `"cat archive.tar.gz | hexdump -v -n2 -e'"%02x"'`" = "8b1f"
```
To check that the file is indeed compressed (`8b1f` is for gzip).

host-bin/edit-chroot
dnschneid: If it fails here, it would fail in the one or two times we do the same check above (release detection, validity check, etc).
If we're concerned that someone will do an -s that results in smaller than one block, we should limit the minimum -s to something reasonable.

Actually, we should just have -s specify in megabytes.

installer/functions
dnschneid: Woohoo!

host-bin/edit-chroot
dnschneid: Ah, that makes a lot of sense.  Excellent sleuthing!

host-bin/edit-chroot
```diff
@@ -385,8 +434,33 @@ for NAME in "$@"; do
         fi
         echo -n "  Restoring $src to $CHROOT..." 1>&2
         mkdir -p "$CHROOT"
-        spinner 1 tar --checkpoint=200 --checkpoint-action=exec=echo \
-            --one-file-system -xaf "$src" -C "$CHROOT" --strip-components=1
+        if [ -f "$src.part01" ]; then
+            # Detect the type of compression before sending it through a fifo
+            for tarparam in -z -j -J -Z --no-auto-compress fail; do
```
drinkcat: `-Z` is not needed: Chromium OS does not ship with `compress` anyway (I don't have it either on my Arch install).

dnschneid: Yeah, I included it to be complete, just in case it gets added down the line, or someone adapts the scripts to something else.  It doesn't hurt anything.

host-bin/edit-chroot
drinkcat: Tried to estimate minimum block size:

```
cat big.tar | dd count=1 bs=1K iflag=fullblock | tar --test-label -v; echo $?
gzip -c big.tar | dd count=1 bs=50K iflag=fullblock | tar --test-label -vz; echo $?
bzip2 -c big.tar | dd count=1 bs=900K iflag=fullblock | tar --test-label -vj; echo $?
xz -c big.tar | dd count=1 bs=20K iflag=fullblock | tar --test-label -vJ; echo $?
```

It depends on the compression ratio of the archive, but bzip2 has a clear minimum at 900K (it compresses blocks of 900K of input).

Maybe 10M to be safe? (lzma -6 in memory dictionary is 8MiB) I mean, it's not like somebody is going to backup to floppy disks (right???!?)

dnschneid: I guess we can break floppy disks.  10M it is.

Issue #713, opened 2014-03-22
Logitech Ultrathin Touch Mouse T630 does not work


hubbamj:
So, I have tried every version of ubuntu I can install with crouton, and none of them seem to allow the best mouse on the market, the Logitech T630 to work. The mouse works fine in chromium os but the second you transfer over to ubuntu, all you can get to register are clicks, not actual mouse movement. 

Other bluetooth mice work, such as Microsoft bluetooth mice, so I don't know if it is logitech bluetooth that does not work, or just this logitech mouse. Any ideas would be appreciated. I am new to linux, so please treat me like an idiot. 

I am using an HP Chromebook 14 and I have tried Ubuntu 12.04, 13.10 and Unity, gnome, and xfce desktops on each (yes, a lot of re installing)

dnschneid:
It's worth trying the branch in #1302 to see if it fixes your problems.  It may be cmt taking over the mouse and not letting standard events through, which means we would need to either reactivate the standard events in crouton (probably the right short-term solution) or get cmt working better (the long-term solution).

trisk:
I can confirm I've had the same issue with an older Logitech Wireless Touchpad for the past year. The Logitech touchpad devices default to mouse emulation, with firmware handling gestures, but cmt has special support for multitouch that probably puts the devices in a different mode.

The physical buttons work, but not touch gestures or movement. When starting Xorg, synaptics picks up the device by default due to the "touchpad" input class but synaptics is the wrong driver, anyway.

I've tried:
* Overriding the driver to evdev, which still doesn't produce events (probably because of the device state?)
* Installing Hugh Greenberg's cmt driver packages for trusty and using the same 40-touchpad-cmt.conf
The latter results in:
```
[    49.700] (II) Using input driver 'cmt' for 'Logitech Unifying Device. Wireless PID:4011'
[    49.700] (**) Logitech Unifying Device. Wireless PID:4011: always reports core events
[    49.700] (**) /dev/input/event15(**) Logitech Unifying Device. Wireless PID:4011: always reports core events
[    49.705] (**) Option "config_info" "udev:/sys/devices/pci0000:00/0000:00:1d.0/usb4/4-1/4-1.8/4-1.8.1/4-1.8.1.3/4-1.8.1.3:1.2/0003:046D:C52B.0003/001D:046D:4011.0005/input/input16/event15"
[...]

PR #1302, opened 2015-01-03
Experimental cmt target

jacklevy:
Experimental cmt target imports Chromium OS multi-touch driver binaries and configures Ubuntu to use them instead of the standard Synaptics driver.

Advantages vs. Synaptics driver:

- Reliable two-finger scrolling
- Three-finger multi-touch detection
- Better palm/edge detection

Disadvantages:

- Not a proper port of cmt; may not work on all platforms/devices, may break at any time as a result of updates to Chromium OS
- Won't work if your chroot does not use the same architecture as the Chromium OS host (64-bit vs. 32-bit)


dnschneid:
Ooh, for once a PR that's not a thousand lines!

While I'm looking at this could you please sign the [CLA](https://cla.developers.google.com/clas/new?kind=KIND_INDIVIDUAL&domain=DOMAIN_GOOGLE)?

chroot-bin/croutonwheel
```diff
@@ -12,11 +12,27 @@ l=6
...
 x=1
-# Set the default scaling and constant values
-xs='0.2'
-xc='0.01'
-ys='0.2'
-yc='0.05'
+
+if [ -L /etc/crouton/cmt ]; then
+  # Enable xinput Australian Scrolling
+  xinput --set-prop --type=int --format=8 "Atmel maXTouch Touchpad" "Australian Scrolling" 1
```
dnschneid: We're either going to need to provide a way to configure this, or detect it from the DE's settings if possible.

Also, we need to figure out if it's the same on Cypress and other trackpads...

chroot-bin/croutonwheel
```diff
@@ -46,8 +62,10 @@ if [ $# -ge "$OPTIND" ]; then
     exit 1;
 fi
 
-# Operate on the Chromium OS X11 server
-eval "`host-x11`"
+if [ ! -L /etc/crouton/cmt ]; then
```
dnschneid: Can you rebase off of my `xmethods` branch?  This should check if the xmethod is `xorg` and /etc/crouton/cmt is a link (use -h instead of -L).  Although -d might be better, since you do want the mount to exist (in case the target was partially removed for some reason)...

chroot-bin/croutonwheel
```diff
@@ -64,8 +82,12 @@ flock 3
 {
     croutonxi2event &
     xi="$!"
-    aura="`croutonwmtools list ni | awk '/aura_root/ {print $NF}'`"
-    xev -id "$aura" &
+
+    if [ ! -L /etc/crouton/cmt ]; then
```
dnschneid: Probably should do the xmethod/link test once at the top and store the result in a variable.

chroot-etc/cmt-xorg.conf
```diff
@@ -0,0 +1,3 @@
+Section "Server Flags"
```
dnschneid: cmt only works with `xorg`, so how about modifying chroot-etc/xserverrc-xorg to add -ignoreABI to the flags?

chroot-etc/croutonwheel.desktop
```diff
@@ -0,0 +1,6 @@
+[Desktop Entry]
```
dnschneid: Can't this be automatically started in croutonxinitrc-wrapper?  The same place all the synclient stuff got if'd out.

host-bin/enter-chroot
```diff
@@ -280,6 +280,45 @@ fi
...
+  # cmt driver
+  mkdir -p $CHROMIUM_CMT/usr/lib/xorg/modules/input
+  cp /usr/lib64/xorg/modules/input/cmt_drv.so $CHROMIUM_CMT/usr/lib/xorg/modules/input
+  ln -sfT $CHROOT_CMT/usr/lib/xorg/modules/input/cmt_drv.so $CHROOT/usr/lib/xorg/modules/input/cmt_drv.so
+
+  # libs loaded by cmt driver
+  for lib in libevdev libgestures libjsoncpp
+  do
+    cp /usr/lib64/$lib.so $CHROMIUM_CMT/usr/lib
+    ln -sfT $CHROOT_CMT/usr/lib/$lib.so $CHROOT/usr/lib/$lib.so
```
dnschneid: These are only required by the Xorg process, right?  Can you modify xserverrc-xorg to add an LD_LIBRARY_PATH instead of adding a bunch of symlinks to /usr/lib?

dnschneid: Ditto for the two sets below.

host-bin/enter-chroot
```diff
@@ -280,6 +280,45 @@ fi
...
+  do
+    xorgconffile=$(basename $xorgconf)
+    cp $xorgconf $CHROMIUM_CMT/etc/X11/xorg.conf.d
+    ln -sf $CHROOT_CMT/etc/X11/xorg.conf.d/$xorgconffile $CHROOT/etc/X11/xorg.conf.d/$xorgconfile
+  done
+
+  # cmt driver
+  mkdir -p $CHROMIUM_CMT/usr/lib/xorg/modules/input
+  cp /usr/lib64/xorg/modules/input/cmt_drv.so $CHROMIUM_CMT/usr/lib/xorg/modules/input
+  ln -sfT $CHROOT_CMT/usr/lib/xorg/modules/input/cmt_drv.so $CHROOT/usr/lib/xorg/modules/input/cmt_drv.so
```
dnschneid: Please use fixabslinks for the xorg directory in the chroot.

targets/cmt
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh -e
+# Copyright (c) 2013 The Chromium OS Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+REQUIRES='x11'
+DESCRIPTION='Installs cmt binaries from Chromium OS host as touchpad driver.'
```
dnschneid: `'Improves xorg trackpad performance by using Chromium OS'\''s multitouch driver.'`

targets/cmt
```diff
@@ -0,0 +1,26 @@
...
+
+### Append to prepare.sh:
+
+# provides xte for croutonwheel
+install xautomation
+
+# chroot /etc/crouton/cmt symlink activates cmt functionality
+CHROMIUM_CMT=/var/host/cmt
+CROUTON_ETC=/etc/crouton
+ln -sf $CHROMIUM_CMT $CROUTON_ETC
```
dnschneid: Quote everything as good practice

plevesque:
How can I try this patch on my chromebook?

dnschneid:
[Download @jacklevy's branch](https://github.com/jacklevy/crouton/archive/master.tar.gz), extract using `cd ~/Downloads; tar -xf master.tar.gz`, then use installer/main.sh to update (or install) your chroot `sudo sh ~/Downloads/*/installer/main.sh ...`

Beware though, it's not up-to-date with master, and of course has all those comments pending...

jacklevy:
speaking of pending comments @dnschneid, i appreciate the time you took to provide this feedback and i do intend to make changes accordingly. this one has been a slow burner..

dnschneid:
@jacklevy no worries.  I consider PRs on open-source projects like a modern-day chess-by-mail...it'll happen eventually as both sides find time :)

chroot-etc/croutonwheel.desktop
jacklevy: the original issue was that `croutonwheel` couldn't start until after `startunity` calls `exec gnome-session-wrapper ubuntu`, otherwise `xwininfo` complained and failed since X hasn't started yet.  use of `xwininfo` has since been replaced by `croutonwmtools`, and you mentioned that once the smartcycle branch is merged, `croutonwheel` should work when launched from `croutonxinitrc-wrapper`.  is this still the case?

trying to find a proper place to start `croutonwheel` reliably in unity was actually one of the more difficult parts of this PR. if this still won't work from `croutonxinitrc-wrapper`, the good news is that the `croutonwheel.desktop` solution (executed via `/etc/xdg/autostart`) is a solid alternative that plays well with ubuntu.

plevesque:
Ok I have tested and mouse is moving, both left and right click are working but scrolling is not working.  I'm testing in kde.  Thanks for the support

jacklevy:
@plevesque is `croutonwheel` running?  perhaps the startup method i used does not work with KDE, although i would be surprised given that it worked with xfce.  check for the process via `ps` -- if it is not running then try starting it manually (`/usr/local/bin/croutonwheel`) and see if scrolling works.

plevesque:
No it is not running and starting it manually return this error:
/usr/local/bin/croutonwheel
unable to find device Atmel maXTouch Touchpad

jacklevy:
sounds like your chromebook does not use an Atmel maXTouch Touchpad, which is an assumption i shouldn't have hard-coded into `croutonwheel`.  luckily this was only necessary to configure Australian Scrolling, which was a preference anyway.

to circumvent this problem, you can edit `/usr/local/bin/croutonwheel` and comment out line 18 ("`xinput --set-prop --type=int --format=8 "Atmel maXTouch Touchpad" "Australian Scrolling" 1`").   if you are inclined, give that a try and let me know if scrolling works (in a non-Australian way).

for the final PR i will have to think about how to best decouple this from being Atmel-specific.

plevesque:
Ok my chromebook use Elan Touchpad, so replacing Atmel maXTouch Touchpad with Elan Touchpad cause the script to run perfectly.  And it start properly on boot up with kde.  Now can I have this "cmt" in sync with the master?

jacklevy:
glad to hear it.  this branch is in sync with master as of the PR date (1/3/15).  i will sync it up again when i revise the PR to incorporate david's feedback.  in the meantime, you can always fork my fork and then merge from dnschneid:master upstream to get the latest.

chroot-etc/croutonwheel.desktop
dnschneid: Display is definitely available in croutonxinitrc-wrapper, so calling croutonwheel there should be fine.

trisk:
I originally commented on #713, but I'll move it here for clarity:
Yes, the cmt target works and addresses my issue. I am using trusty in crouton on a dev channel Chromium build.

I had to symlink ```/usr/lib/x86_64-linux-gnu/libstdc++.so.6``` -> ```/usr/local/lib/libstdc++.so``` to get ```GLIBCXX_3.4.20``` versioned symbols required by ```/usr/lib/libgestures.so.0```. I expected specifying ```export LD_LIBRARY_PATH="/usr/local/lib:$LD_LIBRARY_PATH"``` in ```xserverrc-local``` to be sufficient to pick up the right ```libstdc++.so.6```, but it was not.

After copying the external ```40-touchpad-cmt.conf``` to ```/etc/crouton/cmt/etc/X11/xorg.conf.d```, it behaves identically to the Chromium OS configuration. Scrolling works in Chrome only, as expected. I'll try to get croutonwheel to work sanely.

trisk:
For some reason croutonwheel didn't start with the desktop file, but moving it to croutonxinitrc-wrapper  (contingent on ```[ "$XMETHOD" = 'xorg' -a -L /etc/crouton/cmt ]```) worked fine. I also was able to enable Australian scrolling for my Logitech touchpad with:
```diff
-  xinput --set-prop --type=int --format=8 "Atmel maXTouch Touchpad" "Australian
+  for id in `xinput list --id-only`; do
+    xinput set-prop --type=int --format=8 "$id" "Australian Scrolling" 1 2>/dev/null
+  done
```
I will create a pull request with my changes.

dnschneid:
Thanks!

trisk:
https://github.com/jacklevy/crouton/pull/1
PR was done against jacklevy's master branch rather than including the entire cmt target, though I also ended up doing a merge with upstream at the time.

commiades:
I am wanting to use a Logitech T630 on an Xfce chroot. I have patched the chroot using the method described by dnschneid on Feb 10.

The mouse now works, but the chroot has some issues, one is that I get this message:

"WARNING: Outdated server version (VF1), expecting VF3. Please update your chroot."

Another is that the keyboard target no longer works, nor does the four fingered salute; instead I get told that crouton integration has triggered full screen.

I backed up before I did this. How can I fix these problems?

dnschneid:
Going to close this (quite nifty) pull-request due to apparent lack of interest.  I think as we move towards containers and wayland and whatever, this won't be needed anyway.

Issue #1198, opened 2014-11-17
priortize targets


DennisLfromGA:
Question/Enhancement: 

It seems the order in which we list targets on the command line is becoming more critical.

With the addition of some of the 'unique' targets like xephyr, xorg, xiwi and some that are mutually exclusive or incompatible, would it be worthwhile to have a 'priortizetargets' function or script segment, kind of like 'deduptargets' in main.sh or maybe something in targets/common, that would re-sort the targets to ensure that certain ones are listed first or at least before other less critical ones and remove unnecessary ones ?

divx118:
Yes, sorry you are totally right. If I understand correctly it should be something like https://github.com/divx118/crouton/commit/71f27c18865ab7524326651aa29b0b3cc7636ade

dnschneid:
That's looking quite good, thanks.  Could you open a PR for that and I'll post some feedback?

dnschneid:
Fixed in master!  Thanks, @divx118

DennisLfromGA:
Yes, thanx @divx118 :+1: 
I have lots of ideas but fewer & fewer solutions... ;-)

PR #1305, opened 2015-01-04
Prioritize targets which PROVIDES

divx118:
Regarding https://github.com/dnschneid/crouton/issues/1198
This should for example prevent installing target xorg when xfce, xiwi are targets and xfce is first defined in `-t xfce,xiwi` or xiwi is already installed and you update with xfce.

### DIFF
--- a/targets/common
+++ b/targets/common
@@ -33,17 +33,12 @@ if [ "$TARGETS" = 'help' ]; then
         echo "	Requires: $REQUIRES"
     fi
 elif [ "$TARGETNOINSTALL" = "p" ]; then
-    # Avoid adding things already provided
-    # This will only check the provides from /etc/crouton/targets
-    # and targets specified with -t.
+    # This will check the provides from /etc/crouton/targets and targets
+    # specified with -t. If there are duplicates we will respect the
+    # order of the targets specified. This means on an update targets
+    # specified with -t will have higher priority then already installed targets.
     for t in $PROVIDES; do
-        # Don't add duplicate PROVIDES. If there are duplicates
-        # we will respect the order of the targets specified.
-        # This means on an update targets specified with -t
-        # will have higher priority then already installed targets.
-        if ! grep -q "^$t=.*\$" "${PROVIDESFILE:-/dev/null}"; then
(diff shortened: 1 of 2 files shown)


targets/common
```diff
@@ -45,6 +58,10 @@ elif [ -z "$TARGETNOINSTALL" -o -n "$RESTOREHOSTBIN" ]; then
     done
     # Source the prerequisites
     for t in $REQUIRES; do
+	# If it is already provided change it to the target what provides it.
+        if grep -q "^$t=.*\$" "${PROVIDESFILE:-/dev/null}"; then
```
dnschneid: No need to grep before the awk.  Store in a temporary variable and then say `t="${provider:-"$t"}"`.

targets/common
```diff
@@ -45,6 +58,10 @@ elif [ -z "$TARGETNOINSTALL" -o -n "$RESTOREHOSTBIN" ]; then
...
+            t="`awk -F= '/'$t'=/{print $2}' "${PROVIDESFILE:-/dev/null}"`"
```
dnschneid: Add double-quotes around the `$t` part.  Not really necessary here, but good practice.

targets/common
```diff
@@ -32,6 +32,19 @@ if [ "$TARGETS" = 'help' ]; then
...
+elif [ "$TARGETNOINSTALL" = "p" ]; then
+    # Avoid adding things already provided
+    # This will only check the provides from /etc/crouton/targets
+    # and targets specified with -t.
+    for t in $PROVIDES; do
+        # Don't add duplicate PROVIDES. If there are duplicates
+        # we will respect the order of the targets specified.
+        # This means on an update targets specified with -t
+        # will have higher priority then already installed targets.
+        if ! grep -q "^$t=.*\$" "${PROVIDESFILE:-/dev/null}"; then
```
dnschneid: There's really no harm in adding duplicate provides if your awk statement down below has `print $2; exit`.  First one will naturally take precedence.

dnschneid:
Minor nits; otherwise looking great.

divx118:
A bit late, but got interrupted yesterday. Fixes applied according to comments. Thanks for the tip on awk.

dnschneid:
Thanks!

Issue #1324, opened 2015-01-07
Super_L key combos not being passed to xiwi


dylanPowers:
Super_L key combos are only working with the standard alphabet keys in xiwi. They aren't working with the number keys, the arrow keys, the spacebar, tab, the enter key, and pretty much anything that isn't alphabetic. 
I did a little debugging with xev and I found that the super_L down event isn't being passed until either the up event occurs or an alphabetic key is pressed.

This is a bit of a show stopper for me and keeps me from using xiwi as my Awesome WM configuration depends a lot on those key combos. 
As mentioned in https://github.com/dnschneid/crouton/issues/1289#issuecomment-68762070 @drinkcat you seem to know a lot about what would be needed to fix this :) You're doing some amazing work!

drinkcat:
Chromium OS under freon (e.g., peppy, dev-channel, `6592.2.0`) always reports "raw" keyboard events (no Search+Left -> Home translation). Which is awesome.

We need to disable the `Super_L` hack when running under freon. Then, we could just use the current xkb keyboard layout (`keyboard target`), or implement our own key translation in Kiwi (might be better as xkb is a bit quirky...).

And I don't think it is worth improving the current situation (under X11).

Issue #1275, opened 2014-12-25
xiwi confuses the keyboard layout


dnschneid:
xiwi appears to pass through post-Chromium-keyboard-layout keypresses, which then confuses X11.  That is, if you use a keyboard layout that maps s -> r and r -> p, pressing the S key in xiwi will result in an R keypress in X11.  If you are then silly enough to set up the same keyboard layout inside of X11, that R will then be helpfully translated into a P by X11.

xiwi should send raw keypresses, disregarding any keyboard layout settings.

dylanPowers:
Is this related to certain key combo's not being passed? Things like super+enter, super+space, super+arrows, and super+num don't work. Or because they're connected to actions in chromium OS should I forget about them working how I want?

dafyddg:
I have successfully updated my xfce on trusty crouton installation on the HP 14" Chromebook with the xiwi and extension targets. Everything works fine except that using the German keyboard layout a large number of keys (special characters, punctuation) yield incorrect characters. Several keys produce the same characters. Updating again with the x11 target removes xiwi and also the normal German character mapping.
Is there a fix for this? If so, details would be appreciated. I have not seen one so far. Thanks.

drinkcat:
Looks like we may need to use the key's code string to guess the key:

Pressing semicolon with US keyboard:
`Key UP: C:Semicolon/KC:ba/KS:3b searchstate:0` (outputs a semicolon)
And pressing it with CH(FR) keyboard (should be a 'é'):
`UP: C:Semicolon/KC:de/KS:27 searchstate:0` (outputs a quote, no matter the layout in the chroot)

Provided crbug.com/425156 is fixed. Need to check how Chromoting does it, too.

drinkcat:
So, Chrome Remote Desktop uses `GetCode()` (the string above).

The NaCl module will need to translate this string to a X11 keycode (i.e. key position on the keyboard).

What we need to do is to modify the protocol to pass X11 keycodes, instead of keysyms (translated symbol, layout-dependant), so that the crouton X11 server can do the appropriate translation. This means we need to bump the protocol version to `VF2`. Unfortunately, the extension is completely broken when it comes to error reporting: it cannot distinguish between protocol errors, version mismatch, or "normal" disconnect. And we cannot break compatibility with "old" chroots unless we have clear and user friendly reporting.

I started working on this, but it's going to take a bit of time to implement and test properly.

[...]

PR #1339, opened 2015-01-12
Kiwi: Send keycodes instead of keysyms

drinkcat:
This is a partial fix for #1275 (but still much better than current situation, so it can be merged as-is).

The protocol now sends X11 key codes instead of key symbols. Keycodes (mostly) correspond to physical keyboard keys, so that the chroot can translate it, depending on the keyboard layout set in crouton.

For that purpose, I had to update protocol version from VF1 to VF2. The extension is backward compatible, though.

What works:
 - Setting US keyboard in Chromium OS (or any keyboard layout without dead keys), any keyboard layout in crouton should work fine.
 - Extension can connect to old chroot (VF1), in that case keysyms are sent.

What still does not work:
 - Setting a keyboard layout in Chromium OS with dead keys (e.g. US Intl) leads to lost key events (I think it's fixable, as the key code appears when the key is lifted).
 - Basic framework for reverting Search+key mapping is there, if the user wants Search+Left to mean Super+Left and not Home (see #1324). Not implemented yet (we need another configuration check box).

Other things to be implemented in other PR:
 - Warn the user that it is connecting to an old chroot.
 - Eventually, drop support for VF1 (this requires extensive rework of error handling in kiwi)

What I don't think can be fixed:
 - Switching Ctrl/Search/Alt in Chromium OS also results in switched positions in crouton.
 - Search+numbers appears the same as Search+fn keys (F1-F10). I don't think we can distinguish between the 2, to support Super+numbers (see #1324).

### DIFF
--- a/host-ext/nacl_src/keycode_converter.h
+++ b/host-ext/nacl_src/keycode_converter.h
@@ -1,0 +1,215 @@
+/* Copyright (c) 2015 The crouton Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ *
+ * Translates NaCl pp::KeyboardInputEvent::KeyCode() strings to X11 keycodes.
+ */
+
+#include <map>
+#include <string>
+#include <stdint.h>
+
+/* Values in the keycode hashmap. */
+class KeyCode {
+public:
+    KeyCode(uint8_t base, uint8_t search):
+        base_(base), search_(search) {}
+
+    KeyCode(uint8_t base): KeyCode(base, base) {}
+
+    uint8_t GetCode(const bool search_on) const {
+        if (search_on)
+            return search_;
+        else
+            return base_;
+    }
+
+private:
+    const uint8_t base_;  /* Basic keycode */
+    /* Reverse translation of keycode when Search is pressed: e.g.
+     * Search+Left => Home. In this case:
+     * base_ = Home keycode (0x6e), search_ = Left keycode (0x71) */
+    const uint8_t search_;
+};
+
+/* Class with static members only: converts KeyCode string to X11 keycode */
+class KeyCodeConverter {
+public:
+    static uint8_t GetCode(const std::string& str, const bool search_on) {
+        if (!strcodemap_)
+            InitMap();
+
+        auto it = strcodemap_->find(str);
+        /* Not found */
+        if (it == strcodemap_->end())
+            return 0;
+
+        return it->second.GetCode(search_on);
+    }
+
+private:
+    static void InitMap();
+    /* static pointer member, to avoid static variable of class type. */
+    static std::map<std::string, const KeyCode>* strcodemap_;
+};
+
+std::map<std::string, const KeyCode>* KeyCodeConverter::strcodemap_ = NULL;
+
+/* Initialize string to X11 keycode mapping. Must be called once only.
+ *
+ * FIXME: Fill in search_ fields in KeyCode.
+ *
+ * Most of this data can be generated from
+ * ui/events/keycodes/dom4/keycode_converter_data.h in the Chromium source
+ * tree, using something like:
+ * sed -n \
+'s/.*USB_KEYMAP([^,]*, \([^,]*\),.*, \("[^"]*"\).*$/{\2, KeyCode(\1)},/p' \
+keycode_converter_data.h | grep -v "0x00" >> keymap_data.h
+ */
+void KeyCodeConverter::InitMap() {
+    strcodemap_ = new std::map<std::string, const KeyCode>({
+        {"Sleep", KeyCode(0x96)},
+        {"WakeUp", KeyCode(0x97)},
+        {"KeyA", KeyCode(0x26)},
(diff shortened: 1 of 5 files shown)


drinkcat:
(untested on freon, my peppy refuses to switch to dev channel...)

host-ext/nacl_src/keycode_converter.h
```diff
@@ -0,0 +1,215 @@
...
+        return it->second.GetCode(search_on);
+    }
+
+private:
+    static void InitMap();
+    /* static pointer member, to avoid static variable of class type. */
+    static std::map<std::string, const KeyCode>* strcodemap_;
+};
+
+std::map<std::string, const KeyCode>* KeyCodeConverter::strcodemap_ = NULL;
```
dnschneid: std::map is actually really inefficient at this, especially with string input data with common prefixes.

You want either an unordered_map or (more fun) check out gperf.

host-ext/nacl_src/keycode_converter.h
```diff
@@ -0,0 +1,215 @@
...
+#include <string>
+#include <stdint.h>
+
+/* Values in the keycode hashmap. */
+class KeyCode {
+public:
+    KeyCode(uint8_t base, uint8_t search):
+        base_(base), search_(search) {}
+
+    KeyCode(uint8_t base): KeyCode(base, base) {}
```
dnschneid: explicit

dnschneid:
This works well on Freon.  I'm inclined to merge it as-is, since my comments are only nits.

stsquad:
I was having trouble testing this with the current extension. Does the new extension get rolled out automatically?

drinkcat:
@stsquad : The extension auto-updates, when no chroot is running (version in chrome://extensions should now be 2.1.0). You also need to update your chroot.

host-ext/nacl_src/keycode_converter.h
drinkcat: gperf looks fun indeed, I'll keep it in mind. A bit overkill for this, though (FYI, Chrome Remote Desktop simply iterates over the list and compares strings).

I'll switch to unordered_map.

host-ext/nacl_src/keycode_converter.h
drinkcat: Sure. Will fix in next PR.

stsquad:
@drinkcat \o/ with the updated extension I just tested master and I have ~# and `¬ working ;-)

drinkcat:
> \o/ with the updated extension I just tested master and I have ~# and `¬ working ;-)

Good to hear!

host-ext/nacl_src/keycode_converter.h
dnschneid: I know...I've always wanted to find a use for gperf though :)
I'm not even sure unordered_map is necessary.

drinkcat:
See https://code.google.com/p/chromium/issues/detail?id=425156#c16.

Both "What still does not work", and "What I don't think can be fixed" are magically fixed with freon! Now there is still a slight bug as "OSLeft"/Super_L event are sometimes discarded. Happy not to have to implement another horrible hack...

dnschneid:
Awesome!  I think freon will be generally a good thing for crouton.

cribalik:
Just out of curiosity, is there any reason to still keep the fix in [this commit](https://github.com/dnschneid/crouton/commit/9c5e1cc57e9aabb985534d52c7dedebf15bb5151) when using Freon? It seems to be the culprit of [#1558](https://github.com/dnschneid/crouton/issues/1558).

dnschneid:
@cribalik can you confirm that #1275 doesn't re-emerge if you revert this?

cribalik:
@dnschneid I confirmed it with Swedish, Dvorak US and Spanish keyboards in Chrome; the keys recieved from X11 were the same regardless of the layout in Chrome. Switching around the Alt/Search/Ctrl keys also did not have any impact on X11 keycodes. Search + LeftArrow is received as  OSLeft + LeftArrow, and not OSLeft + Home (which is what [the fix](https://github.com/dnschneid/crouton/commit/9c5e1cc57e9aabb985534d52c7dedebf15bb5151) was for I believe). 

This seems to be Freon's doing that suddenly the correct keycodes are sent. Do we have to keep backwards compatability with non-Freon users? If not, I propose reverting the aforementioned fix.

dnschneid:
Many users are still on non-freon systems, so a straight-out revert is not an option (yet).  We'll need to either have two code paths based on whether Freon is present or not (which probably means having the chroot tell the extension one way or the other), or wait until Freon is in stable on all devices to revert.

tedm:
My device (2012 Samsung ARM / stable channel) isn't on Freon yet, but is very stable with precise, and I wouldn't mind losing the GUI for awhile. I've got a blog platform underneath that is very stable (open ssh, mysql, apache2, php5, myphpadmin, wordpress, etc.). I'm willing to move to trusty at some point, but precise has been extremely stable the past few months on the ARM. I mostly run the chroot with enter-chroot and utilize the server processes.

PR #1347, opened 2015-01-16
Raise Chromium OS notifications from crouton

### LINKED ISSUE #581: Raise notification in Chrome OS from chrooted environment
There is a script croutonurlhandler for opening URL in Chrome OS and optionally switching to it. It could be great to be able to raise HTML5/Chrome notifications like this. So someone who prefer to work with crouton from crosh/Secure Shell could be notified using the command like notify-send.

drinkcat:
Creates a new message type, "N", that can be sent from `croutonnotify`. It is a JSON-formatted map, compatible with
`chrome.notifications` option format (see https://developer.chrome.com/apps/notifications), with 2 additional fields:
 - `crouton_id`: The notification ID to pass to `chrome.notifications.create`.
 - `crouton_display`: The display to switch to when the notification is clicked.

This  (partially) fixes #581. Integrating notifications nicely with `libnotify` will be more tricky, I guess.

---

As often, while adding new features, I uncovered a bug in the handling of requests coming from the extension:

Requests from the extension can happen at any time, even between a command and its acknowledgement.

Clarified protocol in https://github.com/dnschneid/crouton/wiki/crouton-extension:-websocket-architecture
We now enforce that requests in the form `?<payload>` must have a reply in the form `?<reply>` (where `?` is the same character), or an error reply `E<error>`. (This is already the case, so no protocol change is required)

Test: Start a Kiwi window, and run:
```
for i in `seq 1 1000`; do croutonnotify -i abcd -m t$i; done
```
Even when switching rapidly between Kiwi and Chromium, the command should never print anything (without this patch, it would show `scros` or `s:1` before repeating `Error: Not connected`), and the extension does not disconnect.

### DIFF
--- a/chroot-bin/croutonnotify
+++ b/chroot-bin/croutonnotify
@@ -29,7 +29,8 @@ Example JSON data:
   "type": "basic",
   "title": "Primary Title",
   "message": "Primary message to display",
-  "crouton_display": ":1"
+  "crouton_display": ":1",
+  "iconUrl": "data:image/png;base64,<base64 encoded png data>"
 }'
 
 . "$(dirname "$0")/../installer/functions"
--- a/chroot-bin/croutonnotify
+++ b/chroot-bin/croutonnotify
@@ -1,0 +1,83 @@
+#!/bin/sh -e
+# Copyright (c) 2015 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+USAGE="${0##*/} -j
+${0##*/} -t title [-m message] [-d display] [-i id]
+Raises a notification in Chromium OS (requires crouton extension).
+
+When -j is specified, reads JSON data from stdin, in the format required
+by chrome.notifications API, with 2 additional fields, \"crouton_id\" and
+\"crouton_display\", corresponding respectively to the notification id, and the
+display to switch to when the notification is clicked (croutoncycle parameter).
+
+Otherwise, constructs a \"basic\" notification with the requested fields.
+
+Options:
+    -j       Read JSON data from stdin (see example below)
+    -t       Title to display (chrome.notifications field: \"title\")
+    -m       Message to display (chrome.notifications field: \"message\")
+    -d       Display to switch to when the notification is clicked (croutoncycle
+             parameter). Default: Do not switch display.
+    -i       ID to pass to chrome.notifications.create: only the last
+             notification with a given ID is shown.
+             Default: Display a new notification.
+
+Example JSON data:
+"'{
+  "type": "basic",
+  "title": "Primary Title",
+  "message": "Primary message to display",
+  "crouton_display": ":1"
+}'
+
+. "$(dirname "$0")/../installer/functions"
+
+DISPLAY=""
+MESSAGE=""
+TITLE=""
+ID=""
+JSON=""
+
+# Process arguments
+while getopts 'd:i:jm:t:' f; do
+    case "$f" in
+    d) DISPLAY="$OPTARG";;
+    i) ID="$OPTARG";;
+    j) JSON="y";;
+    m) MESSAGE="$OPTARG";;
+    t) TITLE="$OPTARG";;
+    \?) error 2 "$USAGE";;
+    esac
+done
+
+# No extra parameters, -j precludes other parameters, and at least title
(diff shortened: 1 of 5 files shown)


chroot-bin/croutonnotify
```diff
@@ -0,0 +1,83 @@
...
+        [ "$JSON" = 'y' -a -n "$DISPLAY$ID$MESSAGE$TITLE" ] ||
+        [ -z "$JSON$TITLE" ]; then
+    error 2 "$USAGE"
+fi
+
+JSONDATA=""
+if [ -z "$JSON" ]; then
+    JSONDATA='{
+        "type": "'"basic"'",
+        "title": "'"$TITLE"'",
```
dnschneid: TITLE, MESSAGE, DISPLAY, and ID parameters [should all be escaped for JSON](http://stackoverflow.com/questions/3020094/how-should-i-escape-strings-in-json) to avoid nasty surprises.

dash and bash are inconsistent about processing escape sequences when expanding variables, but it looks like they both leave \uXXXX sequences intact.  So avoid converting \ into \\'s and use the \uXXXX code instead, and it should remain intact as you use the variables.

host-ext/crouton/background.js
```diff
@@ -404,6 +411,53 @@ function websocketMessage(evt) {
...
         break;
+    case 'N': /* Raise a notification */
+        /* Payload in JSON format, compatible with chrome.extensions specifications */
+        try {
+            var data = JSON.parse(payload);
+
+            if (!data.type)
+                data.type = "basic";
+            if (!data.iconUrl)
+                data.iconUrl = "icon-128.png";
```
dnschneid: Can icon data be serialized into the json and then extracted?  Being able to set the icon to a chroot-local file is important for notifications...

chroot-bin/croutonnotify
```diff
@@ -0,0 +1,83 @@
...
+\"crouton_display\", corresponding respectively to the notification id, and the
+display to switch to when the notification is clicked (croutoncycle parameter).
+
+Otherwise, constructs a \"basic\" notification with the requested fields.
+
+Options:
+    -j       Read JSON data from stdin (see example below)
+    -t       Title to display (chrome.notifications field: \"title\")
+    -m       Message to display (chrome.notifications field: \"message\")
+    -d       Display to switch to when the notification is clicked (croutoncycle
```
dnschneid: Won't the vast majority of -d parameters just be "$DISPLAY"?
How about making -d parameterless and just use the $DISPLAY variable, which really just works the same as notify-send.

drinkcat: Sure. For most cases that'll be ok. And we can still use `-j` if something unusual is required.

chroot-bin/croutonnotify
drinkcat: Done.

host-ext/crouton/background.js
drinkcat: I wasn't aware, but it already works. For example, you can add:
```
"iconUrl": "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12P4//8/w38GIAXDIBKE0DHxgljNBAAO9TXL0Y4OHwAAAABJRU5ErkJg\
gg==",
```

(see: http://en.wikipedia.org/wiki/Data_URI_scheme)

Added a short example in the help text.

dnschneid: Could we take it one step further and provide a `-i` parameter which automatically encodes the image for you (switch id to `-I` since it'll be much less commonly used)?
base64 is easy (and is already a dependency of the extension), and it looks like Chrome figures out what type of file it is regardless of the explicit MIME type (so data:image/png should work even if it's a jpg).

drinkcat: Sure, done. I used the extension to guess the MIME type, defaulting to png.

dnschneid:
Switching doesn't work with xiwi?

dnschneid:
Notification fails to display if the MIME type is invalid, but no error gets reported back.  Try sending an xpm as the icon.

dnschneid:
Oops, git typo :)
Will follow through though. I have a patch for some of my comments.

dnschneid:
Fixed everything but the notification failure; filed #1402 for that.  Yeah launch and iterate!

PR #1353, opened 2015-01-19
CRAS: Remove unneeded patches, patch away SBC dependency

drinkcat:
Bump minimum version fo 5712/R35, and remove unneeded patches in `targets/audio`.

Also, drop SBC dependency from `cras_test_client` (it is only needed for a test within `cras_test_client`). I'll check with upstream how we can fix that better (`ifdef`, separate test, or remove it altogether).


drinkcat:
Tested in `2015-01-19_19-08-12_drinkcat_chroagh_cras-fixes_30`.

dnschneid:
Yesss negative lines

PR #1432, opened 2015-02-22
Add support in croutoncycle for multiple displays

linux4kix:
Under xiwi copy and paste wasn't working properly because
the second display was showing up as aura_root_x as well as the
normal aura_root_0.  This makes a few changes necessary to handle
any aura_root_ root window as type cros.

### DIFF
--- a/chroot-bin/croutoncycle
+++ b/chroot-bin/croutoncycle
@@ -103,7 +103,7 @@ if host-x11 true 2>/dev/null; then
     winlist="`host-x11 croutonwmtools list nim | \
               sort | awk '{ printf $NF " " }'`"
     aurawin="`host-x11 croutonwmtools list ni | \
-              awk '$1 == "aura_root_0" { print $NF; exit }'`"
+              awk '$1 ~ /aura_root_/ { printf $NF " " }'`"
     tty="`cat '/sys/class/tty/tty0/active'`"
 else
     # No X11 server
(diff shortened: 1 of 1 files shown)


chroot-bin/croutoncycle
```diff
@@ -123,7 +123,7 @@ if [ "$tty" = 'tty1' ]; then
     for disp in $winlist; do
         if [ "${disp%"*"}" != "$disp" ]; then
             curdisp="$disp"
-            if [ -n "$xiwiactive" -a "${disp%"*"}" = "$aurawin" -a \
+            if [ -n "$xiwiactive" -a "$aurawin" != "${aurawin%${curdisp%"*"}*}" -a \
```
dnschneid: Do you need additional quoting here? `"${aurawin%"${curdisp%"*"}"*}"`

dnschneid: Also, please keep to 80 characters.  Using disp instead of curdisp and moving the -a to the next line should fix that.

chroot-bin/croutoncycle
```diff
@@ -279,7 +279,7 @@ if [ "${destdisp#:}" = "$destdisp" ]; then
         croutonwmtools raise "${destdisp%"*"}"
     fi
 
-    if [ -n "$xiwiactive" -a "${destdisp%"*"}" = "$aurawin" ]; then
+    if [ -n "$xiwiactive" -a "$aurawin" != "${aurawin%${destdisp%"*"}*}" ]; then
```
dnschneid: Ditto on quoting

chroot-bin/croutoncycle
```diff
@@ -103,7 +103,7 @@ if host-x11 true 2>/dev/null; then
     winlist="`host-x11 croutonwmtools list nim | \
               sort | awk '{ printf $NF " " }'`"
     aurawin="`host-x11 croutonwmtools list ni | \
-              awk '$1 == "aura_root_0" { print $NF; exit }'`"
+              awk '$1 ~ /aura_root_/ { printf $NF " " }'`"
```
dnschneid: Be a little more restrictive on that regex.  `/^aura_root_[0-9]+$/` perhaps.

chroot-bin/croutoncycle
dnschneid: Actually, i'm worried about substring matches here.  How about making the aurawin list with a guaranteed space on either side of each entry, and then including the guard spaces in the suffix remove operations?

dnschneid:
I think this is obsolete, since xephyr has been removed.

PR #1540, opened 2015-03-26
	croutontriggerd: Support use of right-{ctrl,shift,alt} in trigger.

mkasick:
The old xbindkeys trigger used to support using both left and right ctrl/shift.  I prefer to use right-ctrl and right-shift as I find it easier to enter the sequence that way.  This commit restores support for left and right ctrl/shift in croutontriggerd.

### DIFF
--- a/chroot-bin/croutontriggerd
+++ b/chroot-bin/croutontriggerd
@@ -28,6 +28,9 @@ STATE_UP=0
 KEY_LEFTCTRL=29
 KEY_LEFTALT=56
 KEY_LEFTSHIFT=42
+KEY_RIGHTCTRL=97
+KEY_RIGHTALT=100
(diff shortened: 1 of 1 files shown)


dnschneid:
No love for AltGr?

mkasick:
I checked before making this, and AltGr was actually not supported by xbindkeys either.  Although that might be due to the default pc105 keyboard setup in my chroots.

Would you prefer I add that too?

dnschneid:
Yeah, it'd be nice.  If shift and ctrl work, the right-alt key on Chromebook keyboards should work too.

mkasick:
OK.  I've updated the patch and tested on my Chromebox.  All three of the right-ctrl, shift, and alt now work.

dnschneid:
Awesome.  Thanks!

PR #1542, opened 2015-03-26
Add croutonversion

DennisLfromGA:
Add croutonversion for troubleshooting.

### DIFF
--- a/chroot-bin/croutonxinitrc-wrapper
+++ b/chroot-bin/croutonxinitrc-wrapper
@@ -43,6 +43,10 @@ else
 fi
 
 # Run crouton-specific commands:
+
+# Show chroot specifics for troubleshooting
+croutonversion
+
 if [ -z "$XMETHOD" ]; then
     if [ -f '/etc/crouton/xmethod' ]; then
         read -r XMETHOD _ < /etc/crouton/xmethod


dnschneid:
Thanks!

DennisLfromGA:
Thank you.

BTW - Using i3xiwi now with my lxde desktop - nice!

On Fri, Mar 27, 2015 at 10:50 PM, David Schneider <notifications@github.com>
wrote:

> Thanks!
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dnschneid/crouton/pull/1542#issuecomment-87151052>.
>



-- 
*DennyL@GMail*

DennisLfromGA:
Question:
To get the latest and greatest crouton version and keep the i3xiwi branch can I do the following? 

`sudo sh ~/Downloads/crouton -u 1-trusty`

followed by:

`sudo CROUTON_BRANCH=i3xiwi sh ~/Downloads/crouton -u 1-trusty`

dnschneid:
I'm not releasing compiled i3xiwis, so that second command won't work.  I've just rebased i3xiwi, so if you update using the main.sh in this branch, you'll be fine.

DennisLfromGA:
I'm sorry but I don't understand at all :(

> I'm not releasing compiled i3xiwis, so that second command won't work.

What are *compiled i3xiwis* ?

The 2nd command is what I used to install the i3xiwi branch, was that the wrong way?

> I've just rebased i3xiwi, so if you update using the main.sh in this branch, you'll be fine.

Totally lost here, do you mean I just need to do the 1st command without specifying a branch?

Please feel free to ignore these noob questions, I know you're busy and I don't want to slow you down. I'm afraid I know just enough to be dangerous ;(

dnschneid:
If you don't see your branch listed here: https://github.com/dnschneid/crouton/tree/releases, then it means CROUTON_BRANCH won't work for it.  I'm not sure what happened when you ran that command, but I doubt it successfully updated the chroot.

If you want to try an arbitrary branch, you have to do the following:

1. Go to the branch page (on the fork page, select the correct branch)
2. There's a "Download ZIP" button on the right side. Right-click it and copy the URL to the clipboard
3. Change the .zip to a .tar.gz and download the file
4. In ~/Downloads, untar the tarball
5. Use the untarred installer/main.sh as if it were crouton.

I can elaborate on the steps if need be.

DennisLfromGA:
Okay, I thought I could install any *active* crouton branch by using the CROUTON_BRANCH option with crouton, evidently I misunderstood.

Thankfully, you don't need to elaborate, that part I get and have done before.

One last dumb question, I hope - how can I tell if I'm using the new i3xiwi branch?

dnschneid:
`croutonversion` will return "git"...otherwise, there's really no signal inside the chroot.  You can always update to make sure...

PR #1546, opened 2015-03-27
Use I3 for xiwi instead of ratpoison

### LINKED ISSUE #1483: Xiwi and Java GUI applications
Neither Eclipse, nor JetBrains IDEA are working in xiwi window (launched as `xiwi idea` inside chroot). The window is blank, but when I click on some hidden controls, some popups or drop downs are working.

I'm using trusty release.

dnschneid:
Improves floating window behavior and compatibility with software that requires a reparenting window manager.

Window title monitoring is also more efficient.

Fixes #1483

Colors are also set to look nice with stock Chromium OS.

### DIFF
--- a/chroot-etc/xiwi.conf
+++ b/chroot-etc/xiwi.conf
@@ -1,0 +1,27 @@
+# i3 config file (v4)
+# Copyright (c) 2014 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# Style
+new_window none
+new_float normal
+workspace_layout tabbed
+font pango:Sans 8
+
+# Colors                border  backgr. text    indicator
+client.focused          #8E8E8F #EAEAEB #000000 #8E8E8F
+client.focused_inactive #8E8E8F #CACACB #525252 #8E8E8F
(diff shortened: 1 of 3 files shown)


fczuardi:
I haven't tested this version yet, will ```xiwi i3``` still work with this?

dnschneid:
What are you trying to do?  If you just want to launch an i3 environment, you're better off configuring your `.xinitrc` and using xinit like on a normal system.

The I3 config used by xiwi will not touch your `.i3/config`.

DennisLfromGA:
I'm running Popcorn-Time that kicks-off in a background process using the i3xiwi branch!

dnschneid:
Merge first, ask forgiveness/apply @drinkcat's suggestions later :shipit:

chroot-etc/xiwi.conf
```diff
@@ -0,0 +1,27 @@
+# i3 config file (v4)
+# Copyright (c) 2014 The crouton Authors. All rights reserved.
```
drinkcat: 2015 ,-)

dnschneid:
I think you can put i3 commands inside of the .xiwirc file, so you can effectively already do that.

Issue #407, opened 2013-10-04
Need to micromanage brightness on resume and VT switch


Mikroprocessor:
Hi!

I have Ubuntu 13.04 on my Acer C7 Chromebook.
Every time i lay down the screen(as a lid on a jar, but not screwing down it) and wait until the laptop has gone to sleep-mode. Then i open the screen again the screen is black. If then i plug in (or plug out) the charger plug(not it plug who fits in the wall, the plug who fits in the round hole on the laptop), then the screen wakes up and i can you the laptop again.

I don't know why it's on that way, but im sure that is a bug. Right?

Med vänliga hälsningar - Best regards
Sweden.

ccaapton:
@dnschneid I'm also experiencing ChromeOS misbehaving, but not crouton using awesomewm. Is there any workaround to avoid reboot? Thx!

midnightradio:
I'm having the same issue on my pixel. It reboots on opening the lid occasionally. At first I thought it's only happening for running xfce and started to keep doing some jobs with just entering chroot without starting the desktop manager. But still it happened even if I didn't run xfce. So I guess it's not a matter of xscreensaver / power manager or desktop environment. Also it keeps happening even when I changed focus to a tab other than the tab for chroot before closing the lid. 

I can close the lid only after saving all file I was working on. :-(

dnschneid:
@midnightradio could you please file a separate issue for that?

dnschneid:
Are people still hitting this bug?

PR #1551, opened 2015-03-28
Nudge display brightness on return to cros

ttjoseph:
For some targets, when croutoncycling back to cros, the display brightness is not restored correctly - the host cros dbus server knows the correct value but the hardware does not (observed on Toshiba B3340, utopic chroot). This admittedly hack-y change nudges the the display brightness down then up so that the old correct value is restored.

### DIFF
--- a/chroot-bin/brightness
+++ b/chroot-bin/brightness
@@ -1,11 +1,11 @@
 #!/bin/sh -e
-# Copyright (c) 2014 The crouton Authors. All rights reserved.
+# Copyright (c) 2015 The crouton Authors. All rights reserved.
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 
 # A shortcut script to control device brightness.
 # Usage:
-#    brightness [b|k] [up|down|# [instant]]
+#    brightness [b|k] [up|down|get-untruncated|#]
 
 # Choose the device
 device='Screen'
@@ -25,7 +25,8 @@ u*) precmd='Increase';;
 d*) precmd='Decrease';;
 [0-9]*) eval $nokbd;
     precmd='Set'; postcmd="Percent double:$1 int32:${2:-"1"}${2:+"2"}";;
-*) eval $nokbd; precmd='Get'; postcmd='Percent'; print='--print-reply';;
+g*) eval $nokbd; precmd='Get'; postcmd='Percent'; print='--print-reply';;
+*) eval $nokbd; precmd='Get'; postcmd='Percent'; print='--print-reply'; truncate=true;;
 esac
 
 host-dbus dbus-send --system --dest=org.chromium.PowerManager \
@@ -34,7 +35,11 @@ host-dbus dbus-send --system --dest=org.chromium.PowerManager \
     read -r junk
     read -r double percent
     if [ -n "$print" ]; then
-        echo "${percent%.*}"
+        if [ -n "$truncate" ]; then
+            echo "${percent%.*}"
+        else
+	    echo "${percent}"
+        fi
     fi
 }
 
--- a/chroot-bin/croutoncycle
+++ b/chroot-bin/croutoncycle
@@ -56,6 +56,36 @@ getname() {
     fi
 }
 
+# Local version of brightness command for use with nudgebrightness below
+mybrightness() {
+    local postcmd=''
+    local precmd=''
+    local percent=''
+    local junk=''
+    case "$1" in
+    [0-9]*) precmd='Set'; postcmd="Percent double:$1 int32:${2:-"1"}${2:+"2"}";;
+    *) precmd='Get'; postcmd='Percent'; print='--print-reply';;
+    esac
+
+    host-dbus dbus-send --system --dest=org.chromium.PowerManager \
+        --type=method_call $print /org/chromium/PowerManager \
+        org.chromium.PowerManager.${precmd}ScreenBrightness${postcmd} | {
+            read -r junk
+            read -r double percent
+            echo "${percent}"
+        }
+}
+
+# Nudges brightness down then restores old value, because when chvt-ing back
+# to cros, the brightness is not always correctly restored (host dbus knows
+# correct value but hardware appears to need to be poked)
+nudgebrightness() {
+    local oldbrightness=`mybrightness get`
+    mybrightness $((${oldbrightness%.*} - 1)) &
+    sleep 0.2
+    mybrightness $oldbrightness &
+}
+
 # Only let one instance run at a time to avoid nasty race conditions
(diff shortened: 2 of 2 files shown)


dnschneid:
This is relevant to #407.
What about using the brightness command to get the current percent, then either re-apply it or jump between two values?  Down then up won't work at minimum brightness, for instance (well, it will, but the end result will be different).

ttjoseph:
Ah, good point. The brightness command truncates the brightness value it returns - on the assumption that there's a good reason for that, I've left it alone and brought that functionality, without rounding, into croutoncycle, then implemented your suggestion.

dnschneid:
Actually, I'd rather you change brightness to have a flag that chooses truncated/untruncated than duplicate the code (probably default to truncated, and add -f to disable truncate or something).  The brightness command is actually very poorly-written (and the "instant" thing doesn't actually work)...

ttjoseph:
Here's a cleaner version of the brightness fix, complete with a flag to the brightness command rather than a duplicated implementation.

chroot-bin/brightness
```diff
@@ -25,7 +25,8 @@ u*) precmd='Increase';;
 d*) precmd='Decrease';;
 [0-9]*) eval $nokbd;
     precmd='Set'; postcmd="Percent double:$1 int32:${2:-"1"}${2:+"2"}";;
-*) eval $nokbd; precmd='Get'; postcmd='Percent'; print='--print-reply';;
+g*) eval $nokbd; precmd='Get'; postcmd='Percent'; print='--print-reply';;
+*) eval $nokbd; precmd='Get'; postcmd='Percent'; print='--print-reply'; truncate=true;;
```
dnschneid: truncate should be defined to be empty above the case

chroot-bin/brightness
```diff
@@ -34,7 +35,11 @@ host-dbus dbus-send --system --dest=org.chromium.PowerManager \
     read -r junk
     read -r double percent
     if [ -n "$print" ]; then
-        echo "${percent%.*}"
+        if [ -n "$truncate" ]; then
+            echo "${percent%.*}"
+        else
+	    echo "${percent}"
```
dnschneid: indentation

chroot-bin/croutoncycle
```diff
@@ -56,6 +56,16 @@ getname() {
     fi
 }
 
+# Nudges brightness down then restores old value, because when
+# chvt-ing back to cros, the brightness is not always correctly restored
+# (host dbus knows the correct value but hardware must be poked)
+nudgebrightness() {
+  local oldbrightness=`brightness get-untruncated`
+  brightness down &
```
dnschneid: Don't fork here; wait for the commands to return.  Also, I suggest changing the brightness by a smaller offset than the default, so it's less obvious.  You'll also want to change it upwards if the volume is already at the minimum.  I suggest something like if oldbrightness < 50, nudge up. otherwise, nudge down.

chroot-bin/croutoncycle
```diff
@@ -88,7 +98,7 @@ for disp in /tmp/.X*-lock; do
             | grep -q 'INTEGER'; then
         displist="$displist $disp"
     elif DISPLAY="$disp" xprop -root 'CROUTON_XMETHOD' 2>/dev/null \
-            | grep -q '= "xiwi"$'; then
+            | grep -q '= "xiwi'; then
```
dnschneid: Not sure where this change came from.  Can you rebase off of the latest master?

chroot-bin/croutoncycle
```diff
@@ -56,6 +56,16 @@ getname() {
     fi
 }
 
+# Nudges brightness down then restores old value, because when
+# chvt-ing back to cros, the brightness is not always correctly restored
+# (host dbus knows the correct value but hardware must be poked)
+nudgebrightness() {
+  local oldbrightness=`brightness get-untruncated`
+  brightness down &
+  sleep 0.2
```
dnschneid: Sleep shouldn't be necessary once you remove the forking

chroot-bin/croutoncycle
```diff
@@ -289,7 +299,7 @@ if [ "${destdisp#:}" = "$destdisp" ]; then
         # Raise the right window after chvting, so that it can update
         if [ "$tty" != 'tty1' ]; then
             sudo -n chvt 1
-            sleep .1
```
dnschneid: I suggest keeping the sleep in there when chvting, otherwise the brightness might get reverted again.

chroot-bin/croutoncycle
```diff
@@ -304,17 +314,19 @@ if [ "${destdisp#:}" = "$destdisp" ]; then
...
-    if xprop -root 'CROUTON_XMETHOD' 2>/dev/null | grep -q '= "xiwi"$'; then
+    xmethod="`xprop -root 'CROUTON_XMETHOD' 2>/dev/null \
+              | sed -n 's/^.*\"\(.*\)\"/\1/p'`"
+    if [ "${xmethod%%-*}" = 'xiwi' ]; then
         if [ -z "$freonowner" -a "$tty" != 'tty1' ]; then
             sudo -n chvt 1
-            sleep .1
+            nudgebrightness
         elif [ "${freonowner:-0}" != 0 ]; then
             kill -USR1 "$freonowner"
```
dnschneid: We want to nudge the brightness here, too

chroot-bin/croutoncycle
```diff
@@ -289,7 +299,7 @@ if [ "${destdisp#:}" = "$destdisp" ]; then
         # Raise the right window after chvting, so that it can update
         if [ "$tty" != 'tty1' ]; then
             sudo -n chvt 1
-            sleep .1
+            nudgebrightness
         fi
         croutonwmtools raise "${destdisp%"*"}"
     elif [ "${freonowner:-0}" != 0 ]; then
```
dnschneid: Freon has the same issue, so nudge the brightness here, too

chroot-bin/croutoncycle
```diff
@@ -56,6 +56,16 @@ getname() {
     fi
 }
 
+# Nudges brightness down then restores old value, because when
+# chvt-ing back to cros, the brightness is not always correctly restored
+# (host dbus knows the correct value but hardware must be poked)
+nudgebrightness() {
+  local oldbrightness=`brightness get-untruncated`
```
dnschneid: Indentation

dnschneid:
Not sure if people are still hitting this bug.  If they are, this needs to be rebased and corrected.

ttjoseph:
No longer hitting this bug. I'd suggest just tossing this PR altogether. Thanks very much for your help and followup.

dnschneid:
Oh, okay!  Thanks again for submitting a fix.

PR #1568, opened 2015-04-01
xiwi apps in a window and in a tab!

### LINKED ISSUE #1453: Launch xiwi app in a small window
As implemented in PR #1442 , the xiwi app windows are launched into full-screen mode, just like the launching the DEs. It would be better if the xiwi apps can be launched in the "default" resolution as they would do inside a normal desktop environment to provide better experience.

dnschneid:
Implements both windowed- and tab launching of chroots. xiwi will default to windowed, xiwi -F will go full-screen, and xiwi -T will go tabbed (using uppercase to avoid conflicts with enter-chroot when we make a startxiwi script down the road).

It's a bit hard to set up to test, as you have to build and run the updated extension along with the code.  If you CAN test though, please do (and report back), because there's quite a lot changed.

TODO: I need to change the order of the info in the title when you're in a tab (chroot+application AFTER titlebar instead of before), because the tab is too small to give useful information.

Resolves #1453

### DIFF
--- a/chroot-bin/xiwi
+++ b/chroot-bin/xiwi
@@ -109,5 +109,9 @@ else
         \?) error 2 "$USAGE";;
         esac
     done
+    eval "exe=\"\$$OPTIND\""
+    if ! hash "$exe" 2>/dev/null; then
+        error 2 "${0##*/}: $exe: not found"
+    fi
     exec /usr/local/bin/xinit "$xiwicmd" / "$@"
 fi
--- a/chroot-bin/xiwi
+++ b/chroot-bin/xiwi
@@ -61,17 +61,16 @@ elif [ "$1" = '/' ]; then
                     kill "$monpid" 2>/dev/null
                 fi
                 monwid="$wid"
-                xprop -spy -notype -id "$wid" 'WM_NAME' 2>/dev/null | {
-                    while read _ title; do
+                (xprop -spy -notype -id "$wid" 'WM_NAME' 2>/dev/null || echo) \
+                    | while read _ title; do
                         title="${title%\"}"
                         xprop -root -f CROUTON_NAME 8s -set CROUTON_NAME \
-                            "$name/$1: ${title#*\"}"
+                            "$name/$1${title:+": "}${title#*\"}"
                         {
                             echo -n 'C'
                             croutoncycle l
                         } | websocketcommand >/dev/null
-                    done
-                } &
+                    done &
                 monpid="$!"
             done
             if [ -n "$monpid" ]; then
--- a/chroot-bin/xiwi
+++ b/chroot-bin/xiwi
@@ -5,13 +5,12 @@
 
 # Runs the specified X11 application in its own X server in Chromium OS.
 
-. "`dirname "$0"`/../installer/functions"
-xiwicmd="`readlink -f "$0"`"
-
-if [ "$#" = 0 ]; then
-    echo "Usage: ${0##*/} [-f] APPLICATION [PARAMETERS ...]
+USAGE="Usage: ${0##*/} [-f] [-F|-T] APPLICATION [PARAMETERS ...]
 Launches a windowed session in Chromium OS for any graphical application.
 All parameters are passed to the specified application.
+
+By default, the app is launched in a window. Passing -F will launch the app
+full-screen, and passing -T will launch the app in a tab.
 
 xiwi will normally close when the application returns. Some gui applications
 fork before or during normal operation, which can confuse xiwi and cause it to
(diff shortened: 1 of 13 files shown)


chroot-bin/xiwi
```diff
@@ -20,20 +19,30 @@ it from forking, you can use -f to prevent xiwi from quitting automatically.
...
-manager, specify the full path of the application." 1>&2
-    exit 2
+with 'start' or is 'xinit'. You can cycle through multiple windows inside the
+application via Ctrl-Alt-Tab/Ctrl-Alt-Shift-Tab, or close them via
+Ctrl-Alt-Shift-Escape.  If APPLICATION begins with 'start' but you still want to
+use the default window manager, specify the full path of the application."
+
+. "`dirname "$0"`/../installer/functions"
+xiwicmd="`readlink -f "$0"`"
+OPTSTRING='FfTt'
```
drinkcat: Do we really want to support `-t`?

host-bin/enter-chroot
```diff
@@ -138,8 +138,8 @@ fi
 
 # Check to ensure that the XMETHOD requested has been installed
 if [ -n "$TMPXMETHOD" ]; then
-    if ! grep -q "^$TMPXMETHOD$" "$CHROOTS/$NAME/.crouton-targets" 2>/dev/null; then
-        error 1 "$CHROOTS/$NAME does not contain XMETHOD '$TMPXMETHOD'"
+    if ! grep -q "^$TMPXMETHOD\(-\|\$\)" "$CHROOTS/$NAME/.crouton-targets"; then
```
drinkcat: Shouldn't this be `grep -q "^${TMPXMETHOD%%-*}$" ...`?

host-ext/crouton/background.js
```diff
@@ -491,27 +517,38 @@ function websocketMessage(evt) {
...
         display = payload
-        match = display.match(/^:([0-9]+)$/)
-        displaynum = match ? match[1] : null
+        match = display.match(/^:([0-9]+)([- ][^- ]*)*$/)
+        displaynum = match ? match[1] : null;
+        mode = null;
+        if (displaynum) {
+            display = ":" + displaynum;
+            mode = match[2].length >= 2 ? match[2].charAt(1) : 'f';
+            if ('fwt'.indexOf(mode) == -1) {
```
drinkcat: Brilliant. ,-)

host-ext/crouton/background.js
```diff
@@ -520,32 +557,49 @@ function websocketMessage(evt) {
...
+                chrome.windows.get(winid, function(win) {
+                    if (win.state == "maximized")
+                        chrome.windows.update(winid, {state: 'fullscreen'});
+                });
+            }
         } else {
             /* Open a new window */
             kiwi_win_[display] = new Object();
             kiwi_win_[display].id = -1;
+            kiwi_win_[display].isTab = mode == 't';
```
drinkcat: `(mode == 't')` ?

host-ext/crouton/background.js
```diff
@@ -589,13 +643,14 @@ function websocketClose() {
     checkUpdate(false);
 }
 
-/* Called when window in focus changes: feeback to the extension so the
+/* Called when window/tab in focus changes: feeback to the extension so the
```
drinkcat: feedback

host-ext/crouton/window.js
```diff
@@ -262,8 +269,10 @@ function handleFocusBlur(evt) {
 }
 
 /* Start in full screen */
-chrome.windows.update(chrome.windows.WINDOW_ID_CURRENT,
-                      {'state': "fullscreen"}, function(win) {})
+if (location.search.search(/[&?]mode=f(&|$)/i) != -1) {
```
drinkcat: Wouldn't it be better to do the argument parsing a bit earlier then? (the parameter extraction code and global variable assignments can be moved here, and force a `setHiDPI` and `setDebug` in the listener below)

chroot-bin/xiwi
dnschneid: I was just being nice :(

host-bin/enter-chroot
dnschneid: oops, yes.

host-ext/crouton/background.js
dnschneid: In my defense I inherited that one, and actually fixed it once before resetting the file accidentally :P

host-ext/crouton/window.js
dnschneid: Yeah, I'll do that.

Also, I just read [[http://speakingjs.com/es5/ch16.html]]...I suspect we have a lot of unintended global variables in the code...

elwebmaster:
So how do we test this?

SilverWave:
WANT.

This has pushed me to start using xiwi and chromium rather than sitting in xfce.

Opening the chroot in an actual tab, not a window, would be the holy grail.

SHIPITNOW

LOL

dnschneid:
Roger wilco!

DennisLfromGA:
'criat' is COOL !
Right now, we have to use `sudo enter-chroot xiwi -h` (or whatever) to see all of xiwi's options.
Could that USAGE message be added in xiwi TIPS possibly? I know that's not exactly the right place... 

Thanx a bunch for this :)

screenshot 2015-04-04 at 11 34 26 pm

dnschneid:
Ooh, I forgot about adding to TIPS.  I don't really want to make it much longer though, so if you can propose a replacement for the second paragraph that doesn't exceed ~5-6 lines at 80 chars width, I'd happily bring it in.

DennisLfromGA:
@dnschneid -
Which second paragraph - 1 or 2 ?
( I'm assuming 2 ;)
```
    USAGE="Usage: ${0##*/} [-f] [-F|-T] APPLICATION [PARAMETERS ...]
    Launches a windowed session in Chromium OS for any graphical application.
    All parameters are passed to the specified application.  

1>  By default, the app is launched in a window. Passing -F will launch the app
    full-screen, and passing -T will launch the app in a tab.  

2>  xiwi will normally close when the application returns. Some gui applications
    fork before or during normal operation, which can confuse xiwi and cause it to
    quit prematurely. If your application does not have a parameter that prevents
    it from forking, and crouton is unable to automatically detect the fork, you can
    use -f to prevent xiwi from quitting automatically.
    xiwi will quit if you close the Chromium OS window when nothing is displayed.  

    A default window manager will full-screen all windows, unless APPLICATION begins
    with 'start' or is 'xinit'. You can cycle through multiple windows inside the
    application via Ctrl-Alt-Tab/Ctrl-Alt-Shift-Tab, or close them via
    Ctrl-Alt-Shift-Escape.  If APPLICATION begins with 'start' but you still want to
    use the default window manager, specify the full path of the application."
```

DennisLfromGA:
@dnschneid / @drinkcat,

I've got another, hopefully better, idea.

Leave the `enter-chroot` USAGE and `xiwi` USAGE & TIPS as is and display them as needed in a `startxiwi` host-bin script.

I've got the beginnings of one here - [startxiwi](https://github.com/DennisLfromGA/crouton/blob/master/host-bin/startxiwi).
It's kind of cobbled together but I think it supports the idea.

When you get a chance, please take a look and let me know what you think.

-DennisL

SilverWave:
Both of my use cases are covered, awesome.

Single apps work fine (e.g. pw manager or xterm).
sudo enter-chroot -n trusty2 xiwi -T xterm

Running Xfce in a tab (easy to find).
sudo enter-chroot -n trusty2 xiwi -T startxfce4

Cheers.

drinkcat:
Lagging behind... lgtm, just one small nit.

PR #1569, opened 2015-04-02
Add check for trusty, and install pulseaudio

### LINKED ISSUE #1550: gnome target requires pulseaudio for sound to work
Using a fresh download of crouton, I created a new chroot with release `trusty` and target `gnome`. Although playing sounds via `aplay some.wav` works, the sound in any gui application doesn't (skype/firefox). Also, in the sound settings 'Output' menu, where it says 'Choose a device for sound output', there aren't any devices listed. After installing pulseaudio, everything works, and `cras` is listed as a sound device.

This problem seems similar to the one reported in #1328 for the `lxde` target. I also tried `xfce`, and there the sounds works out of the box. The `xfce` target comes with pulseaudio preinstalled though.

I suggest installing `pulseaudio` for the `gnome` and `lxde` targets, or perhaps even for the `audio` target.

My machine is an ASUS C300M. Here's the output of `croutonversion`:

```
chronos@localhost / $ sudo enter-chroot -n trusty-gnome-basic croutonversion
Entering /mnt/stateful_partition/crouton/chroots/trusty-gnome-basic...
crouton: version 1-20150327195116~master:cf2aacba
release: trusty
architecture: amd64
xmethod: xorg
targets: gnome
host: version 6680.78.0 (Official Build) stable-channel quawks 
kernel: Linux localhost 3.10.18 #1 SMP Fri Mar 20 17:27:56 PDT 2015 x86_64 x86_64 x86_64 GNU/Linux
freon: no
```

Thank you for crouton! It's a lifesaver.

jamesjnadeau:
sound does not work without it and tracking down why is not easy.

This should fix #1550 

I have never set up a build env for this, so I have not a clue if what I changed works, I'm going to try that next

### DIFF
--- a/targets/gnome
+++ b/targets/gnome
@@ -13,6 +13,10 @@ install --minimal evolution-data-server gnome-control-center \
         gnome-screensaver gnome-session gnome-session-fallback \
         gnome-shell gnome-themes-standard gvfs-backends nautilus \
         unzip gtk2-engines-pixbuf
+if [ "${TARGETNOINSTALL:-c}" = 'c' ] &&
+        [ "$DISTRO" = 'ubuntu' ] && release = trusty; then
+        install --minimal pulseaudio
+fi
 
 TIPS="$TIPS
 You can start GNOME via the startgnome host command: sudo startgnome


dnschneid:
You can just add pulseaudio to the list of packages in the install command above.

Are you sure this is the only fix for #1550 though?  Is there a strong dependency somewhere?

jamesjnadeau:
From my understanding, trusty creates the requirement for pulseaudio, I'm not sure if other versions require this, but having it might mess up other instances. 

My thought process was only doing it in this instance would reduce problems for others I have not tested. I'm open to putting it in with the main install command, I was just not sure of the side effects and I'm overly cautious. 

I still haven't gotten around to testing with my own build environment yet..... Hopefully this weekend I'll get a chance and I'll see what pulseaudio does to different targets. I'll report back here.

smibarber:
Included pulseaudio via #2818

Issue #1399, opened 2015-02-05
Advanced Encryption Options


Timvrakas:
Firstly, i'm absolutely loving my new Linux net book :) Thanks people!

I have an idea for a new method of encryption.
I think it seems rather silly to have to type in my extraneously long password no less than 3 times to get into the chroot, and then a 4th if I need sudo. The machine takes a few seconds to boot, but logging in takes almost a minute (Perhaps I'm spoiled by chrome :) ).
Ok, so I have a new system (or a plan at least)
1. Encrypted chroot files on external partition (yay thanks drinkcat!)
2. Un-encrypted private key for chroot in user statefull partition (encrypted by TPM/Google Login)
3. Autostart on Login script (possibly a program that listens for a key-press to start crouton)

In addition, if you are super crazy you could not include the encrypted key on encrypted, only backed up, but then you lose the developer-mode-wipe-protection.

I'm going to look into it but I expect to have trouble at some point.
Thanks

dnschneid:
I think the key here (no pun intended) is to allow for no passphrase if -k is specified, and for a way to remove the passphrase with edit-chroot if the keyfile is separate.  Definitely worth implementing; feel free to ask questions here or on #crouton-devel as you start digging into it.

Timvrakas:
Thanks. As soon as I back up Ill get to work. 
Do you have any kind of timeframe on merging drinkcat's seperate-partition with the latest version? even in an experimental branch would be nice. I'm running that now but its severely out of date.

dnschneid:
It'll be a while.  separate-partition is really invasive and requires a lot of testing, and there are a lot of other things to get working (like GPU acceleration on Freon)...
If you post on the PR asking for a rebase he might find the chance to do it for you, but I doubt it'll get much testing.

Timvrakas:
I just rebased it, it wasn't too hard. As soon as I test it I'll ask him to
pull it into his repo. Then I'll start on the keys.

Timvrakas:
I'm confused about what this block of code in mount-chroot does

    if [ -z "$KEYFILE" ]; then
            KEYFILE="$CHROOTSRC/.ecryptfs"
            if [ -f "$KEYFILE" ]; then
                header="`head -n1 "$KEYFILE"`"
                if [ -n "$header" ]; then
                    KEYFILE="$header"
                fi
            fi
        elif [ "${KEYFILE#/}" = "$KEYFILE" ]; then
[...]

PR #1585, opened 2015-04-06
Unencrypted keys in user directory for single-password access (from #1399)

Timvrakas:
NOTE: Continuation of #1399

OK here it is!

Adds new encryption system that uses unencrypted keys the the encrypted user partition, to allow for easier encrypted login. `edit-chroot -u` or `mount-chroot -u` will create the keys, and they will be detected from then on.

I think edit-chroot and mount-chroot are good to go. I tested them on my C720
I am unsure if (or how) I should edit main.sh and add an option to set this up on creation of a chroot, for now you have to run edit-chroot -u or mount-chroot -u to create the keys.

I'm still new to all this (crouton, git, bash, life) so feel free to offer comments, questions, concerns, criticisms, critiques, complaints.

Thanks!

### DIFF
--- a/host-bin/mount-chroot
+++ b/host-bin/mount-chroot
@@ -63,7 +63,7 @@ addtrap "stty echo 2>/dev/null"
 
 # Function to prompt the user for a passphrase. Sets $passphrase.
 promptNewPassphrase() {
-    NOPASS=''
+    local NOPASS=''
     if [ "$KEYFILE" != "$CHROOTSRC/.ecryptfs" ]; then
         ENABLENOPASS='y'
     fi
--- a/host-bin/mount-chroot
+++ b/host-bin/mount-chroot
@@ -13,7 +13,6 @@ ENABLENOPASS=''
 ENCRYPT=''
 KEYFILE=''
 PRINT=''
-NOPASS=''
 NOENC=''
 ROOT="`readlink -m '/var/run/crouton'`"
 MOUNTOPTS='rw,dev,exec,suid'
@@ -65,7 +64,7 @@ addtrap "stty echo 2>/dev/null"
 # Function to prompt the user for a passphrase. Sets $passphrase.
 promptNewPassphrase() {
     NOPASS=''
-    if [ ! "$CHROOTSRC/.ecryptfs" = "$KEYFILE" ]; then
+    if [ "$KEYFILE" != "$CHROOTSRC/.ecryptfs" ]; then
         ENABLENOPASS='y'
     fi
     while [ -z "$passphrase" -a -z "$NOPASS" ]; do
@@ -73,7 +72,7 @@ promptNewPassphrase() {
         echo_tty -n "Choose an encryption passphrase for $NAME: "
         read -r passphrase
         echo_tty ''
-        echo -n 'Please confirm your passphrase: ' 1>&2
+        echo_tty -n 'Please confirm your passphrase: '
         read -r confirmation
         echo_tty ''
         if [ ! "$confirmation" = "$passphrase" ]; then
@@ -92,13 +91,13 @@ Do you want to continue? [y/N]: '
                 if [ "${response#[Yy]}" = "$response" ]; then
                     continue
                 else
-                NOPASS='y'
-                echo_tty 'No password set, keys will not be encrypted'
+                    NOPASS='y'
+                    echo_tty 'No password set, keys will not be encrypted'
                 fi
             else
                 echo_tty \
-'You tried to specify no passphrase, but to do that,
-you must store you keyfile in a custom location with -k.'
+'Passphrases cannot be empty unless the keyfile
+is stored separately from the chroot using -k.'
                 continue
             fi
         fi
@@ -279,7 +278,7 @@ $passphrase" | ecryptfs-wrap-passphrase "$wrappedfnek" -
                     echo | cat - "$wrappedkey" "$wrappedfnek" > "$KEYFILE"
                     unset key fnek
                 elif [ -z "$NOENC" -a -n "$NOPASS" ]; then
-                    #Unwrap previously wrapped passphrase
+                    # Unwrap previously wrapped passphrase
                     key="`echo -n "$oldpassphrase" \
                         | ecryptfs-unwrap-passphrase "$wrappedkey" - 2>/dev/null`"
                     fnek="`echo -n "$oldpassphrase" \
--- a/host-bin/mount-chroot
+++ b/host-bin/mount-chroot
@@ -9,11 +9,15 @@ APPLICATION="${0##*/}"
 BINDIR="`dirname "\`readlink -f "$0"\`"`"
 CHROOTS="`readlink -m "$BINDIR/../chroots"`"
 CREATE=''
+ENABLENOPASS=''
 ENCRYPT=''
 KEYFILE=''
 PRINT=''
+NOPASS=''
+NOENC=''
 ROOT="`readlink -m '/var/run/crouton'`"
 MOUNTOPTS='rw,dev,exec,suid'
+PAD='................'
 
 USAGE="$APPLICATION [options] name [...]
 
@@ -60,27 +64,46 @@ addtrap "stty echo 2>/dev/null"
 
 # Function to prompt the user for a passphrase. Sets $passphrase.
 promptNewPassphrase() {
-    echo_tty -n "Choose an encryption passphrase for $NAME: "
-    [ -t 0 ] && stty -echo
-    while [ -z "$passphrase" ]; do
+    NOPASS=''
+    if [ ! "$CHROOTSRC/.ecryptfs" = "$KEYFILE" ]; then
+        ENABLENOPASS='y'
+    fi
+    while [ -z "$passphrase" -a -z "$NOPASS" ]; do
+        [ -t 0 ] && stty -echo
+        echo_tty -n "Choose an encryption passphrase for $NAME: "
         read -r passphrase
-        if [ -z "$passphrase" ]; then
-            echo_tty ''
-            echo_tty -n 'You must specify a passphrase: '
-            continue
-        fi
         echo_tty ''
-        echo_tty -n 'Please confirm your passphrase: '
+        echo -n 'Please confirm your passphrase: ' 1>&2
         read -r confirmation
+        echo_tty ''
         if [ ! "$confirmation" = "$passphrase" ]; then
             passphrase=''
-            echo_tty ''
-            echo_tty -n 'Passphrases do not match; try again: '
+            echo_tty -n 'Passphrases do not match.'
+            continue
+        fi
+        [ -t 0 ] && stty echo
+        if [ -z "$passphrase" ]; then
+            if [ -n "$ENABLENOPASS" ]; then
+                echo_tty -n \
+'You did not specify a passphrase.  Your keys will not be encrypted.
+Only do this if you are confident that your keys are stored in a secure location.
+Do you want to continue? [y/N]: '
+                read -r response
+                if [ "${response#[Yy]}" = "$response" ]; then
+                    continue
+                else
+                NOPASS='y'
+                echo_tty 'No password set, keys will not be encrypted'
+                fi
+            else
+                echo_tty \
+'You tried to specify no passphrase, but to do that,
+you must store you keyfile in a custom location with -k.'
+                continue
+            fi
         fi
         confirmation=''
     done
-    [ -t 0 ] && stty echo
-    echo_tty ''
 }
 
 # Mount each chroot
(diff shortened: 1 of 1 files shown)


dnschneid:
Thanks for working on this!

I don't think there needs to be a new parameter, though.  When you go to encrypt the chroot, and only if you specify the `-k` parameter, you should be allowed to leave the passphrase empty.  crouton should confirm that you want to leave the passphrase in the clear, and note that this should only be done if the keyfile is located in encrypted storage (such as your user profile) or on removable media.  When you go to mount the chroot, the chroot should automatically find the unwrapped key (as it already does for -k).

Timvrakas:
@dnschneid I will get on that. It shouldn't take so long now that i'm familiar with the code.

On an off topic git-related tangent, If I want to roll back my current commit, and start fresh (I think that would be easiest), should I `git revert` or `git reset --hard` and then `push --force`? I looked it up, but I'm unsure which is correct. 

Thanks

dnschneid:
If you want to start fresh and throw away the work, `git reset --hard origin/master` followed by `git push origin encryption --force-with-lease` will safely update github.

Timvrakas:
OK thanks for the tip.

The question I have with the -k parameter is that if .ecryptfs is a pointer for the moved keyfile, then the only location for the keys is in the user directory, which is lost if the user account is deleted. I would prefer to have the keys in 2 places, both encrypted in the folder of the chroot and decrypted in the user home (this is especially true with separate-partition).

Any suggestions for this?

dnschneid:
That's kind of the point of moving the key around.  You can manually back up your key somewhere else (and then point to it if need-be by passing -k to enter-chroot), but for security, crouton shouldn't be silently duplicating your keys in multiple places.

host-bin/mount-chroot
```diff
@@ -145,10 +150,28 @@ for NAME in "$@"; do
             fi
         fi
 
-        # Detect the key file
+        # Detect unencrypted keys
+        if [ -z "$UKEYFILE" ]; then
+            UKEYFILE="/home/chronos/user/crouton/$NAME.chkey"
```
tonyxue: Why make ```~/crouton``` visible? If it is for storing keys, I think ```~/.crouton``` will be more suitale. Just like ```~/.ssh```

Timvrakas:
Is that the sort of change you were looking for? I guess that makes more sense.
FYI the previous commit was moved to timvrakas/old-encryption
WARNING: The above not tested, I will not have access to my chrome book until later tomorrow.

host-bin/mount-chroot
```diff
@@ -64,11 +66,6 @@ promptNewPassphrase() {
     [ -t 0 ] && stty -echo
     while [ -z "$passphrase" ]; do
         read -r passphrase
-        if [ -z "$passphrase" ]; then
```
dnschneid: Actually, I think this should still be enforced if keyfile wasn't set by the user.

host-bin/mount-chroot
```diff
@@ -77,6 +74,19 @@ promptNewPassphrase() {
             echo 1>&2
             echo -n 'Passphrases do not match; try again: ' 1>&2
         fi
+        if [ -z "$passphrase" ]; then
+            echo 'You did not specify a passphrase! Your keys will not be encrypted.
+            You should only do this if you have specified to store the keys in your
```
dnschneid:
```
      echo -n \
'You did not specify a passphrase.  Your keys will not be encrypted.
Only do this if you are confident that your keys are stored in a secure location.
Do you want to continue? [y/N] ' 1>&2
```

host-bin/mount-chroot
```diff
@@ -77,6 +74,19 @@ promptNewPassphrase() {
...
+            encrypted user folder (/home/chronos/user/...) or on removable media.
+            Enter "Yes" to confirm, or "No" to set a passphrase' 1>&2
+            read -r response
+            echo 1>&2
+            if [ "$response" = "N*" -o "$response" = "n*" ]; then
```
dnschneid: Default to no: `if [ "${response#[Yy]" = "$response" ]; then`

host-bin/mount-chroot
```diff
@@ -185,29 +195,30 @@ for NAME in "$@"; do
...
-            echo -n "$fnek
+                echo -n "$fnek
 $passphrase" | ecryptfs-wrap-passphrase "$wrappedfnek" -
-            unset key fnek
-            echo | cat - "$wrappedkey" "$wrappedfnek" > "$KEYFILE"
+                unset key fnek
+                echo | cat - "$wrappedkey" "$wrappedfnek" > "$KEYFILE"
+            else
+                echo > "$KEYFILE"
+                echo -n '$$$$$$$$$$$$$$$$' >> "$KEYFILE"
```
dnschneid: Dollar signs in shell scripts scare me.  How about `.`'s instead?

host-bin/mount-chroot
```diff
@@ -185,29 +195,30 @@ for NAME in "$@"; do
...
+                echo -n "$key" >> "$KEYFILE"
+                echo -n "$fnek" >> "$KEYFILE"
+                echo -n '$$$$$$$$$$$$$$$$' >> "$KEYFILE"
```
dnschneid: The first 16 characters of a wrapped key are the hex signature, so it'd be good to consistently put the padding at the beginning of each key.

I do like this approach overall.

host-bin/mount-chroot
```diff
@@ -216,32 +227,81 @@ $passphrase" | ecryptfs-wrap-passphrase "$wrappedfnek" -
             tail -c 160 "$KEYFILE" | head -c 80 > "$wrappedkey"
             tail -c 80 "$KEYFILE" > "$wrappedfnek"
 
+            PAD1="`cat "$wrappedkey" | head -c 16`"
```
dnschneid: echo, not cat.  But you can also do:

`if [ "${wrappedkey#................}" != "$PAD1" -a "${wrappedfnek#................}" != "$wrappedfnek" ]; then`

dnschneid: Oops, nevermind, you're totally right.

But you CAN pass the file directly to head.

host-bin/mount-chroot
dnschneid: Also, can we combine this into one call to echo, so it's atomic?

host-bin/mount-chroot
```diff
@@ -216,32 +227,81 @@ $passphrase" | ecryptfs-wrap-passphrase "$wrappedfnek" -
...
+                        | ecryptfs-unwrap-passphrase "$wrappedkey" - 2>/dev/null`"
+                    fnek="`echo -n "$oldpassphrase" \
+                        | ecryptfs-unwrap-passphrase "$wrappedfnek" - 2>/dev/null`"
+                    echo > "$KEYFILE"
+                    echo -n '$$$$$$$$$$$$$$$$' >> "$KEYFILE"
+                    echo -n "$key" >> "$KEYFILE"
+                    echo -n "$fnek" >> "$KEYFILE"
+                    echo -n '$$$$$$$$$$$$$$$$' >> "$KEYFILE"
+                elif [ -n "$NOENC" -a -n "$NOPASS" ]; then
+                    echo "Changing from nothing to nothing? You must be high-security..."
```
dnschneid: Haha.  Alternatively: `echo "Applying double-ROT13 encryption to keys..."`

dnschneid:
But yes, this was pretty much how I was expecting it to behave (kudos for the very clean way of handling the unwrapped keys).

Timvrakas:
I still have a problem. They 63-char keys are coming back to haunt me. They are not properly handled. But It should be good otherwise.

host-bin/mount-chroot
```diff
@@ -60,22 +62,32 @@ addtrap "stty echo 2>/dev/null"
 
 # Function to prompt the user for a passphrase. Sets $passphrase.
 promptNewPassphrase() {
+    NOPASS=''
     echo -n "Choose an encryption passphrase for $NAME: " 1>&2
     [ -t 0 ] && stty -echo
-    while [ -z "$passphrase" ]; do
+    while [ -z "$passphrase" -a -z "$NOPASS" ]; do
         read -r passphrase
-        if [ -z "$passphrase" ]; then
```
Timvrakas: About this... I wasn't sure if the warning about secure location was enough, or the blank passphrase should be linked to the -k flag. Right now they are independent features. I get the secure location thing, but seeing as the crouton default is no encryption at all, it feels like overkill.

dnschneid: The point of the `-e` flag is to make your chroot install more secure.  If we allow blank passwords without further configuration, then people will assume that even with a blank passphrase, things are fine (they will ignore the later warnings).  Unless you see actual value in allowing co-located unencrypted passphrases, I'd rather disallow it to avoid any chance of a false sense of security.

host-bin/mount-chroot
```diff
@@ -185,29 +199,27 @@ for NAME in "$@"; do
...
-            echo -n "$fnek
+                echo -n "$fnek
 $passphrase" | ecryptfs-wrap-passphrase "$wrappedfnek" -
-            unset key fnek
-            echo | cat - "$wrappedkey" "$wrappedfnek" > "$KEYFILE"
+                unset key fnek
+                echo | cat - "$wrappedkey" "$wrappedfnek" > "$KEYFILE"
+            else
+                echo -n "
+................$key................$fnek" >> "$KEYFILE"
```
dnschneid: Truncate, not append

host-bin/mount-chroot
```diff
@@ -60,22 +62,34 @@ addtrap "stty echo 2>/dev/null"
...
+            echo 1>&2
+            echo -n \
+'You did not specify a passphrase.  Your keys will not be encrypted.
+Only do this if you are confident that your keys are stored in a secure location.
+Do you want to continue? [y/N] ' 1>&2
+            read -r response
+            echo 1>&2
+            if [ "${response#[Yy]}" = "$response" ]; then
+                continue
+			else
```
dnschneid: indentation

host-bin/mount-chroot
```diff
@@ -216,32 +228,78 @@ $passphrase" | ecryptfs-wrap-passphrase "$wrappedfnek" -
             tail -c 160 "$KEYFILE" | head -c 80 > "$wrappedkey"
             tail -c 80 "$KEYFILE" > "$wrappedfnek"
 
+            PAD1="`head -c 16 "$wrappedkey"`"
+            PAD2="`tail -c 16 "$wrappedfnek"`"
```
dnschneid: this now needs to be `head`...I would look at the rest of the key handling (especially below) to make sure it's up-to-date with the new ...xxx...yyy format.

host-bin/mount-chroot
```diff
@@ -216,32 +228,78 @@ $passphrase" | ecryptfs-wrap-passphrase "$wrappedfnek" -
...
+
+            if [ "$PAD1" = '................' ] && [ "$PAD2" = '................']; then
```
dnschneid: -a is fine here

host-bin/mount-chroot
```diff
@@ -185,29 +199,27 @@ for NAME in "$@"; do
...
-            echo -n "$fnek
+                echo -n "$fnek
 $passphrase" | ecryptfs-wrap-passphrase "$wrappedfnek" -
-            unset key fnek
-            echo | cat - "$wrappedkey" "$wrappedfnek" > "$KEYFILE"
+                unset key fnek
+                echo | cat - "$wrappedkey" "$wrappedfnek" > "$KEYFILE"
+            else
+                echo -n "
+................$key................$fnek" > "$KEYFILE"
```
dnschneid: Can you put the 16 periods into a variable at the top?  Counting them all each time is error-prone.

host-bin/mount-chroot
```diff
@@ -216,32 +228,78 @@ $passphrase" | ecryptfs-wrap-passphrase "$wrappedfnek" -
...
+                    unset key fnek
+                elif [ -z "$NOENC" -a -n "$NOPASS" ]; then
+                    key="`echo -n "$oldpassphrase" \
+                        | ecryptfs-unwrap-passphrase "$wrappedkey" - 2>/dev/null`"
+                    fnek="`echo -n "$oldpassphrase" \
+                        | ecryptfs-unwrap-passphrase "$wrappedfnek" - 2>/dev/null`"
+                    echo -n "
+................$key................$fnek" > "$KEYFILE"
+                elif [ -n "$NOENC" -a -n "$NOPASS" ]; then
+                    echo "Changing from nothing to nothing? You must be high-security..."
```
dnschneid: I guess the point I was making with the alternate wording was that I'd rather not direct snark at users (each one of them is a special flower and whatever).

dnschneid:
Is hexdump still giving you a bad byte count, even with the `32/1` update?  It seemed fixed to me.

host-bin/mount-chroot
Timvrakas: Sorry, that was gedit trying to use tabs instead of spaces.
Now that I mention it, should I be using something more advanced than gedit to do this type of work?

host-bin/mount-chroot
Timvrakas: Oh okay, ROT-13 it is :)

Timvrakas:
The issue with the hexdump is that old keys (made before the fix) are still too short. I will have to do line-by-line separation.

host-bin/mount-chroot
Timvrakas: Hey in my defense I caught this before you pointed it out :)

dnschneid:
Good point.  I don't think you need to do line-by-line though.  Something like:

If first 16 characters of key is ....., it's unencrypted (don't check fnek to confirm).  Then you can easily extract the keys (which are just hex):

```
contents="`tail -n+1 keyfile`"
key="${keyfile#${DOTS}}"
key="${key%%.*}"
fnek="${keyfile##*.}"
```

host-bin/mount-chroot
dnschneid: Serves me right for checking things as you're pushing new commits :)

host-bin/mount-chroot
dnschneid: Not really.  I use vim, but you'd do just as well configuring gedit to not insert tabs.

Timvrakas:
OK I finished testing and worked out a few bugs. It all works for the most parts.
TODO:
1. Deal with the malformed (short) keys (as mentioned above)
2. The -ee change password should check for request password.
3. Link empty-password to -k flag (as mentioned above)

Timvrakas:
Well this is going to sound lame, but I just lost the completed above TODO. Somebody pressed the spacebar..... grr. I will have it done again in a few days.

dnschneid:
Whoops!  No rush :)

Timvrakas:
Haha, well it looks like I got side-tracked for a long time....
Anyway now I have to deal with the new echo_tty functions

host-bin/mount-chroot
```diff
@@ -60,27 +64,46 @@ addtrap "stty echo 2>/dev/null"
...
+        echo_tty -n "Choose an encryption passphrase for $NAME: "
         read -r passphrase
-        if [ -z "$passphrase" ]; then
-            echo_tty ''
-            echo_tty -n 'You must specify a passphrase: '
-            continue
-        fi
         echo_tty ''
-        echo_tty -n 'Please confirm your passphrase: '
+        echo -n 'Please confirm your passphrase: ' 1>&2
```
dnschneid: This should be echo_tty

host-bin/mount-chroot
```diff
@@ -60,27 +64,46 @@ addtrap "stty echo 2>/dev/null"
...
+                if [ "${response#[Yy]}" = "$response" ]; then
+                    continue
+                else
+                NOPASS='y'
+                echo_tty 'No password set, keys will not be encrypted'
+                fi
+            else
+                echo_tty \
+'You tried to specify no passphrase, but to do that,
+you must store you keyfile in a custom location with -k.'
```
dnschneid:
```
Passphrases cannot be empty unless the keyfile
is stored separately from the chroot using -k.
```

host-bin/mount-chroot
```diff
@@ -60,27 +64,46 @@ addtrap "stty echo 2>/dev/null"
 
 # Function to prompt the user for a passphrase. Sets $passphrase.
 promptNewPassphrase() {
-    echo_tty -n "Choose an encryption passphrase for $NAME: "
-    [ -t 0 ] && stty -echo
-    while [ -z "$passphrase" ]; do
+    NOPASS=''
+    if [ ! "$CHROOTSRC/.ecryptfs" = "$KEYFILE" ]; then
+        ENABLENOPASS='y'
```
dnschneid: Does ENABLENOPASS need to be a global variable if you redetermine it each time?

host-bin/mount-chroot
```diff
@@ -60,27 +64,46 @@ addtrap "stty echo 2>/dev/null"
 
 # Function to prompt the user for a passphrase. Sets $passphrase.
 promptNewPassphrase() {
-    echo_tty -n "Choose an encryption passphrase for $NAME: "
-    [ -t 0 ] && stty -echo
-    while [ -z "$passphrase" ]; do
+    NOPASS=''
+    if [ ! "$CHROOTSRC/.ecryptfs" = "$KEYFILE" ]; then
```
dnschneid: !=

host-bin/mount-chroot
```diff
@@ -60,27 +64,46 @@ addtrap "stty echo 2>/dev/null"
...
+            if [ -n "$ENABLENOPASS" ]; then
+                echo_tty -n \
+'You did not specify a passphrase.  Your keys will not be encrypted.
+Only do this if you are confident that your keys are stored in a secure location.
+Do you want to continue? [y/N]: '
+                read -r response
+                if [ "${response#[Yy]}" = "$response" ]; then
+                    continue
+                else
+                NOPASS='y'
```
dnschneid: indentation

host-bin/mount-chroot
```diff
@@ -186,63 +209,112 @@ for NAME in "$@"; do
...
+                elif [ -n "$NOENC" -a -z "$NOPASS" ]; then
+                    # Wrap previously unwrapped pasphrase
+                    echo -n "$key
+$passphrase" | ecryptfs-wrap-passphrase "$wrappedkey" -
+                    echo -n "$fnek
+$passphrase" | ecryptfs-wrap-passphrase "$wrappedfnek" -
+                    echo | cat - "$wrappedkey" "$wrappedfnek" > "$KEYFILE"
+                    unset key fnek
+                elif [ -z "$NOENC" -a -n "$NOPASS" ]; then
+                    #Unwrap previously wrapped passphrase
```
dnschneid: space after # (sorry for nits :) )

dnschneid:
This is shaping up very well!

dnschneid:
The last thing I'd like to see is additions to the 17-encryption tests to handle no passwords (both with -k and without -k specified).

host-bin/mount-chroot
```diff
@@ -60,27 +63,46 @@ addtrap "stty echo 2>/dev/null"
 
 # Function to prompt the user for a passphrase. Sets $passphrase.
 promptNewPassphrase() {
-    echo_tty -n "Choose an encryption passphrase for $NAME: "
-    [ -t 0 ] && stty -echo
-    while [ -z "$passphrase" ]; do
+    NOPASS=''
```
dnschneid: now it's a local variable, so add `local`

dnschneid:
Considering how critical this is to users' data, I can't merge this unless there are tests in place to prove it works.  Could you add tests to [17-encryption](https://github.com/Timvrakas/crouton/blob/encryption/test/tests/17-encryption) to stress the new code paths?

Timvrakas:
OK, I'm still figuring out the testing architecture, But I will figure it out eventually.

dnschneid:
@Timvrakas shouldn't be too complex (you can run the encryption tests on your device by dropping the repo outside your chroot and running `./test/run.sh 17`).  If you need help, swing by \#crouton-devel for advice.  It'd be awesome to get this feature merged, but we have to be careful.

gbrld:
Much simpler to just store some long random password in the home folder, then modify mount-chroot to read it out.  Additionally I recommend moving all the scripts from /usr/local/bin to within the eCryptfs home folder (i.e. /home/chronos/user/crouton/) and making a symlink to the chroot storage (i.e. /home/chronos/user/chroot) 

```
mount-chroot:
...
PWFILE="/home/chronos/user/crouton/pwfile"
...
            if [ -e "$PWFILE" ]; then
                passphrase="`cat $PWFILE`"
            else
                echo_tty -n "Enter encryption passphrase for $NAME: "
                [ -t 0 ] && stty -echo
                if [ -z "$passphrase" ]; then
                    read -r passphrase
                fi
                [ -t 0 ] && stty echo
                echo_tty ''
            fi
...
```

dnschneid:
> Much simpler to just store some long random password in the home folder, then modify mount-chroot to read it out.

The password would have to have as many bits of entropy as the key itself, assuming the wrapping function actually can use that much data.  Seems like one way or another this would weaken the chroot protection somewhat, compared to storing the key directly in encrypted storage.  Of course, it's still better (and more convenient) than the current approach.

> Additionally I recommend moving all the scripts from /usr/local/bin to within the eCryptfs home folder (i.e. /home/chronos/user/crouton/) and making a symlink to the chroot storage (i.e. /home/chronos/user/chroot)

This is also very prudent.  To be clear, you have to run the scripts directly from the encrypted storage.

PR #1596, opened 2015-04-09
host-bin/startxiwi

DennisLfromGA:
A script that wraps enter-chroot to launch a xiwi chroot_app in a new windowed or tabbed session.

### DIFF
--- a/host-bin/startxiwi
+++ b/host-bin/startxiwi
@@ -8,52 +8,42 @@ set -e
 APPLICATION="${0##*/}"
 OPTS_ENTER=''
 OPTS_XIWI=''
-USAGE="$APPLICATION [options] CHROOT_APP [options]
 
-Wraps enter-chroot to launch a xiwi-app in a new windowed or tabbed session.
-You can launch any individual chroot graphical application in crouton windows.
+USAGE="$APPLICATION [options] chroot_app [parameters]
 
-By default, it will log into the primary user on the first chroot found.
+Wraps enter-chroot to launch a window or tab in Chromium OS for any graphical application.
+Applications launched in this way show in independent windows or tabs.
 
-Enter '-h' for 'enter-chroot' options and enter '-H' for 'xiwi-app' options.
+By default, it will use the primary user on the first xiwi-enabled chroot found.
+
+Enter '-h' for help with 'enter-chroot' and 'xiwi' options.
 "
-USAGE_ENTER="enter-chroot [options]
 
-  -b          Fork and run the specified command silently in the background.
-  -c CHROOTS  Directory the chroots are in. Default: /mnt/stateful_partition/crouton/chroots
-  -l          Make the command part of a login. Parameters are passed directly
-              to the chroot command, and a call to su is appended.
-  -k KEYFILE  Override the auto-detected encryption key location.
-  -n NAME     Name of the chroot to enter. Default: first one found in CHROOTS
-  -t TARGET   Only enter the chroot if it contains the specified TARGET.
-  -u USERNAME Username (or UID) to log into. Default: 1000 (the primary user)
-  -X XMETHOD  (Override the auto-detected XMETHOD for this session.) - deprecated, 'xiwi' is enforced.
-  -x          Does not log in, but directly executes the command instead.
-              Note that the environment will be empty (sans TERM).
-              Specify -x a second time to run the /prepare.sh script.
-"
-USAGE_XIWI="xiwi [options] CHROOT_APP [options]
+USAGE_HELP="Usage: enter-chroot [options]
 
-  -f          Prevent xiwi from quitting automatically.[1]
-  -F          Will launch the app full-screen.[2]
-  -T          Will launch the app in a tab.
+`enter-chroot -h 2>&1 | grep -e ' -[bckntu]'`
 
-By default, the CHROOT_APP is launched in a window.
+    Options: -l, -x, and -X are deprecated for the xiwi application.
 
-  [1] xiwi will normally close when the application returns. Some gui applications
-      fork before or during normal operation, which can confuse xiwi and cause it to
-      quit prematurely. If your application does not have a parameter that prevents
-      it from forking, and crouton is unable to automatically detect the fork, you can
-      use -f to prevent xiwi from quitting automatically.
-      xiwi will quit if you close the Chromium OS window when nothing is displayed.
+
+Usage: xiwi [options] chroot_app [parameters]
+
+`enter-chroot -t xiwi xiwi -h 2>&1 | grep -e ' -[FT]' | sed 's/^/    /g'`
+
+    xiwi will normally close when the application returns. Some gui applications
+    fork before or during normal operation, which can confuse xiwi and cause it to
+    quit prematurely. If your application does not have a parameter that prevents
+    it from forking, and crouton is unable to automatically detect the fork, you can
+    use -f to prevent xiwi from quitting automatically.
+    xiwi will quit if you close the Chromium OS window when nothing is displayed.
   
-  [2] A default window manager will full-screen all windows, unless CHROOT_APP begins
-      with 'start' or is 'xinit'. You can use -F to force the CHROOT_APP full-screen.
+    A default window manager will full-screen all windows, unless chroot_app begins
+    with 'start' or is 'xinit'. You can use -F to force the chroot_app full-screen.
 
-  You can cycle through multiple windows inside the
-  application via Ctrl-Alt-Tab/Ctrl-Alt-Shift-Tab, or close them via
-  Ctrl-Alt-Shift-Escape.  If CHROOT_APP begins with 'start' but you still want to
-  use the default window manager, specify the full path of the application.
+    You can cycle through multiple windows inside the
+    application via Ctrl-Alt-Tab/Ctrl-Alt-Shift-Tab, or close them via
+    Ctrl-Alt-Shift-Escape.  If chroot_app begins with 'start' but you still want to
+    use the default window manager, specify the full path of the application.
 "
 
 # Exits the script with exit code $1, spitting out message $@ to stderr
@@ -64,30 +54,27 @@ error() {
     exit "$ecode"
 }
 
-options=':bc:hlk:n:t:u:xfFHT'
+options='bc:hk:n:t:u:fFT'
 while getopts $options OPT; do
   case ${OPT} in
-    b) OPTS_ENTER="$OPTS_ENTER -b";;
-    c) OPTS_ENTER="$OPTS_ENTER -c ${OPTARG}";;
-    h) error 0 "$USAGE_ENTER";;
-    l) OPTS_ENTER="$OPTS_ENTER -l";;
-    k) OPTS_ENTER="$OPTS_ENTER -k ${OPTARG}";;
-    n) OPTS_ENTER="$OPTS_ENTER -n ${OPTARG}";;
-    t) OPTS_ENTER="$OPTS_ENTER -t ${OPTARG}";;
-    u) OPTS_ENTER="$OPTS_ENTER -u ${OPTARG}";;
-    x) OPTS_ENTER="$OPTS_ENTER -x";;
-    f) OPTS_XIWI="$OPTS_XIWI -f";;
-    F) OPTS_XIWI="$OPTS_XIWI -F";;
-    H) error 0 "$USAGE_XIWI";;
-    T) OPTS_XIWI="$OPTS_XIWI -T";;
-   \?) echo "Unknown option: -$OPTARG" 1>&2
-       error 2 "$USAGE";;
-    :) echo "Missing option argument for -$OPTARG"
-       error 2 "$USAGE";;
-    *) echo "Unimplemented option: -$OPTARG" 1>&2
-       error 2 "$USAGE";;
+    b) OPTS_ENTER="$OPTS_ENTER -${OPT}";;
+    c|k|n|t|u) OPTS_ENTER="$OPTS_ENTER -${OPT} ${OPTARG}";;
+    h) error 2 "$USAGE_HELP";;
+    f|F|T) OPTS_XIWI="$OPTS_XIWI -${OPT}";;
+   \?) error 2 "$USAGE";;
   esac
 done
 shift "$((OPTIND-1))"
 
-sh -e "`dirname "\`readlink -f "$0"\`"`/enter-chroot" $OPTS_ENTER -t xiwi xiwi $OPTS_XIWI "$@"
+for param in "$OPTS_ENTER"; do
+  opts_enter="$opts_enter'`echo -n "$param" | sed "s/'/'\\\\\\''/g"`'"
+  opts_enter="`echo -n "$opts_enter" | sed -e "s/^ //" -e "s/'//g"`"
+done
+for param in "$OPTS_XIWI"; do
+  opts_xiwi="$opts_xiwi'`echo -n "$param" | sed "s/'/'\\\\\\''/g"`'"
+  opts_xiwi="`echo -n "$opts_xiwi" | sed -e "s/^ //" -e "s/'//g"`"
+done
+
+eval exec sh -e "`dirname "\`readlink -f "$0"\`"`/enter-chroot" -t xiwi $opts_enter exec xiwi $opts_xiwi "$@" \
+     || ret=$?
+exit $ret
--- a/host-bin/startxiwi
+++ b/host-bin/startxiwi
@@ -27,16 +27,16 @@ USAGE_ENTER="enter-chroot [options]
   -n NAME     Name of the chroot to enter. Default: first one found in CHROOTS
   -t TARGET   Only enter the chroot if it contains the specified TARGET.
   -u USERNAME Username (or UID) to log into. Default: 1000 (the primary user)
-  -X XMETHOD  (Override the auto-detected XMETHOD for this session.) - deprecated, 'xiwi' is enforced.
+  -X XMETHOD  Override the auto-detected XMETHOD for this session.
   -x          Does not log in, but directly executes the command instead.
               Note that the environment will be empty (sans TERM).
               Specify -x a second time to run the /prepare.sh script.
 "
 USAGE_XIWI="xiwi [options] CHROOT_APP [options]
 
-  -f          prevent xiwi from quitting automatically.[1]
-  -F          will launch the app full-screen.[2]
-  -T          will launch the app in a tab.
+  -f          Prevent xiwi from quitting automatically.[1]
+  -F          Will launch the app full-screen.[2]
+  -T          Will launch the app in a tab.
 
 By default, the CHROOT_APP is launched in a window.
 
@@ -50,29 +50,31 @@ By default, the CHROOT_APP is launched in a window.
   [2] A default window manager will full-screen all windows, unless CHROOT_APP begins
       with 'start' or is 'xinit'. You can use -F to force the CHROOT_APP full-screen.
 
-      You can cycle through multiple windows inside the
-      application via Ctrl-Alt-Tab/Ctrl-Alt-Shift-Tab, or close them via
-      Ctrl-Alt-Shift-Escape.  If CHROOT_APP begins with 'start' but you still want to
-      use the default window manager, specify the full path of the application.
+  You can cycle through multiple windows inside the
+  application via Ctrl-Alt-Tab/Ctrl-Alt-Shift-Tab, or close them via
+  Ctrl-Alt-Shift-Escape.  If CHROOT_APP begins with 'start' but you still want to
+  use the default window manager, specify the full path of the application.
 "
+
+if [ "$#" = 0 ]; then echo "$USAGE"; exit 2; fi
 
 while getopts bc:hlk:n:t:u:xfFHT OPT; do
   case ${OPT} in
-    [b]) OPTS_ENTER="$OPTS_ENTER -b";;
-    [c]) OPTS_ENTER="$OPTS_ENTER -c ${OPTARG}";;
-    [h]) echo "$USAGE_ENTER"; exit 0;;
-    [l]) OPTS_ENTER="$OPTS_ENTER -l";;
-    [k]) OPTS_ENTER="$OPTS_ENTER -k ${OPTARG}";;
-    [n]) OPTS_ENTER="$OPTS_ENTER -n ${OPTARG}";;
-    [t]) OPTS_ENTER="$OPTS_ENTER -t ${OPTARG}";;
-    [u]) OPTS_ENTER="$OPTS_ENTER -u ${OPTARG}";;
-    [x]) OPTS_ENTER="$OPTS_ENTER -x";;
-    [X]) echo "'-X' option ignored, XMETHOD is always 'xiwi'";;
-    [f]) OPTS_XIWI="$OPTS_XIWI -f";;
-    [F]) OPTS_XIWI="$OPTS_XIWI -F";;
-    [H]) echo "$USAGE_XIWI"; exit 0;;
-    [T]) OPTS_XIWI="$OPTS_XIWI -T";;
-    \?)  echo "$USAGE"; exit 2;;
+    b) OPTS_ENTER="$OPTS_ENTER -b";;
+    c) OPTS_ENTER="$OPTS_ENTER -c ${OPTARG}";;
+    h) echo "$USAGE_ENTER"; exit 0;;
+    l) OPTS_ENTER="$OPTS_ENTER -l";;
+    k) OPTS_ENTER="$OPTS_ENTER -k ${OPTARG}";;
+    n) OPTS_ENTER="$OPTS_ENTER -n ${OPTARG}";;
+    t) OPTS_ENTER="$OPTS_ENTER -t ${OPTARG}";;
+    u) OPTS_ENTER="$OPTS_ENTER -u ${OPTARG}";;
+    x) OPTS_ENTER="$OPTS_ENTER -x";;
+    X) echo "'-X' option ignored, XMETHOD is always 'xiwi'";;
+    f) OPTS_XIWI="$OPTS_XIWI -f";;
+    F) OPTS_XIWI="$OPTS_XIWI -F";;
+    H) echo "$USAGE_XIWI"; exit 0;;
+    T) OPTS_XIWI="$OPTS_XIWI -T";;
+    \?) echo "$USAGE"; exit 2;;
   esac
 done
 shift "$((OPTIND-1))"
--- a/host-bin/startxiwi
+++ b/host-bin/startxiwi
@@ -14,35 +14,30 @@ USAGE="$APPLICATION [options] chroot_app [parameters]
 Wraps enter-chroot to launch a window or tab in Chromium OS for any graphical application.
 Applications launched in this way show in independent windows or tabs.
 
-By default, it will use the primary user on the first xiwi-enabled chroot found.
+By default, it will use the primary user on the first xiwi-enabled chroot found and launch
+the chroot_app in a window.
 
-Enter '-h' for help with 'enter-chroot' and 'xiwi' options.
-"
+    -b          Fork and run the specified command silently in the background.
+    -c CHROOTS  Directory the chroots are in. Default: /usr/local/chroots
+    -f          Prevent xiwi from quitting automatically.[1]
+    -h          Help with $APPLICATION options, with brief explanations.
+    -k KEYFILE  Override the auto-detected encryption key location.
+    -n NAME     Name of the chroot to enter. Default: first one found in CHROOTS
+    -t TARGET   Only enter the chroot if it contains the specified TARGET.
+    -u USERNAME Username (or UID) to log into. Default: 1000 (the primary user)
+    -F          Launch the chroot_app full-screen.
+    -T          Launch the chroot_app in a tab.
 
-USAGE_HELP="Usage: enter-chroot [options]
-
-`enter-chroot -h 2>&1 | grep -e ' -[bckntu]'`
-
-    Options: -l, -x, and -X are deprecated for the xiwi application.
-
-
-Usage: xiwi [options] chroot_app [parameters]
-
-`enter-chroot -t xiwi xiwi -h 2>&1 | grep -e ' -[FT]' | sed 's/^/    /g'`
-
-    xiwi will normally close when the application returns. Some gui applications
-    fork before or during normal operation, which can confuse xiwi and cause it to
-    quit prematurely. If your application does not have a parameter that prevents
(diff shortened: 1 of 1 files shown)


host-bin/startxiwi
```diff
@@ -0,0 +1,93 @@
...
+# Copyright (c) 2015 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+set -e
+
+APPLICATION="${0##*/}"
+OPTS_ENTER=''
+OPTS_XIWI=''
+USAGE="$APPLICATION [options] CHROOT_APP [options]
```
dnschneid: `$APPLICATION [options] chroot_app [parameters]`

host-bin/startxiwi
```diff
@@ -0,0 +1,93 @@
...
+
+Wraps enter-chroot to launch a xiwi-app in a new windowed or tabbed session.
+You can launch any individual chroot graphical application in crouton windows.
```
dnschneid: "xiwi-app" is jargon.  How about mirroring (and updating) the blurb in `xiwi`?

```
Wraps enter-chroot to launch a window or tab in Chromium OS for any graphical application.
Applications launched in this way show in independent windows or tabs.
```

host-bin/startxiwi
```diff
@@ -0,0 +1,93 @@
...
+
+By default, it will log into the primary user on the first chroot found.
```
dnschneid: `By default, it will use the primary user on the first xiwi-enabled chroot found.`

host-bin/startxiwi
```diff
@@ -0,0 +1,93 @@
...
+
+Enter '-h' for 'enter-chroot' options and enter '-H' for 'xiwi-app' options.
```
dnschneid: I think -h should just have two sections: enter-chroot options and xiwi options.
`enter-chroot` options should be outputted by actually running `enter-chroot -h` and selecting the relevant options.  Perhaps:

```
Chroot options:
`enter-chroot -h 2>&1 | grep -e ' -[bckntu]'`
```

host-bin/startxiwi
```diff
@@ -0,0 +1,93 @@
...
+  -k KEYFILE  Override the auto-detected encryption key location.
+  -n NAME     Name of the chroot to enter. Default: first one found in CHROOTS
+  -t TARGET   Only enter the chroot if it contains the specified TARGET.
+  -u USERNAME Username (or UID) to log into. Default: 1000 (the primary user)
+  -X XMETHOD  (Override the auto-detected XMETHOD for this session.) - deprecated, 'xiwi' is enforced.
+  -x          Does not log in, but directly executes the command instead.
+              Note that the environment will be empty (sans TERM).
+              Specify -x a second time to run the /prepare.sh script.
+"
+USAGE_XIWI="xiwi [options] CHROOT_APP [options]
```
dnschneid: I'm not sure of a good way to deduplicate the help text, but most (if not all) of the text in `xiwi`'s help text probably should be here.

host-bin/startxiwi
```diff
@@ -0,0 +1,93 @@
...
+By default, the CHROOT_APP is launched in a window.
+
+  [1] xiwi will normally close when the application returns. Some gui applications
+      fork before or during normal operation, which can confuse xiwi and cause it to
+      quit prematurely. If your application does not have a parameter that prevents
+      it from forking, and crouton is unable to automatically detect the fork, you can
+      use -f to prevent xiwi from quitting automatically.
+      xiwi will quit if you close the Chromium OS window when nothing is displayed.
+  
+  [2] A default window manager will full-screen all windows, unless CHROOT_APP begins
```
dnschneid: -F actually has nothing to do with the default window manager.  The default window manager (i3) manages windows inside of the chroot in all modes (tab, window, full-screen).

host-bin/startxiwi
```diff
@@ -0,0 +1,93 @@
...
+
+# Exits the script with exit code $1, spitting out message $@ to stderr
+error() {
+    local ecode="$1"
+    shift
+    echo "$*" 1>&2
+    exit "$ecode"
+}
+
+options=':bc:hlk:n:t:u:xfFHT'
```
dnschneid: -l, -x, and now -H probably shouldn't be recognized.

Also, it seems enter-chroot doesn't actually recognize -h.  That should be fixed...

host-bin/startxiwi
```diff
@@ -0,0 +1,93 @@
...
+while getopts $options OPT; do
+  case ${OPT} in
+    b) OPTS_ENTER="$OPTS_ENTER -b";;
+    c) OPTS_ENTER="$OPTS_ENTER -c ${OPTARG}";;
```
dnschneid: c/k/n/t/u can be combined into one case statement

dnschneid: Also, OPTARG/OPTS_ENTER is very tricky here, as you need to be able to handle paths with spaces and whatnot.  One trick is to do what enter-chroot does, and prepare a string with each parameter '-escaped.  That means surrounding ${OPTARG} with single quotes, and passing the contents of OPTARG through sed to replace instances of single-quote with the line `'\''` (see enter-chroot).  Finally, when you go to actually call enter-chroot, you need to actually wrap it in an eval, so that the resulting OPTS_ENTER gets expanded and split.  HOWEVER, you need to make sure nothing else gets doubly-expanded (otherwise you'd have to manually escape that as well), so you need to escape out any other variable references in the eval line.

Basically, it's messy, and another reason why you want to combine all the OPTS_ENTER/OPTARG cases so you only have to do messy stuff once here, and once down below.

host-bin/startxiwi
```diff
@@ -0,0 +1,93 @@
...
+    h) error 0 "$USAGE_ENTER";;
+    l) OPTS_ENTER="$OPTS_ENTER -l";;
+    k) OPTS_ENTER="$OPTS_ENTER -k ${OPTARG}";;
+    n) OPTS_ENTER="$OPTS_ENTER -n ${OPTARG}";;
+    t) OPTS_ENTER="$OPTS_ENTER -t ${OPTARG}";;
+    u) OPTS_ENTER="$OPTS_ENTER -u ${OPTARG}";;
+    x) OPTS_ENTER="$OPTS_ENTER -x";;
```
dnschneid: These too can be combined into one case statemen.

host-bin/startxiwi
```diff
@@ -0,0 +1,93 @@
...
+    echo "$*" 1>&2
+    exit "$ecode"
+}
+
+options=':bc:hlk:n:t:u:xfFHT'
+while getopts $options OPT; do
+  case ${OPT} in
+    b) OPTS_ENTER="$OPTS_ENTER -b";;
+    c) OPTS_ENTER="$OPTS_ENTER -c ${OPTARG}";;
+    h) error 0 "$USAGE_ENTER";;
```
dnschneid: Should be error 2 (I think we're consistent about this, but maybe not)

host-bin/startxiwi
```diff
@@ -0,0 +1,93 @@
...
+    k) OPTS_ENTER="$OPTS_ENTER -k ${OPTARG}";;
+    n) OPTS_ENTER="$OPTS_ENTER -n ${OPTARG}";;
+    t) OPTS_ENTER="$OPTS_ENTER -t ${OPTARG}";;
+    u) OPTS_ENTER="$OPTS_ENTER -u ${OPTARG}";;
+    x) OPTS_ENTER="$OPTS_ENTER -x";;
+    f) OPTS_XIWI="$OPTS_XIWI -f";;
+    F) OPTS_XIWI="$OPTS_XIWI -F";;
+    H) error 0 "$USAGE_XIWI";;
+    T) OPTS_XIWI="$OPTS_XIWI -T";;
+   \?) echo "Unknown option: -$OPTARG" 1>&2
```
dnschneid: Everywhere else we just catch ? (wildcard) and `error 2 "$USAGE"`.  getopts will print out an error itself.

host-bin/startxiwi
```diff
@@ -0,0 +1,93 @@
...
+       error 2 "$USAGE";;
+    :) echo "Missing option argument for -$OPTARG"
+       error 2 "$USAGE";;
+    *) echo "Unimplemented option: -$OPTARG" 1>&2
+       error 2 "$USAGE";;
+  esac
+done
+shift "$((OPTIND-1))"
+
+sh -e "`dirname "\`readlink -f "$0"\`"`/enter-chroot" $OPTS_ENTER -t xiwi xiwi $OPTS_XIWI "$@"
```
dnschneid: -t xiwi should probably go before $OPTS_ENTER.

dnschneid: Also, need more exec's.  The whole line should be an exec, and prior to "xiwi" there should be an exec.

dnschneid:
Overall I definitely like it though :)

DennisLfromGA:
Updated 'xiwi' to comply with above, I think...
I still need to update both `enter-chroot -h` option and `xiwi` USAGE statement.

tonyxue:
@DennisLfromGA Why not just call the command ```xiwi``` in the host-bin? So it is consistent with commands like ```startgnome```, which are the same name inside and outside the chroots.

btw, did you PM me? Sorry I didn't see it. My bouncer was online.

DennisLfromGA:
@tonyxue - You have a point. 
I think most users are familiar with the 'start*' naming convention though so I don't think '`startxiwi`' would seem foreign to them. Also, I would imagine most users don't know what the program name that executes inside the chroot is.
Ultimately, this decision is up to @dnschneid, I'll be glad to name it anything he prefers.

host-bin/startxiwi
```diff
@@ -0,0 +1,82 @@
...
+OPTS_XIWI=''
+
+USAGE="$APPLICATION [options] chroot_app [parameters]
+
+Wraps enter-chroot to launch a window or tab in Chromium OS for any graphical application.
+Applications launched in this way show in independent windows or tabs.
+
+By default, it will use the primary user on the first xiwi-enabled chroot found.
+
+Enter '-h' for help with 'enter-chroot' and 'xiwi' options.
```
dnschneid: I don't think you need to split USAGE and USAGE_HELP

host-bin/startxiwi
```diff
@@ -0,0 +1,82 @@
...
+"
+
+USAGE_HELP="Usage: enter-chroot [options]
+
+`enter-chroot -h 2>&1 | grep -e ' -[bckntu]'`
+
+    Options: -l, -x, and -X are deprecated for the xiwi application.
```
dnschneid: "deprecated" isn't really the right word.  I don't think you need to even mention enter-chroot (unnecessary implementation details).  Just list the parameters as if they were the script's.

host-bin/startxiwi
```diff
@@ -0,0 +1,82 @@
...
+
+
+Usage: xiwi [options] chroot_app [parameters]
+
+`enter-chroot -t xiwi xiwi -h 2>&1 | grep -e ' -[FT]' | sed 's/^/    /g'`
```
dnschneid: That, unfortunately, requires root, and may give different results depending on which chroot gets used in which instance.  I think for now, just copy-paste the text from xiwi, and when the rebase/parts stuff gets merged we can dedupe the text.

host-bin/startxiwi
```diff
@@ -0,0 +1,82 @@
...
+}
+
+options='bc:hk:n:t:u:fFT'
+while getopts $options OPT; do
+  case ${OPT} in
+    b) OPTS_ENTER="$OPTS_ENTER -${OPT}";;
+    c|k|n|t|u) OPTS_ENTER="$OPTS_ENTER -${OPT} ${OPTARG}";;
+    h) error 2 "$USAGE_HELP";;
+    f|F|T) OPTS_XIWI="$OPTS_XIWI -${OPT}";;
+   \?) error 2 "$USAGE";;
```
dnschneid: I think the other scripts literally do `?` (not `\?`), which also includes `:`.

host-bin/startxiwi
```diff
@@ -0,0 +1,82 @@
...
+    shift
+    echo "$*" 1>&2
+    exit "$ecode"
+}
+
+options='bc:hk:n:t:u:fFT'
+while getopts $options OPT; do
+  case ${OPT} in
+    b) OPTS_ENTER="$OPTS_ENTER -${OPT}";;
+    c|k|n|t|u) OPTS_ENTER="$OPTS_ENTER -${OPT} ${OPTARG}";;
```
dnschneid: The second you drop OPTARG into a combined variable, you lose any hope of recovering the division between parameters.  You need to do the escaping here, not down below.

host-bin/startxiwi
```diff
@@ -0,0 +1,82 @@
...
+   \?) error 2 "$USAGE";;
+  esac
+done
+shift "$((OPTIND-1))"
+
+for param in "$OPTS_ENTER"; do
+  ENTER_OPTS="$ENTER_OPTS'`echo -n "$param" | sed "s/'/'\\\\\\''/g"`'"
+  ENTER_OPTS="`echo -n "$ENTER_OPTS" | sed -e "s/^ //" -e "s/'//g"`"
+done
+for param in "$OPTS_XIWI"; do
```
dnschneid: OPTS_XIWI is safe, since there are no user parameters, so no need to worry about it here.

host-bin/startxiwi
```diff
@@ -0,0 +1,82 @@
...
+  XIWI_OPTS="$XIWI_OPTS'`echo -n "$param" | sed "s/'/'\\\\\\''/g"`'"
+  XIWI_OPTS="`echo -n "$XIWI_OPTS" | sed -e "s/^ //" -e "s/'//g"`"
+done
+
+eval exec sh -e "`dirname "\`readlink -f "$0"\`"`/enter-chroot" -t xiwi $ENTER_OPTS \
+     exec xiwi $XIWI_OPTS "$@" \
```
dnschneid: Not enough levels of indirection.  `eval` will expand the string once, then treat the result as if it were a line of shell.  So it will see your

```exec sh -e "`dirname "\`readlink -f "$0"\`"`/enter-chroot" -t xiwi $ENTER_OPTS exec xiwi $XIWI_OPTS "$@"```

convert it into

```exec sh -e path with spaces/to/enter-chroot -t xiwi -c path with spaces AND a random unclosed "quote exec xiwi -T these are actually five parameters all running together```

and then mayhem and/or invalid syntax errors ensue.  Instead:

```eval "exec sh -e '`dirname "\`readlink -f "$0"\`"`/enter-chroot' -t xiwi $ENTER_OPTS exec xiwi $XIWI_OPTS \"\$@\""```

might work, or if you're concerned about paths to enter-chroot including `'`'s, then:

```
enterchroot="`dirname "\`readlink -f "$0"\`"`/enter-chroot`"
eval "exec sh -e \"\$enterchroot\" -t xiwi $ENTER_OPTS exec xiwi $XIWI_OPTS \"\$@\""
```

The trick is that we're delaying expansion of any variable we don't want to be doubly-expanded.

...eval is scary

**Comment by dnschneid** [comment]
I'm also pretty torn as to whether it should be `startxiwi` or `xiwi` outside of the chroot.

`startxiwi`: goes with the idea that `start*` commands start the chroot, which people seem to believe (also necessitated the croutonnoroot script because everyone starts running start* with root from everywhere).  This, of course, means `enter-chroot` clearly doesn't start a chroot :)

`xiwi` goes with the idea that the command outside the chroot matches the command inside the chroot, which is what I was initially going for two years ago.  This may require xiwi inside the chroot to have a no-root check.

Option three: both, via symlink?

File: host-bin/startxiwi (hunk 13 above)
Comment (DennisLfromGA) [inline host-bin/startxiwi]: Combined.

File: host-bin/startxiwi (hunk 14 above)
Comment (DennisLfromGA) [inline host-bin/startxiwi]: Removed reference to unused options and enter-chroot.

File: host-bin/startxiwi (hunk 15 above)
Comment (DennisLfromGA) [inline host-bin/startxiwi]: Pulled in verbage from enter-chroot and xiwi.

File: host-bin/startxiwi (hunk 16 above)
Comment (DennisLfromGA) [inline host-bin/startxiwi]: I removed the backslash character but it looks like most others (enter-chroot, edit-chroot, etc.) still use it.

File: host-bin/startxiwi (hunk 17 above)
Comment (DennisLfromGA) [inline host-bin/startxiwi]: Relocated the OPTARG manipulation to the getopts routine.

File: host-bin/startxiwi (hunk 18 above)
Comment (DennisLfromGA) [inline host-bin/startxiwi]: Deleted the OPTS_XIWI manipulation loop.

File: host-bin/startxiwi (hunk 19 above)
Comment (DennisLfromGA) [inline host-bin/startxiwi]: Reworked the eval statement per your wise recommendations.

**Comment by DennisLfromGA** [comment]
> Option three: both, via symlink?

1 vote for option three.

File: host-bin/startxiwi [hunk 20]
```diff
@@ -0,0 +1,72 @@
...
+OPTS_XIWI=''
+
+USAGE="$APPLICATION [options] chroot_app [parameters]
+
+Wraps enter-chroot to launch a window or tab in Chromium OS for any graphical application.
+Applications launched in this way show in independent windows or tabs.
+
+By default, it will use the primary user on the first xiwi-enabled chroot found and launch
+the chroot_app in a window.
+
```
dnschneid: Options:

host-bin/startxiwi
```diff
@@ -0,0 +1,72 @@
...
+    -b          Fork and run the specified command silently in the background.
```
dnschneid: Sorry I wasn't clear.  The options derived from enter-chroot can be pulled in via a call to enter-chroot, since that doesn't require root.  The options derived from xiwi are the ones that need to be replicated here.  The result should be presented like it is here.

host-bin/startxiwi
```diff
@@ -0,0 +1,72 @@
...
+    -f          Prevent xiwi from quitting automatically.[1]
+    -h          Help with $APPLICATION options, with brief explanations.
+    -k KEYFILE  Override the auto-detected encryption key location.
+    -n NAME     Name of the chroot to enter. Default: first one found in CHROOTS
+    -t TARGET   Only enter the chroot if it contains the specified TARGET.
+    -u USERNAME Username (or UID) to log into. Default: 1000 (the primary user)
+    -F          Launch the chroot_app full-screen.
+    -T          Launch the chroot_app in a tab.
+
+    [1] xiwi will normally close when the application returns. Some gui applications
```
dnschneid: You can remove the level of indentation.

host-bin/startxiwi
```diff
@@ -0,0 +1,72 @@
...
+
+Wraps enter-chroot to launch a window or tab in Chromium OS for any graphical application.
+Applications launched in this way show in independent windows or tabs.
+
+By default, it will use the primary user on the first xiwi-enabled chroot found and launch
+the chroot_app in a window.
+
+    -b          Fork and run the specified command silently in the background.
+    -c CHROOTS  Directory the chroots are in. Default: /usr/local/chroots
+    -f          Prevent xiwi from quitting automatically.[1]
```
dnschneid: I suggest `(see notes below)` instead of the footnote notation.

host-bin/startxiwi
```diff
@@ -0,0 +1,72 @@
...
+    shift
+    echo "$*" 1>&2
+    exit "$ecode"
+}
+
+options='bc:fhk:n:t:u:FT'
+while getopts $options OPT; do
+    case ${OPT} in
+      b) OPTS_ENTER="$OPTS_ENTER -${OPT}";;
+      c|k|n|t|u) for param in "$OPTARG"; do
```
dnschneid: OPTARG is a single parameter, so you don't need to loop over it.

host-bin/startxiwi
```diff
@@ -0,0 +1,72 @@
...
+                     opts_enter="$opts_enter'$(echo -n "$param" | sed "s/'/'\\\\\\''/g")'"
+                     opts_enter="$(echo -n "$opts_enter" | sed -e "s/^ //" -e "s/'//g")"
+                 done
+                 OPTS_ENTER="$OPTS_ENTER -${OPT} ${opts_enter}";;
+      h) error 2 "$USAGE";;
+      f|F|T) OPTS_XIWI="$OPTS_XIWI -${OPT}";;
+      ?) error 2 "$USAGE";;
```
dnschneid: You're right, we use `\?` everywhere else.  Sorry for the misadvice.

host-bin/startxiwi
```diff
@@ -0,0 +1,72 @@
...
+options='bc:fhk:n:t:u:FT'
+while getopts $options OPT; do
+    case ${OPT} in
+      b) OPTS_ENTER="$OPTS_ENTER -${OPT}";;
+      c|k|n|t|u) for param in "$OPTARG"; do
+                     opts_enter="$opts_enter'$(echo -n "$param" | sed "s/'/'\\\\\\''/g")'"
+                     opts_enter="$(echo -n "$opts_enter" | sed -e "s/^ //" -e "s/'//g")"
+                 done
+                 OPTS_ENTER="$OPTS_ENTER -${OPT} ${opts_enter}";;
+      h) error 2 "$USAGE";;
```
dnschneid: Can be combined with the `\?` entry.

host-bin/startxiwi
```diff
@@ -0,0 +1,72 @@
...
+# Exits the script with exit code $1, spitting out message $@ to stderr
+error() {
+    local ecode="$1"
+    shift
+    echo "$*" 1>&2
+    exit "$ecode"
+}
+
+options='bc:fhk:n:t:u:FT'
+while getopts $options OPT; do
```
dnschneid: either quote `$options` or pull the literal in

host-bin/startxiwi
```diff
@@ -0,0 +1,72 @@
...
+    case ${OPT} in
```
dnschneid: quote, no {'s needed.

host-bin/startxiwi
```diff
@@ -0,0 +1,72 @@
...
+      b) OPTS_ENTER="$OPTS_ENTER -${OPT}";;
```
dnschneid: {} not necessary

host-bin/startxiwi
```diff
@@ -0,0 +1,72 @@
...
+        xiwi will quit if you close the Chromium OS window when nothing is displayed.
+  
+    You can cycle through multiple windows inside the
+    application via Ctrl-Alt-Tab/Ctrl-Alt-Shift-Tab, or close them via
+    Ctrl-Alt-Shift-Escape.  If the chroot_app begins with 'start' but you still want to
+    use the default window manager, specify the full path of the application.
+"
+
+# Exits the script with exit code $1, spitting out message $@ to stderr
+error() {
```
dnschneid: Since the two usage prints will be combined, I would ditch the reimplementation of `error` and just do

```
echo "$USAGE" 1>&2
exit 2
```

down below

host-bin/startxiwi
```diff
@@ -0,0 +1,72 @@
...
+                 done
+                 OPTS_ENTER="$OPTS_ENTER -${OPT} ${opts_enter}";;
+      h) error 2 "$USAGE";;
+      f|F|T) OPTS_XIWI="$OPTS_XIWI -${OPT}";;
+      ?) error 2 "$USAGE";;
+    esac
+done
+shift "$((OPTIND-1))"
+
+eval "exec sh -e '`dirname "\`readlink -f "$0"\`"`/enter-chroot' -t xiwi $OPTS_ENTER \
```
dnschneid: Thinking more about this, I'd err on the side of `'` paranoia (the second suggested implementation).  I should start labelling my external media "David's USB"

dnschneid: With exec, you don't need the `|| ret=$?` nor the `exit $ret`.  The command launched replaces the shell process entirely.

dnschneid:
One thing I was wondering about is if this script should be smart about files passed to it as arguments.  For example, if you cd into the Downloads directory and do `sudo startxiwi gimp image.png`, it would be nice if the script properly expanded the image.png path into a chroot-relative path.

A sneaky way of doing this that *should* behave correctly in almost all instances is to take the roundabout approach: instead of changing the arguments, change the working directory inside the chroot to match the equivalent directory outside the chroot.

1. Handle 100% of the cases by taking the current directory and processing it against the /etc/crouton/shares file.  Bleh.
2. Handle 99.99% of the cases by taking the current directory, and do (1) only if it's inside ~/Downloads
3. Handle 99.98% of the cases by taking the current directory, do (2), but assume downloads is mounted at ~/Downloads inside the chroot.

I suspect (3) is fine for now.
However, I think enter-chroot is missing a good way to set the initial working directory, so maybe this addition is too big for this PR.

host-bin/startxiwi
DennisLfromGA: Done.

host-bin/startxiwi
DennisLfromGA: Done. I just misunderstood, no problemmo.

host-bin/startxiwi
DennisLfromGA: I put in a 'NOTE: blurb' and left it indented, I can change this if I misunderstood...

host-bin/startxiwi
DennisLfromGA: Done.

host-bin/startxiwi
DennisLfromGA: Re-done. Hopefully, I did it right - this time.

host-bin/startxiwi
DennisLfromGA: Done.

host-bin/startxiwi
DennisLfromGA: Removed -h option.

host-bin/startxiwi
DennisLfromGA: Pulled in, like the other scripts.

host-bin/startxiwi
DennisLfromGA: Removed brackets everywhere'ish.

host-bin/startxiwi
DennisLfromGA: Removed same...

host-bin/startxiwi
DennisLfromGA: Removed error function and echo'ed the output.

DennisLfromGA:
> One thing I was wondering about is if this script should be smart about files passed to it as arguments.

This might be nice and convenient I think. Would this also benefit in 'tab completion' of file names, etc.?

> However, I think enter-chroot is missing a good way to set the initial working directory, so maybe this addition is too big for this PR.

I humbly agree ;)

host-bin/startxiwi
```diff
@@ -0,0 +1,54 @@
...
+USAGE="$APPLICATION [options] chroot_app [parameters]
+
+Wraps enter-chroot to launch a window or tab in Chromium OS for any graphical application.
+Applications launched in this way show in independent windows or tabs.
+
+By default, it will use the primary user on the first xiwi-enabled chroot found and launch
+the chroot_app in a window.
+
+Options:
+$(enter-chroot -h 2>&1 | grep -e ' -[bckntu]')
```
dnschneid: enter-chroot needs to be found, much like we do down below.

In fact, since you're already making an enterchroot variable, you can just make it sooner and reuse it!  Bonus!

host-bin/startxiwi
```diff
@@ -0,0 +1,54 @@
...
+    -F          Launch the chroot_app full-screen.
+    -T          Launch the chroot_app in a tab.
+    -f          Prevent xiwi from quitting automatically. (see NOTE below)
+
+NOTE: xiwi will normally close when the application returns. Some gui applications
+      fork before or during normal operation, which can confuse xiwi and cause it to
```
dnschneid: I don't think the indentation helps.

host-bin/startxiwi
```diff
@@ -0,0 +1,54 @@
...
+      use -f to prevent xiwi from quitting automatically.
+      xiwi will quit if you close the Chromium OS window when nothing is displayed.
+  
+You can cycle through multiple windows inside the application
+via Ctrl-Alt-Tab/Ctrl-Alt-Shift-Tab, or close them via Ctrl-Alt-Shift-Escape.
+If the chroot_app begins with 'start' but you still want to
+use the default window manager, specify the full path of the application.
+"
+
+while getopts bc:k:n:t:u:FTf OPT; do
```
dnschneid: Put a quote around the getopts string.  Not that it's really necessary, just that it's good practice.

dnschneid:
Getting close!

Yes, in theory the tab completion you already get in the Chromium OS shell would result in a filename that would work when passed to startxiwi.  Created #1604

DennisLfromGA:
> Quotes around OPTARG are kind of the point though. eval will see those and avoid expanding strange things.
> And I don't think that sed is the right place for cosmetic fixups.

Of course you're right again, I removed the 2nd & 3rd sed's and the eval cleaned it up. :)

DennisLfromGA:
Done'ish - again...
Thanx for being *so* patient with me.
I would've let someone else handle this if I knew I'd bork it so bad - definitely not a 'pie' for me. :(

dnschneid:
Last few nits!  This is a challenging bit of code and you've done quite well, so don't worry about it.
FYI I'm being extra nits-y so you know what to target in future PRs :)

DennisLfromGA:
> FYI I'm being extra nits-y so you know what to target in future PRs :)

Seems like you said that last time... :)

dnschneid:
And this PR is even harder than your edit-chroot one!  Congrats!

I think the last step is to add it to the `xiwi` target (whoops).

DennisLfromGA:
> I think the last step is to add it to the xiwi target (whoops).

Sorry, but to be clear, do you mean add it to the xiwi target TIPS example?

dnschneid:
Nope, to HOSTBIN, otherwise it won't get installed :)

dnschneid:
TIPS would be good too, though

DennisLfromGA:
Okay, gotcha.
Shall I tweak the USAGE in `host-bin/xiwi` to show the Options: like in `startxiwi` ?

dnschneid:
Sure!

DennisLfromGA:
Coming at cha.

DennisLfromGA:
@dnschneid - I'd like to close this PR and open a new one with both host-bin/startxiwi and targets/xiwi if that's alright with you??? I think it'd be much cleaner that way.

dnschneid:
Either way is fine.

DennisLfromGA:
Okay, I opened a new PR https://github.com/dnschneid/crouton/pull/1621

I'm sorry I couldn't figure out how to merge the existing mods, I had `xiwi` in a branch and `startxiwi` in my master - it was a very confusing setup. I just deleted my 'crouton' fork and started over, made a 'startxiwi' branch and then did the PR - much cleaner this way.

Bonus: Plus it hides all my embarrassing iterations and false starts from before. ;)

PR #1597, opened 2015-04-09
Add a badge to the extension with the number of windows

dnschneid:
Does anyone want this/definitely not want this?
I can also change the color if the matched green causes too much panic ("omg is crouton telling me I have two things I MUST DO?")

badge screenshot

### DIFF
--- a/host-ext/crouton/background.js
+++ b/host-ext/crouton/background.js
@@ -142,6 +142,11 @@ function refreshUI() {
     );
     chrome.browserAction.setTitle({title: 'crouton: ' + icon});
 
+    chrome.browserAction.setBadgeText(
+        {text: windows_.length > 1 ? '' + (windows_.length-1) : ''}
+    );
+    chrome.browserAction.setBadgeBackgroundColor({color: '#2E822B'});
+
     var views = chrome.extension.getViews({type: "popup"});
     for (var i = 0; i < views.length; views++) {
         var view = views[i];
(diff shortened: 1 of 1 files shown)


theblacklion:
Perhaps a grey badge would cause less "panic"?

Overall I honestly don't care having this because "my" extension is only being shown when the new-style menu is open... and the window manager does it's job quite fine..

tedm:
I would prefer it the same color it is now, without the badge. You could change the block to a more rounded C. Took me quite a while to figure out it was a C, looks Greek.

Watney:
Perhaps make the count display a user-selectable option :)

tonyxue:
I think someone used to iOS will be driven crazy by that. lol
maybe we should not make the number 'pop out of the icon' but 'blend into the icon' instead. That would make this group of people comfortable. :p

DennisLfromGA:
I may be in the minority but I like the proposal as is - square C, green
blocked number.

On Thu, Apr 9, 2015 at 7:08 AM, Xue, Zhaoxin (Tony) <
notifications@github.com> wrote:

> I think someone used to iOS will be driven crazy by that. lol
> maybe we should not make the number 'pop out of the icon' but 'blend into
> the icon' instead. That would make this group of people comfortable. :p
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dnschneid/crouton/pull/1597#issuecomment-91197204>.
>



-- 
*DennyL@GMail*

divx118:
+1  @DennisLfromGA  - square C, green blocked number.

dnschneid:
If you check out the commit, it's just using the standard badge rendering API.  Changing the color is really all I can do (simply).

Voice your opinion and I'll merge or not merge sometime next week :)

dnschneid:
Wow, when I asked for opinions, I didn't expect *this* many (zero).  Anyway, I'll merge as-is and see if anyone complains.

mark-hahn:
We all love it!

PR #1608, opened 2015-04-11
Tweaked usage message

DennisLfromGA:
Tweaked USAGE to be consistent with startxiwi.

### DIFF
--- a/chroot-bin/xiwi
+++ b/chroot-bin/xiwi
@@ -10,6 +10,8 @@ Launches a windowed session in Chromium OS for any graphical application.
 Applications launched in this way show in independent windows or tabs.
 All parameters are passed to the specified application.
 
+By default, the app is launched in a window.
+
 Options:
     -F          Launch the APPLICATION full-screen.
     -T          Launch the APPLICATION in a tab.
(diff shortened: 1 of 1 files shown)


chroot-bin/xiwi
```diff
@@ -7,17 +7,21 @@
 
 USAGE="Usage: ${0##*/} [-f] [-F|-T] APPLICATION [PARAMETERS ...]
 Launches a windowed session in Chromium OS for any graphical application.
+Applications launched in this way show in independent windows or tabs.
 All parameters are passed to the specified application.
 
-By default, the app is launched in a window. Passing -F will launch the app
```
dnschneid: I think you lost the "by default" part.

DennisLfromGA:
I lost it intentionally because I added the Options: section.
Maybe this file should be left alone?

dnschneid:
I don't see any mention of what the default behavior is, though.  Your startxiwi help text does describe it.

chroot-bin/xiwi
DennisLfromGA: Your right, sorry, I'll put it back in.
I guess I glossed over the `By default, the app is launched in a window.` part.

PR #1624, opened 2015-04-15
Error while "make all"

kanner:
I think there is an error in determining $(SRCTARGETS) in Makefile, trying to build "croutonfreon" (not "croutonfreon.so"). I got errors while "make all":

gcc -g -Wall -Werror -Os src/freon.c  -o croutonfreon
/usr/lib/gcc/x86_64-pc-linux-gnu/4.8.3/../../../../lib64/crt1.o: In function `_start':
(.text+0x20): undefined reference to `main'
/tmp/ccanIAv9.o: In function `preload_init':
/home/toor/projects/crouton/src/freon.c:43: undefined reference to `dlsym'
/home/toor/projects/crouton/src/freon.c:44: undefined reference to `dlsym'
/home/toor/projects/crouton/src/freon.c:45: undefined reference to `dlsym'

First error -- src/freon.c have no main() (thats because we`re trying to build it as binary, not library)
Other errors -- just missing "-ldl" (presents in "croutonfreon.so" target).

I think you should build libraries as separate target (as in my patch) or just move freon.c to lib/ folder or something like that.

src/xorg.c is missing anyway..

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ $(EXTTARGET): $(EXTSOURCES) Makefile
 $(EXTPEXE): $(EXTPEXESOURCES)
 	$(MAKE) -C host-ext/nacl_src
 
-$(SRCTARGETS): src/$(patsubst crouton%,src/%.c,$@) $($@_DEPS) Makefile
+$(SRCTARGETS): $(patsubst crouton%,src/%.c,$@) $($@_DEPS) Makefile
 	gcc $(CFLAGS) $(patsubst crouton%,src/%.c,$@) $($@_LIBS) -o $@
 
 $(LIBSTARGETS): $(patsubst crouton%.so,src/%.c,$@) $($@_DEPS) Makefile
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,7 @@ croutoncursor_LIBS = -lX11 -lXfixes -lXrender
 croutonfbserver_LIBS = -lX11 -lXdamage -lXext -lXfixes -lXtst
 croutonwmtools_LIBS = -lX11
 croutonxi2event_LIBS = -lX11 -lXi
+croutonfreon.so_LIBS = -ldl
 
 croutonwebsocket_DEPS = src/websocket.h
 croutonfbserver_DEPS = src/websocket.h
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -4,7 +4,8 @@
 
 TARGET = crouton
 EXTTARGET = crouton.zip
-SRCTARGETS = $(patsubst src/%.c,crouton%,$(wildcard src/*.c))
+SRCTARGETS = $(patsubst src/%.c,crouton%,$(shell find src/ -name *.c -and -not -name freon.c))
```
dnschneid: Defining LIBS first and using patsubst to provide a pattern to `$(filter-out pattern...,text)` would be way cleaner.

Makefile
```diff
@@ -4,7 +4,8 @@
...
+LIBS = croutonfreon.so #croutonxorg.so
```
dnschneid: Oops, xorg.c got removed a while ago, but I forgot to remove it from the Makefile.  You can eliminate the last trace of it.

dnschneid:
Thanks for the cleanup!

kanner:
all done

dnschneid:
Yeah, that's precisely it.  The last thing is that you could turn the croutonfreon.so makefile rule into a generic .so rule (like how the other one is), but that can either happen now or whenever we decide to add a second .so.  Your call.

kanner:
Lets do it now.
+ noticed wrong dependency path for SRCTARGETS (like src/src/*.c)

Makefile
```diff
@@ -31,6 +33,7 @@ croutoncursor_LIBS = -lX11 -lXfixes -lXrender
 croutonfbserver_LIBS = -lX11 -lXdamage -lXext -lXfixes -lXtst
 croutonwmtools_LIBS = -lX11
 croutonxi2event_LIBS = -lX11 -lXi
+croutonfreon.so_LIBS = -ldl
```
dnschneid: I think for LIBS you can assume -ldl is needed, instead of having to specify it for each .so

dnschneid: Well, I guess if we override something without pulling in the original target we wouldn't need it.  So nevermind on that front.

dnschneid:
Can you please [sign the CLA](https://cla.developers.google.com/clas/new?kind=KIND_INDIVIDUAL&domain=DOMAIN_GOOGLE)?

kanner:
Done, just forgot to do it. Thanks

dnschneid:
Thanks!

Issue #1045, opened 2014-09-15
Non-forking rc.local hangs crouton startup


mcvicar:
Had crouton up and running perfectly with precise for the past few weeks with a basic LAMP stack, plus Samba and open-ssh server.

I stupidly restarted precise (sudo shutdown -r now) by accident, which restarted the whole chromebox and now I can't get precise back up. If I attempt to start it all I get is 
```
sudo enter-chroot startxfce4
Entering /mnt/stateful_partition/crouton/chroots/precise...
```
Eventually I get bored of it sitting there doing nothing, so
```
^C
Unmounting /mnt/stateful_partition/crouton/chroots/precise..
```
I've tried a full power start of the chromebox, a fresh installer and `sudo sh ~/Downloads/crouton -n  precise -u` (which mounts, updates and unmounts fine). 

There doesn't appear to be an obvious way of verbosely outputting start-up, is there a way to do that?

PR #1650, opened 2015-04-21
Fix non-forking rc.local hangs crouton startup

tonyxue:
I was experiencing the issue mentioned in #1045 .

It's actually not a fix but just a kindly reminder for people, in case people are panic and don't know what's happening with the chroot.

### DIFF
--- a/host-bin/enter-chroot
+++ b/host-bin/enter-chroot
@@ -689,14 +689,13 @@ else
         ( 
             sleep 10
             echo "Still running chroot's rc.local..." 2>&1
-            while true 
-            do
-                    sleep 1
-            done
         ) & forkpid=$! 
         chrootcmd 'exec /etc/rc.local >/dev/null 2>/dev/null </dev/null' \
             || ret=$?
+        addtrap "kill '$forkpid' 2>/dev/null"
         kill "$forkpid"
+        wait "$forkpid"
+        undotrap
         if [ ! "$ret" = 0 ]; then
             echo "WARNING: /etc/rc.local failed with code $ret" 1>&2
             ret=0
--- a/host-bin/enter-chroot
+++ b/host-bin/enter-chroot
@@ -688,15 +688,15 @@ else
     if [ -n "$firstrun" -a -x "$CHROOT/etc/rc.local" ]; then
         ( 
             sleep 10
-            if [ ! $(ps -ef | grep -c /etc/rc.local) -eq 1 ]; then
-            echo "
-                        
-You rc.local is running for too long...
-Check the file if this is not expected!"
-            fi
-        ) &
+            echo "Still running chroot's rc.local..." 2>&1
+            while true 
+            do
+                    sleep 1
+            done
+        ) & forkpid=$! 
         chrootcmd 'exec /etc/rc.local >/dev/null 2>/dev/null </dev/null' \
             || ret=$?
+        kill "$forkpid"
         if [ ! "$ret" = 0 ]; then
             echo "WARNING: /etc/rc.local failed with code $ret" 1>&2
             ret=0
--- a/host-bin/enter-chroot
+++ b/host-bin/enter-chroot
@@ -686,16 +686,18 @@ if [ -n "$NOLOGIN" ]; then
 else
     # Check and run rc.local
     if [ -n "$firstrun" -a -x "$CHROOT/etc/rc.local" ]; then
+        
         ( 
             sleep 10
-            echo "Still running chroot's rc.local..." 2>&1
-        ) & forkpid=$! 
+            echo "Still running chroot's rc.local..." 2>&1 
+        ) & noticepid=$!
+        
+        addtrap "kill '$noticepid' 2>/dev/null"
         chrootcmd 'exec /etc/rc.local >/dev/null 2>/dev/null </dev/null' \
             || ret=$?
-        addtrap "kill '$forkpid' 2>/dev/null"
-        kill "$forkpid"
-        wait "$forkpid"
         undotrap
+        kill "$noticepid" 2>/dev/null
+        wait "$noticepid" || true
         if [ ! "$ret" = 0 ]; then
             echo "WARNING: /etc/rc.local failed with code $ret" 1>&2
             ret=0
--- a/host-bin/enter-chroot
+++ b/host-bin/enter-chroot
@@ -692,7 +692,7 @@ else
         chrootcmd 'exec /etc/rc.local >/dev/null 2>/dev/null </dev/null' \
             || ret=$?
         undotrap
-        kill "$noticepid" 2>/dev/null
+        kill "$noticepid" 2>/dev/null || true
         wait "$noticepid" || true
         if [ ! "$ret" = 0 ]; then
             echo "WARNING: /etc/rc.local failed with code $ret" 1>&2
--- a/host-bin/enter-chroot
+++ b/host-bin/enter-chroot
@@ -684,8 +684,19 @@ if [ -n "$NOLOGIN" ]; then
         fi
     fi
 else
-    # Run rc.local
+    # Check and run rc.local
     if [ -n "$firstrun" -a -x "$CHROOT/etc/rc.local" ]; then
+        if [ $(grep -cE '^[a-Z]|^\/' "$CHROOT/etc/rc.local") > 1 ]; then
+                ( 
+                  sleep 10
+                  if [ ! $(ps -ef | grep -c /etc/rc.local) -eq 1 ]; then
+                        echo "
+                        
+You rc.local is running for too long...
+Check the file if this is not expected!"
+                  fi
+                ) &
+        fi
         chrootcmd 'exec /etc/rc.local >/dev/null 2>/dev/null </dev/null' \
             || ret=$?
         if [ ! "$ret" = 0 ]; then
--- a/host-bin/enter-chroot
+++ b/host-bin/enter-chroot
@@ -686,17 +686,15 @@ if [ -n "$NOLOGIN" ]; then
 else
     # Check and run rc.local
     if [ -n "$firstrun" -a -x "$CHROOT/etc/rc.local" ]; then
-        if [ $(grep -cE '^[a-Z]|^\/' "$CHROOT/etc/rc.local") > 1 ]; then
-                ( 
-                  sleep 10
-                  if [ ! $(ps -ef | grep -c /etc/rc.local) -eq 1 ]; then
-                        echo "
+        ( 
+            sleep 10
+            if [ ! $(ps -ef | grep -c /etc/rc.local) -eq 1 ]; then
+            echo "
                         
 You rc.local is running for too long...
 Check the file if this is not expected!"
-                  fi
-                ) &
-        fi
+            fi
+        ) &
         chrootcmd 'exec /etc/rc.local >/dev/null 2>/dev/null </dev/null' \
             || ret=$?
         if [ ! "$ret" = 0 ]; then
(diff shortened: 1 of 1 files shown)


dnschneid:
You can actually do this more simply by storing the PID of that forked process (`$!`) and killing it after `chrootcmd` returns.  That way as long as the fork survives, you know that rc.local hasn't quit (so you don't have to grep for processes).

host-bin/enter-chroot
```diff
@@ -684,8 +684,17 @@ if [ -n "$NOLOGIN" ]; then
...
 else
-    # Run rc.local
+    # Check and run rc.local
     if [ -n "$firstrun" -a -x "$CHROOT/etc/rc.local" ]; then
+        ( 
+            sleep 10
+            if [ ! $(ps -ef | grep -c /etc/rc.local) -eq 1 ]; then
+            echo "
+                        
+You rc.local is running for too long...
```
dnschneid: A simple `echo "Still running the chroot's /etc/rc.local..." 1>&2` is probably sufficient.

tonyxue:
yea.. I thought about that and that's what you've suggested in #1045. But I think enter-chroot should by default prompt nothing, just like the start up of the system that rc.local is executed silently. And I'd prefer a clean command line.

But here people have to know if rc.local is running too long because it is run differently from it is in a normal system startup.

dnschneid:
Oh, I agree.  My suggestion doesn't change the behavior of your patch; it just simplifies it.

tonyxue:
Hah, I realized that I misunderstood your reply at the same time you replied me. :P 

I will change the method I use

tonyxue:
@dnschneid updated.

It will now continuously prompt to the user until finished.
So the user will not think that something after running rc.local goes wrong.

host-bin/enter-chroot
```diff
@@ -684,10 +684,19 @@ if [ -n "$NOLOGIN" ]; then
         fi
     fi
 else
-    # Run rc.local
+    # Check and run rc.local
     if [ -n "$firstrun" -a -x "$CHROOT/etc/rc.local" ]; then
+        ( 
+            sleep 10
+            echo "Still running chroot's rc.local..." 2>&1
+            while true
```
dnschneid: You can just let it exit.  The PID will hang around until you use "wait" below.

host-bin/enter-chroot
```diff
@@ -684,10 +684,19 @@ if [ -n "$NOLOGIN" ]; then
...
+            sleep 10
+            echo "Still running chroot's rc.local..." 2>&1
+            while true 
+            do
+                    sleep 1
+            done
+        ) & forkpid=$! 
         chrootcmd 'exec /etc/rc.local >/dev/null 2>/dev/null </dev/null' \
             || ret=$?
+        kill "$forkpid"
```
dnschneid: `wait "$forkpid"` after kill so that the process gets cleaned up.

host-bin/enter-chroot
```diff
@@ -684,10 +684,19 @@ if [ -n "$NOLOGIN" ]; then
...
+    # Check and run rc.local
     if [ -n "$firstrun" -a -x "$CHROOT/etc/rc.local" ]; then
+        ( 
+            sleep 10
+            echo "Still running chroot's rc.local..." 2>&1
+            while true 
+            do
+                    sleep 1
+            done
+        ) & forkpid=$!
```
dnschneid: Also, you may want to `addtrap "kill '$forkpid' 2>/dev/null"` and then `undotrap` after chrootcmd, so that a ctrl-c kills the forked process as well.

host-bin/enter-chroot
```diff
@@ -684,10 +684,18 @@ if [ -n "$NOLOGIN" ]; then
...
+        ( 
+            sleep 10
+            echo "Still running chroot's rc.local..." 2>&1
+        ) & forkpid=$! 
         chrootcmd 'exec /etc/rc.local >/dev/null 2>/dev/null </dev/null' \
             || ret=$?
+        addtrap "kill '$forkpid' 2>/dev/null"
+        kill "$forkpid"
+        wait "$forkpid"
+        undotrap
```
tonyxue: @dnschneid Updated. Am I using the addtrap and undotrap correctly? should be this way

tonyxue:
@dnschneid 
>wait "$forkpid" after kill so that the process gets cleaned up.
Ok.. this is the bit causing the problem and the bit I don't get.
Why wait for a process that is already killed? This would return an error.

dnschneid:
Sorry, it goes:

1. addtrap
2. chrootcmd
3. undotrap
4. kill
5. wait

Since the script is the parent process, it needs to wait for the child to get the return code and let it get cleaned up (otherwise you end up with a zombie process).  You should `wait pid || true` so the error is ignored.

host-bin/enter-chroot
```diff
@@ -684,10 +684,18 @@ if [ -n "$NOLOGIN" ]; then
         fi
     fi
 else
-    # Run rc.local
+    # Check and run rc.local
     if [ -n "$firstrun" -a -x "$CHROOT/etc/rc.local" ]; then
+        ( 
+            sleep 10
+            echo "Still running chroot's rc.local..." 2>&1
+        ) & forkpid=$!
```
dnschneid: separate line

dnschneid: Also, please use a better variable name, such as `monitorpid`

tonyxue:
updated and it is working

tonyxue:
@dnschneid so if nothing else to fix, I hope this can be merged.

host-bin/enter-chroot
```diff
@@ -684,10 +684,20 @@ if [ -n "$NOLOGIN" ]; then
         fi
     fi
 else
-    # Run rc.local
+    # Check and run rc.local
     if [ -n "$firstrun" -a -x "$CHROOT/etc/rc.local" ]; then
+
```
dnschneid: I think this would be more readable (and without spacer lines around it):

```
(sleep 10; echo "Still running chroot's rc.local..." 2>&1) &
noticepid="$!"
```

dnschneid:
One last style nit (sorry for the delay)

host-bin/enter-chroot
```diff
@@ -684,10 +684,16 @@ if [ -n "$NOLOGIN" ]; then
         fi
     fi
 else
-    # Run rc.local
+    # Check and run rc.local
     if [ -n "$firstrun" -a -x "$CHROOT/etc/rc.local" ]; then
+        ( sleep 10;echo "Still running chroot's rc.local..." 2>&1 ) &
```
tonyxue: @dnschneid Is it good in this way? Sorry for being not good at coding style and thanks for helping me correct it :)

tedm:
@tonyxue I hope this isn't too far off topic, but could you tell me if my rc-local info for mysql is optimized below? I don't understand exactly why I need all that, when sometimes it will start with just 

`/usr/sbin/mysqld &`

also that commented out pure-ftpd line seems to prevent entering chroot. I copied that line from when pure-ftpd first installed and ran itself. I can probably debug that one. Am curious about the line above

`/usr/sbin/mysqld &` though

# By default this script does nothing.

/sbin/iptables -P INPUT ACCEPT


mkdir -p -m0755 /var/run/sshd
/usr/sbin/sshd

#/usr/sbin/vsftpd

#/usr/sbin/pure-ftpd -l pam -E -u 1000 -O clf:/var/log/pure-ftpd/transfer.log -8 UTF-8 -

service apache2 start

export HOME=/etc/mysql
umask 007
[ -d /var/run/mysqld ] || install -m 755 -o mysql -g root -d /var/run/mysqld
/usr/sbin/mysqld &

exit 0

tonyxue:
@tedm So why don't you start all your applications using ```service```? I think that's the easiest way.

> ```export HOME=/etc/mysql```

And why are you doing this one? This is making all login users' HOME variable to ```/etc/mysql```, you sure you want to do that?

tedm:
@tonyxue I'll try again with service on these, but there was a time, when service [server] start didn't always work the way it is working with apache.

Since the method of starting mysqld from the wiki page has been working, I've been using that. I'll try alternatives, but for many crouton versions, unless I was making an error in /etc/rc.local, the /usr/sbin/[service] has been the most stable, but as I mentioned, was somehow not reliable with mysqld, but wiki method is. 

My env and echo $HOME is still /home/ted, I think it is just setting the home for mysql which gets a low system user id in /etc/passwd:

`mysql:x:104:109:MySQL Server,,,:/nonexistent:/bin/false`

mysql is working fine, at a user command stand-alone with:

`$mysql -u root -p`

to enter, as well as apps on top like phpmyadmin and wordpress, so I think it's OK, but am curious about the explanation of that syntax for rc.local.

tonyxue:
@tedm 
>Since the method of starting mysqld from the wiki page has been working, I've been using that.

Sorry didn't realize that's from the wiki. I thought you wrote the commands and just checking whether you know what you are doing...

I think the method employed in the wiki is just trying to cope with any possible situation of installing mysql as not everyone installs it by apt. If you can start your mysqld by service, then better do it I think.

>but am curious about the explanation of that syntax for rc.local.

ok so the syntax of this one: ```[ -d /var/run/mysqld ] || install -m 755 -o mysql -g root -d /var/run/mysqld``` ? That's generally making sure that the directory exist. If not then create it with the correct ownerships and permissions. I think it is trying to avoid errors from starting up mysqld.

tedm:
@tonyxue Thanks for the explanation Tony. I've just now taken commented out the wiki mysql start process and manually tried these methods:

(precise)root@localhost:/etc/init.d# /etc/init.d/mysql start
Rather than invoking init scripts through /etc/init.d, use the service(8)
utility, e.g. service mysql start
initctl: Unknown job: mysql

Since the script you are attempting to invoke has been converted to an
Upstart job, you may also use the start(8) utility, e.g. start mysql
(precise)root@localhost:/etc/init.d# service mysql start
start: Unknown job: mysql
(precise)root@localhost:/etc/init.d# /etc/init.d/mysqld
bash: /etc/init.d/mysqld: No such file or directory
(precise)root@localhost:/etc/init.d# /usr/sbin/mysqld &
[1] 5981
(precise)root@localhost:/etc/init.d# 150429  8:44:47 [Warning] Using unique option prefix key_buffer instead of key_buffer_size is deprecated and will be removed in a future release. Please use the full name instead.
150429  8:44:47 [Note] /usr/sbin/mysqld (mysqld 5.5.43-0ubuntu0.12.04.1) starting as process 5981 ...

but, the last process above isn't starting it, although if I try several times, it sometimes does, so I'm wondering if the tests the wiki does is adding a bit of time, as the directory does exist, is owned by root:root, and has permissions of 755 (mysdqld in /usr/sbin).

tonyxue:
>Unlike apache and sshd, there are no mysql files in /var/run 

I've not installed mysql in crouton but that's the case. It uses install to create the directory first.

tonyxue:
@dnschneid 

Anything else to fix or improve?

host-bin/enter-chroot
```diff
@@ -684,10 +684,16 @@ if [ -n "$NOLOGIN" ]; then
...
+        noticepid="$!"
+        addtrap "kill '$noticepid' 2>/dev/null"
         chrootcmd 'exec /etc/rc.local >/dev/null 2>/dev/null </dev/null' \
             || ret=$?
+        undotrap
+        kill "$noticepid" 2>/dev/null
```
dnschneid: Whoops, this probably needs an || true.

host-bin/enter-chroot
dnschneid: Just missing a space after the semicolon

tonyxue:
Updated. Thanks for pointing that out. Didn't notice that.

dnschneid:
Thanks!  Works marvelously.

tonyxue:
I should thank you!!! I am very impressed by your patience and high standard!!

Issue #1603, opened 2015-04-10
Gnome Doesn't Play Nice w/ setres (xrandr)


sprc:
Gnome doesn't want to play nice with the setres script when trying to switch to a custom resolution (such as 1280x850 on a Chromebook Pixel). Similar results when attempting to pick a custom resolution using xrandr --output eDP1 --mode <custom mode>. It flickers to the specified resolution but immediately switches back. You can then manually go select the resolution from the Display activity, but that's not ideal - I'm trying to script this for a remote desktop scenario.

Interestingly, after setting the custom mode of 1280x850, the same thing happens when trying to switch back to a standard resolution.

As a side note, xrandr --output eDP1 --auto --scale 0.5x0.5 works without needing to add a custom mode for 1280x850, but is tangibly fuzzier due to what I'm guessing is a software filter vs hardware filter.

dnschneid:
What if you just do the --addmode part of the setres script, and then issue whatever gnome-settings command is necessary to switch the resolution "the gnome way"?

sprc:
I haven't had any luck finding a command to switch the resolution "the gnome way", unfortunately.

sprc:
Hmm So I've discovered that if you deactivate the xrandr gnome-settings-daemon plugin:

gsettings set org.gnome.settings-daemon.plugins.xrandr active false

...then do your xrandr command, the custom resolution sticks.

I'm guessing this has other ramifications but I don't know what they are. Also, using gsettings to reactivating the xrandr daemon plugin will instantly punt you back out of the custom resolution.

dnschneid:
Well, that probably makes the GUI not work for changing display settings.  If you're happy with setres, then that's probably fine...

sprc:
A Gnome dev mentioned taking a look at fixing this someday, and mentioned we could do a couple of things to work around it in the meantime without disabling the xrandr daemon plugin:

1) Explicitly tell mutter directly that we want a new configuration by using the org.gnome.Mutter.DisplayConfig D-Bus interface, via gdbus or dbus-send. This is theoretically the most 'correct' way to accomplish this, but it looks pretty complex and my google-fu tells me no one's actually managed to do this yet via shell script.

2) Modify ~/.config/monitors.xml to show our new resolution and then make the xrandr change.

[...]

PR #1666, opened 2015-04-25
Add handling for custom resolution workaround for GNOME

sprc:
GNOME Bug 747750: Gnome instantly reverts resolution change if made directly via xrandr and resolution is 'custom'. We work around this by modifying monitors.xml to show the resolution we want, changing it, waiting a couple of seconds, and reverting monitors.xml back to its original condition. It's kind of gross but it works.

Shell scripting is admittedly not my main language, so there may be more efficient ways of doing this. It also may be large enough to merit being modularized somehow.

I have not yet implemented a flag to permenantly write the monitors.xml change but plan to at some point.

### DIFF
--- a/chroot-bin/setres
+++ b/chroot-bin/setres
@@ -35,8 +35,81 @@ if [ "${xmethod%%-*}" != "xiwi" ]; then
         mode="${mode%\"}"
         xrandr --newmode "$mode" $data 2>/dev/null || true
         xrandr --addmode "$o" "$mode"
-        xrandr --output "$o" --mode "$mode"
-        echo "$mode"
+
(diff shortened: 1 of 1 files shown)


sprc:
Relevant issue is [Gnome Doesn't Play Nice w/ setres (xrandr) (#1603)](https://github.com/dnschneid/crouton/issues/1603)

sprc:
Doesn't seem like there's anyone else looking for this. Shall we close for now and revisit if there's interest down the line?

dnschneid:
Sure, that's reasonable.  Thanks for the patch; it's good to have in case we need it down the road.

sprc:
FYI, this approach doesn't appear to work in later versions of GNOME. If we do ever decide we want to try to address the issue, we'll have to find another way.

PR #1689, opened 2015-04-29
Changed gnome-session-fallback to gnome-session-flashback, same thing bu...

Voronsky:
Sorry for the comment cutting off. 

Gnome-session-fallback is no longer found in the current vivid repository, the last one instance of it found was for utopic. Also it seems that "flashback" is to be the new standard for support of GNOME 3, and thus if one tries to find "gnome-session-fallback" you get this

<code>Package gnome-session-fallback is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  gnome-session-flashback

E: Package 'gnome-session-fallback' has no installation candidate</code>

and the chroot install for gnome will fail to complete the setup. Changing gnome-session-fallback to flashback will fix this.


targets/gnome
```diff
@@ -10,7 +10,7 @@ CHROOTBIN='crouton-noroot startgnome gnome-session-wrapper'
 
 ### Append to prepare.sh:
 install --minimal evolution-data-server gnome-control-center \
-        gnome-screensaver gnome-session gnome-session-fallback \
+        gnome-screensaver gnome-session gnome-session-flashback \
```
dnschneid: `gnome-session-flashback` doesn't exist prior to trusty or in wheezy/kali, so you need to do: `gnome-session-flashback,ubuntu~precise+ubuntu~quantal+ubuntu~raring+ubuntu~saucy+debian~wheezy+kali=gnome-session-fallback`....which really isn't the best syntax.

Alternatively, you could set a variable with the package name ahead of time, and use `if release -lt trusty -lt wheezy -eq kali` to see when fallback is necessary.

tonyxue: Isn't these two linked to each other in the repo?

Voronsky: @dnschneid  thats right, that just occurred to me right after I did the pull request, since the flashback seems to be a package that will be used as the default for the latest releases of ubuntu and GNOME and onward, but not necessarily the case for the previous prior to trusty. I'll give the alternative a shot. Thanks!

dnschneid:
Also, could you please sign the [CLA](https://cla.developers.google.com/clas/new?kind=KIND_INDIVIDUAL&domain=DOMAIN_GOOGLE)?

Voronsky:
Yup, just signed it!

dnschneid:
I think this PR is still pending fixes.

divx118:
@Voronsky Any news on the pending fixes for this PR. It is a bit of a pane when testing newer releases with the gnome target when this isn't merged yet. Thanks.

Voronsky:
So I deeply apologize for my extremely late response to this, long story short, my RL got a bit crazy and I forgot I had opened this for this repo. The new merge request contains a function that uses `lsb_release` to check what distro is being used and whether it is one known to require fallback or flashback when installing gnome. I am not familiar with POSIX until I wrote what is in the merge request. The idea is since the OS gets installed first, from there check what codename is it, and then check the set array - which contains known fallback needing distros - and if it is one fallback distros, install with gnome-session-fallback. Otherwise install the the previous packages but this time added gnome-session-flashback. I felt this way it would be easier to make a set so future users can simply add any distro name missing that only has gnome-session-fallback and not flashback in their repos.

Currently I have tested with the following
-Trusty
-Precise
-Wheezy
-Xenial

Apparently Jessie and Sid use flashback, so I did not add those into the set of fallback distros.

targets/gnome
```diff
@@ -9,11 +9,32 @@ CHROOTBIN='crouton-noroot startgnome gnome-session-wrapper'
...
+    set "squeeze" "wheezy" "kali" "trusty" "precise" 
+    name=$(lsb_release -cs)
+    #echo "$name"
+    for i do
+	if [ "$i" = "$name" ]; then
+	   echo "1" #The distro installed needs gnome-session-fallback
+	fi
+    done
+}
+fallback_need=$(check_distro)
```
divx118: You probably didn't know this, but there is a global release check function https://github.com/dnschneid/crouton/blob/master/installer/functions#L123
So you can just use something like
 ```
if release -ge wheezy -ge kali -ge trusty; then
    package-gnome-session="gnome-session-flashback"
else
   package-gnome-session="gnome-session-fallback"
fi
install --minimal evolution-data-server gnome-control-center \
          gnome-screensaver gnome-session $package-gnome-session \
          gnome-shell gnome-themes-standard gvfs-backends nautilus \
          unzip gtk2-engines-pixbuf
``` 
As you see I used a variable so you don't need the repetitive code below.
Thanks for contributing.

Voronsky: @divx118  thank you! Your solution is way more beautiful and the release function is definitely helpful and I was unawares of such utility. Thanks again, may I write a comment to credit you?

divx118: You don't have to credit me :), I will already be happy if we get another contributor on board. I just wanted to point you in the right direction.

Voronsky:
@divx118  and @dnschneid  this PR is updated to reflect the changes discussed throughout this conversation, thanks again guys!

SoccerPie:
I'm trying to install Xenial with a gnome target. Could someone please give me a crouton file that has fallback changed to flashback, because I don't know how to install crouton using the master file I edited. Please and thank you guys.

alexluecke:
@SoccerPie 

From the Crouton readme:

> If you're modifying crouton, you'll probably want to clone or download the repo and then either run installer/main.sh directly, or use make to build your very own crouton. You can also download the latest release, cd into the Downloads folder, and run sh crouton -x to extract out the juicy scripts contained within, but you'll be missing build-time stuff like the Makefile.

From this info we know we can clone the repository and make any changes we want. You can modify the `targets/gnome`  target under the repository root, specifically line 13 where `gnome-session-fallback` should be changed to `gnome-session-flashback`. After making this change, we need to use the `installer/main.sh` method suggested in lieu of the usual `sudo sh ~/Downloads/crouton` command. Run the following to install gnome (with whatever additional targets you want):

    sudo sh installer/main.sh -r xenial -t gnome 

I successfully installed gnome with the xenial release using this method.

PR #1693, opened 2015-04-30
Added ctrl+esc / ctrl+shft+esc as alternative to ctrl+alt+shft+forward /...

penntaylor:
`ctrl+alt+shift+forward / backward` is too hard to use with one hand, and the right hand has to go waaaay over there to the left, too far away from its snuggly home. So, `ctrl+esc` to trigger `croutoncycle next` and `ctrl+shift+esc` to trigger `croutoncycle prev`.


dnschneid:
I'd rather provide a simple means to configure the keyboard shortcut than provide arbitrary alternatives.  Thanks for the patch, though.

Issue #1664, opened 2015-04-23
Colors in console output


dnschneid:
I think we need to add colors to the console output.  Not because they're fancy, and pretty, and ooooooooh shiny...

...where was I?  Right, colors.  Really, we need colors so that the super-helpful crouton messages stand out from all the junk that's usually ignored.  Once you get past 3 or 4 lines of text everyone's eyes just glaze over and nobody notices crouton's error messages or whatever.

Color might help, or at least look nicer when being ignored.

In particular, we need red when installs or updates fail, so people don't think their chroots are up-to-date.

Tips might be nice in blue or green or whatever's visible but not too distracting.

DennisLfromGA:
Yup, that'd be really helpful.
In this same vane (maybe), it would be handy to have a debug option that would `set -x` or something so you wouldn't have to instruct them on the use of `2>&1 | tee ~/Downloads/trusty-debug.log`

DennisLfromGA:
And for extra bling:
```
    bold=$(     tput bold   || tput md      ) # Start bold
    stout=$(    tput smso   || tput so      ) # Start stand-out
    estout=$(   tput rmso   || tput se      ) # End stand-out
    under=$(    tput smul   || tput us      ) # Start underline
    eunder=$(   tput rmul   || tput ue      ) # End   underline
    italic=$(   tput sitm   || tput ZH      ) # Start italic
    eitalic=$(  tput ritm   || tput ZR      ) # End   italic
```

tedm:
is this a pie?

tonyxue:
@tedm Well, I'd rather say this is many pies :P

DennisLfromGA:
@ted - Add red for cherry pie, green for apple pie, etc. ;)

tedm:
@DennisLfromGA Thanks, but I looked at the related two issues, and I think there is some kind of color defined table in the works. Classic case of Rube Goldberg-ness ;)

DennisLfromGA:
@tedm - Sorry, I was only attempting humor and failed I suppose - I wasn't being serious... ;)

tedm:
@DennisLfromGA I get it! lol ;) Seeing all these colors coming, and being used to only black and white bash terminals, I've ordered a Toshiba FHD 13" Intel 4GB/16GB refurb from Groupon. Should be here mid-week.

DennisLfromGA:
On Sun, May 31, 2015 at 5:04 PM, tedm <notifications@github.com> wrote:

[...]

PR #1710, opened 2015-05-03
Adding echo_color function to support colors in the shell output.

divx118:
Well I added some color, just to the error echo and usage. red only when the exit code is 1 all other the color will be yellow. blue for the usage. This is just a simple way to add some color. To be more flexible we should probably write a little echo function. See #1664

### DIFF
--- a/installer/functions
+++ b/installer/functions
@@ -47,7 +47,7 @@ error() {
         # color yellow 33m
         color="33m"
     fi
-    echo "\033[$color$*\033[" 1>&2
+    echo "\033[$color$*\033[0m" 1>&2
     exit "$ecode"
 }
 
--- a/installer/functions
+++ b/installer/functions
@@ -43,8 +43,8 @@ error() {
     # color red 31m
     local color="31m"
     shift
-    if [ $ecode -ne 1 ]; then
-        # color yellow 33m"
+    if [ "$ecode" -ne 1 ]; then
+        # color yellow 33m
         color="33m"
     fi
     echo "\033[$color$*\033[" 1>&2
--- a/installer/functions
+++ b/installer/functions
@@ -40,14 +40,12 @@ installscript() {
 # Exits the script with exit code $1, spitting out message $@ to stderr
 error() {
     local ecode="$1"
-    # color red 31m
-    local color="31m"
+    local color="lred"
     shift
     if [ "$ecode" -ne 1 ]; then
-        # color yellow 33m
-        color="33m"
+        color="lyellow"
     fi
-    echo "\033[$color$*\033[0m" 1>&2
+    echo_color "$color" "$*"
     exit "$ecode"
 }
 
@@ -224,6 +222,41 @@ echo_tty() {
     else
         echo "$@" 1>&2
     fi
+}
+
+# Set a console color (shorthand)
+# usage: echo_color [l][color] [string]
+# Specifying "l" uses the light version of the color.
+# Color can be red, green, yellow, blue, magenta, cyan.
+# Color can be specified with just the first letter. 
+# example: echo_color "lr" "light red"
+echo_color() {
+    # If not outputting to a tty print no colors.
+    if [ ! -t 2 ]; then
+        shift
+        echo "$*" 1>&2 
+        return
+    fi
+    printf "\033["
+    local c="$1"
+    if [ "${c#l}" != "$c" ]; then
+        printf "1;"
+        c="${c#l}"
+    else
+        printf "0;"
+    fi
+    case "$c" in
+        r*) printf "31m";;
+        g*) printf "32m";;
+        y*) printf "33m";;
+        b*) printf "34m";;
+        m*) printf "35m";;
+        c*) printf "36m";;
+         *) printf "37m";;
+    esac
+    shift
+    echo "$*" 1>&2
+    printf "\033[0m"
 }
 
 # Websocket interface
--- a/installer/functions
+++ b/installer/functions
@@ -40,9 +40,11 @@ installscript() {
 # Exits the script with exit code $1, spitting out message $@ to stderr
 error() {
     local ecode="$1"
+    # color red 31m
     local color="31m"
     shift
-    if [ ! $ecode -eq 1 ]; then
+    if [ $ecode -ne 1 ]; then
+        # color yellow 33m"
         color="33m"
     fi
     echo "\033[$color$*\033[" 1>&2
--- a/installer/functions
+++ b/installer/functions
@@ -40,12 +40,12 @@ installscript() {
 # Exits the script with exit code $1, spitting out message $@ to stderr
 error() {
     local ecode="$1"
-    local color="lred"
     shift
-    if [ "$ecode" -ne 1 ]; then
-        color="lyellow"
+    if [ "$ecode" -eq 1 ]; then
+        echo_color "r" "$*" 1>&2
+    else
+        echo "$*" 1>&2
     fi
-    echo_color "$color" "$*"
     exit "$ecode"
 }
 
@@ -131,18 +131,18 @@ alias getopts_nextarg='getopts_ret=1
 # Example:  release -ge quantal -ge wheezy
 release() {
     if [ "$(($# % 2))" -ne 0 ]; then
-        error 3 "invalid parameters to release(): $*"
+        error 3 "$(echo_color "y" "invalid parameters to release(): $*")"
     fi
     # Load up the list of releases; this will be replaced with a literal list
     local releases="`cat "$DISTRODIR/releases" 2>/dev/null`"
     if [ -z "$releases" ]; then
-        error 3 "list of releases for $DISTRO not found"
+        error 3 "$(echo_color "y" "list of releases for $DISTRO not found")"
     fi
     # End-of-word regex for awk
     local eow='([^a-z]|$)'
     local relnum="`echo "$releases" | awk "/^$RELEASE$eow/ {print NR; exit}"`"
     if [ -z "$relnum" ]; then
-        error 3 "$RELEASE not found in $DISTRO"
+        error 3 "$(echo_color "y" "$RELEASE not found in $DISTRO")"
     fi
     while [ "$#" -ge 2 ]; do
         local cmp="`echo "$releases" | awk "/^$2$eow/ {print NR; exit}"`"
@@ -224,24 +224,24 @@ echo_tty() {
     fi
 }
 
-# Outputs colored text to stderr
+# Outputs colored text to stdout
 # usage: echo_color [l][color] [string]
-# Specifying "l" uses the light version of the color.
+# Specifying "t" (thick) uses bold text.
 # Color can be red, green, yellow, blue, magenta, cyan.
 # Color can be specified with just the first letter. 
-# example: echo_color "lr" "light red"
+# example: echo_color "tr" "bold red"
 echo_color() {
     # If not outputting to a tty print no colors.
     if [ ! -t 2 ]; then
         shift
-        echo "$@" 1>&2
+        echo "$@"
         return
     fi
     printf "\033["
     local c="$1"
-    if [ "${c#l}" != "$c" ]; then
+    if [ "${c#t}" != "$c" ]; then
         printf "1;"
-        c="${c#l}"
+        c="${c#t}"
     else
         printf "0;"
     fi
@@ -260,7 +260,7 @@ echo_color() {
         s=''
         shift
     fi
-    echo '-n' "$@" 1>&2
+    echo '-n' "$*"
     printf "\033[0m$s"
 }
 
--- a/installer/functions
+++ b/installer/functions
@@ -40,8 +40,12 @@ installscript() {
 # Exits the script with exit code $1, spitting out message $@ to stderr
 error() {
     local ecode="$1"
+    local color="31m"
     shift
-    echo "$*" 1>&2
+    if [ ! $ecode -eq 1 ]; then
+        color="33m"
+    fi
+    echo "\033[$color$*\033[" 1>&2
     exit "$ecode"
 }
 
--- a/installer/functions
+++ b/installer/functions
@@ -224,7 +224,7 @@ echo_tty() {
     fi
 }
 
-# Set a console color (shorthand)
+# Outputs colored text to stderr
 # usage: echo_color [l][color] [string]
 # Specifying "l" uses the light version of the color.
 # Color can be red, green, yellow, blue, magenta, cyan.
@@ -234,7 +234,7 @@ echo_color() {
     # If not outputting to a tty print no colors.
     if [ ! -t 2 ]; then
         shift
-        echo "$*" 1>&2 
+        echo "$@" 1>&2
         return
     fi
     printf "\033["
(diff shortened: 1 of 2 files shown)


installer/functions
```diff
@@ -40,8 +40,12 @@ installscript() {
 # Exits the script with exit code $1, spitting out message $@ to stderr
 error() {
     local ecode="$1"
+    local color="31m"
     shift
-    echo "$*" 1>&2
+    if [ ! $ecode -eq 1 ]; then
```
dnschneid: `if [ "$ecode" -ne 1 ]; then`

installer/functions
```diff
@@ -40,8 +40,12 @@ installscript() {
 # Exits the script with exit code $1, spitting out message $@ to stderr
 error() {
     local ecode="$1"
+    local color="31m"
```
dnschneid: Could you put comments to mark what the colors are?

divx118:
Line comments added and adjusted the $ecode check.

dnschneid:
Considering #1753 (and the fact that echo "\escapes" isn't portable), I think it'd be worth it to make either a bunch of global constants like $RED and $BLACK, or to put something like the following function (that I used once in a random project) into installer/functions (@drinkcat for style suggestions):

```
# Set a console color (shorthand)
# usage: c [l][color]
# Specifying "l" uses the light version of the color.
# Color can be black, green, cyan, red, purple, or orange.
# Color can be specified with just the first letter. 
# example: echo "$(c lr)light red $(c g)green $(c)normal color"
c() {
    printf "\033["
    if [ -z "${1:-}" ]; then
        printf "0m"
        return
    fi
    local c="$1"
    if [ "${c#l}" != "$c" ]; then
        printf "1;"
        c="${c#l}"
    else
        printf "0;"
    fi
    case "$c" in
        b*) printf "34m";;
        g*) printf "32m";;
        c*) printf "36m";;
        r*) printf "31m";;
        p*) printf "35m";;
        o*) printf "33m";;
         *) printf "30m";;
    esac
}
```

tonyxue:
I immediately realized after opening #1753 that we need some sort of 'Universal Coloring Engine' so that everyone is outputting in the same format and color.

We can create a function like ```crouton_echo``` to replace the default ```echo``` usage so we can do ```crouton_echo blue "Hello World"``` .

I think the global variables may not seem so smart.

dnschneid:
That has potential too...you could always mix-and-match colors by using -n as well.  "crouton_echo" isn't very descriptive though; how about `echo_color`?

I'm ambivalent whether we take the `echo_color` approach or the `$(c)` approach described above.

divx118:
I prefer the $(c) approach. This has the advantage when only needing to highlight one word it is possible without splitting up the echo lines in the code.

tonyxue:
Sure,  the name of the function I gave isn't the most suitable one.  I just randomly came up with one for expressing the idea.

drinkcat:
> I prefer the $(c) approach. This has the advantage when only needing to highlight one word it is possible without splitting up the echo lines in the code.

Is it something we plan to do? I'm not sure ,-) Also, `c` makes it easier to forget to reset the color at the end of the line... Acually, I wouldn't necessarily mind having both `echo_color` and `c` functions (with `echo_color` as a simple wrapper). We could even have a `cat_color`, if that's useful in some contexts.

Finally, we want to check that the output is a tty before outputing colors, so we don't introduce binary characters when stdout/err is redirected to a file. (that's what `ls` does, for example)

divx118:
@drinkcat You are right we don't want to make the terminal like a candy shop :) I created an echo_color function now.

installer/functions
```diff
@@ -220,6 +224,41 @@ echo_tty() {
     fi
 }
 
+# Set a console color (shorthand)
```
dnschneid: `# Outputs colored text to stderr`

installer/functions
```diff
@@ -220,6 +224,41 @@ echo_tty() {
...
+# usage: echo_color [l][color] [string]
+# Specifying "l" uses the light version of the color.
+# Color can be red, green, yellow, blue, magenta, cyan.
+# Color can be specified with just the first letter. 
+# example: echo_color "lr" "light red"
+echo_color() {
+    # If not outputting to a tty print no colors.
+    if [ ! -t 2 ]; then
+        shift
+        echo "$*" 1>&2
```
dnschneid: Use `"$@"` so that -n works

installer/functions
```diff
@@ -220,6 +224,41 @@ echo_tty() {
...
+        r*) printf "31m";;
+        g*) printf "32m";;
+        y*) printf "33m";;
+        b*) printf "34m";;
+        m*) printf "35m";;
+        c*) printf "36m";;
+         *) printf "37m";;
+    esac
+    shift
+    echo "$*" 1>&2
```
dnschneid: ditto

installer/functions
```diff
@@ -220,6 +224,41 @@ echo_tty() {
...
+    printf "\033[0m"
```
dnschneid: It'd be nice to squeeze this in before the newline, so that line buffering doesn't result in weird colors if multiple things are outputting to the terminal

dnschneid: That may involve manually interpreting -n, by the way.

targets/post-common
```diff
@@ -68,9 +68,9 @@ else
 fi
 
 if [ -n "$TIPS" ]; then
-    echo "
-Here's some tips:
-$TIPS" 1>&2
+    echo_color "lblue" \
```
dnschneid: there's an extra newline in there that you've lost.

dnschneid:
Actually, I think USAGE is one such example where we wouldn't want the entire echo to be a single block of color...I'm not sure it should be any color at all, since it's the only text being spit out.

installer/functions
divx118: Oops overlooked that will change it.

installer/functions
divx118: Do we also need support for -E? Else I could just check for "-n" on $1 after the shift to determine if I need a new line after the printf.

targets/post-common
divx118: Another Oops.

divx118:
@dnschneid Yep, I know USAGE is indeed something I am also not happy with. It should be no color like you said. Let me think about it.

installer/functions
```diff
@@ -220,6 +224,46 @@ echo_tty() {
...
+# Color can be specified with just the first letter. 
+# example: echo_color "lr" "light red"
+echo_color() {
+    # If not outputting to a tty print no colors.
+    if [ ! -t 2 ]; then
+        shift
+        echo "$@" 1>&2
+        return
+    fi
+    printf "\033["
```
drinkcat: `1>&2` ?

installer/functions
```diff
@@ -220,6 +224,46 @@ echo_tty() {
     fi
 }
 
+# Outputs colored text to stderr
```
drinkcat: Or we could just output to stdout and let the caller redirect... That's more consistent with how `echo` works.

installer/functions
tonyxue: @drinkcat ```1>&2``` here? It's only part of the output.

installer/functions
dnschneid: That's a good point.  We should probably keep the redirect explicit.

installer/functions
```diff
@@ -220,6 +224,46 @@ echo_tty() {
...
+        c*) printf "36m";;
+         *) printf "37m";;
+    esac
+    shift
+    local s='\n'
+    if [ "$1" = '-n' ]; then
+        s=''
+        shift
+    fi
+    echo '-n' "$@" 1>&2
```
dnschneid: This one should be $*, since you've eliminated the first parameter

dnschneid:
So in conclusion: echo_color shouldn't have any redirects inside the function, but usages of echo_color should as appropriate (basically, as they did before with echo).

dnschneid:
This also means you can technically `error 1 "$(echo_color r "ERROR")"`...

divx118:
>This also means you can technically error 1 "$(echo_color r "ERROR")"...

Is this something we want? Or do we leave color red by default in the error function for exit code 1. I like removing the default yellow color for other exit codes > 1. So we can also display the usage message with no color and add the echo_color y to where it is needed.

divx118:
>Is this something we want? Or do we leave color red by default in the error function for exit code 1. I like removing the default yellow color for other exit codes > 1. So we can also display the usage message with no color and add the echo_color y to where it is needed.

Sorry for being impatient, but any feedback on this?

dnschneid:
Sorry, red by default for exit code 1, and no colors otherwise makes sense.  Our usage of error codes appears to be as follows:

0: success (duh)
1: processing error
2: user error (bad parameter to script)
3: coding error (bad parameter to function)
99: target doesn't work on platform (specially handled by tests)

99 should probably be a color too (orange?)...

divx118:
We only seem to have the following colors available (missing orange ;)) See http://ascii-table.com/ansi-escape-sequences.php
```
Foreground colors
30	Black
31	Red
32	Green
33	Yellow
34	Blue
35	Magenta
36	Cyan
37	White
```
```
Background colors
40	Black
41	Red
42	Green
43	Yellow
44	Blue
45	Magenta
46	Cyan
47	White
```
I also see that I should change the comments in the code, because `1;` turns on bold `0;` regular text
See the following screenshot for an example:

 Alt text

`for i in 30m 31m 32m 33m 34m 35m 36m 37m 40m 41m 42m 43m 44m 45m 46m 47m; do printf "\033[1;$i"; echo color; printf "\033[0;0m"; done`

dnschneid:
Whoops.  Yellow works too then :)

divx118:
Al ready added some yellow color to the `error 3` in functions. All the other error calls needs still to be done in files where they are present. I thought of creating a different pull request for that. To keep this a bit clean.
Color codes the way I see it:
 * Red for all exit 1 errors.
 * Yellow exit 2, 3, 99 will depend on the message
 * Blue usage messages and prompts.

There are also some [WARNING messages](https://github.com/divx118/crouton/blob/color/installer/main.sh#L248), I thought of giving them Magenta color. 

Feel free to add any input about color usage... Also should we support background colors in `echo_color`, or just stick to bold/regular text colors?

tonyxue:
> Also should we support background colors in echo_color, or just stick to bold/regular text colors?

I think only colored text would be efficient. The mixed usage of text color and background color may make the interface looks too colorful and confuse the users.

divx118:
@tonyxue yep I agree, that is why I left them out. 
BTW to clarify with usage messages for color blue I meant for example $TIPS. So not the $USAGE they should be just normal color.

tonyxue:
But if all parts of the ```$TIPS``` is in the same color, how could we make message like [this](https://github.com/dnschneid/crouton/blob/master/targets/extension#L51) stands out? Or do we need to make it especially stand out?

As https://github.com/dnschneid/crouton/issues/1804#issuecomment-107759896 that the user obviously didn't see that. And if it still blends with other messages, the users will still not see that.

So the point is we add color to make some part of the output stand out, so we should not employ colors in a large scale. But I am not sure whether ```$TIPS``` is *large scale*. People will still ignore the download link if they choose not to read ```$TIPS``` at all. We should only use a small amount of colors precisely.

tonyxue:
@divx118 And maybe we can modify the title of this PR a bit? This PR is not as simple as the title is describing now. :P

divx118:
@tonyxue It should be possible to highlight some special messages like the one you described with `$(echo_color "m" "You must install the Chromium OS extension for integration with cro...")`. I still think $TIPS should overall have a color, or at least start with a color so you can see the script has ended and you should start reading from the first color line.

divx118:
How about this, we only highlight "Here's some tips:" blue then we highlight only messages in $TIPS blue when they need further user action like the example you posted.

tonyxue:
> How about this, we only highlight "Here's some tips:" blue then we highlight only messages in $TIPS blue when they need further user action like the example you posted.

That sounds nice. +1

tonyxue:
Any update on this?

And any suggestion on adding color for [this](https://github.com/dnschneid/crouton/blob/master/installer/main.sh#L100) ?
Although not many people care about this, it is still essential to know.

dnschneid:
We could colorize it, I suppose, as long as it's not obnoxious.

tonyxue:
Ping!

Are we adding more colors or colored texts? It seems good to merge now.
We could colorize the rest of the texts later.

divx118:
Seems we have now everything to merge. Or are there still any other thoughts ...

DennisLfromGA:
One additional thought/suggestion: Could we add a color, possibly white, to any 'WARNING: ..." messages? 

(I.E.) echo "WARNING: $RELEASE is an unsupported release.

They get spit out early in the install and I think they are often overlooked.

divx118:
@DennisLfromGA That is indeed possible, but those messages can be changed in different pull request(s).  
After this merge, something like the following will highlight WARNING: in yellow color.
```shell
echo "$(echo_color "y" "WARNING:") " $TARBALL is an old-style bootstrap or backup." 1>&2
```

dnschneid:
Sorry for taking so long to merge this.  Go ahead and colorize other messages within reason :)

DennisLfromGA:
@divx118,

> After this merge, something like the following will highlight WARNING: in yellow color.

I was thinking we should maybe steer away from yellow since we're intending to use that for *exit* errors 2,3 & 99.
We could use magenta like you suggested earlier for warnings, I didn't notice your post (https://github.com/dnschneid/crouton/pull/1710#issuecomment-107233202) when I had suggested white - sorry.

Also, has it been decided to abandon the routine @dnschneid posted here: https://github.com/dnschneid/crouton/pull/1710#issuecomment-102679054 ?
I think it might be useful for things such as the 'WARNINGS: ...' when you might not want to colorize the entire message but perhaps just the 'WARNING:' prefix.

Just some ideas/suggestions, be glad to hear others.

divx118:
@DennisLfromGA Yep sorry for the confusion. The yellow was just a quick example. I indeed think magenta is a good color for those and only highlight the "WARNING:" to prevent the output becoming like a rainbow. 
To summarize coloring can be as follows:
 * Red for all exit 1 errors.
 * Yellow exit 2, 3, 99 will depend on the message
 * Blue usage messages and prompts. Also highlight messages in $TIPS that require further user action like installing the chromeos extension.
 * Magenta warning messages just the text "WARNING:"

DennisLfromGA:
@divx118,

I guess without the 'c' function mentioned in https://github.com/dnschneid/crouton/pull/1710#issuecomment-102679054, we could still do something like this -
(if I understand the 'echo_color' function correctly to add '-n'):

`echo "$(echo_color "y" "-n" "WARNING:")"`
`echo " $TARBALL is an old-style bootstrap or backup." 1>&2`

-or even this-

`echo -n "$(echo_color "y" "WARNING:")"`
`echo " $TARBALL is an old-style bootstrap or backup." 1>&2`

I think either approach would work...

divx118:
@DennisLfromGA Yes both will work, but don't forget the redirect behind it same goes for option 2 in your example.  
So option 1 would be:.
```
echo "$(echo_color "m" "-n" "WARNING:")" 1>&2
echo " $TARBALL is an old-style bootstrap or backup." 1>&2
```
You can also use one line:
```
echo "$(echo_color "m" "-n" "WARNING:") $TARBALL is an old-style bootstrap or backup." 1>&2
```

dnschneid:
We could also change echo_color to take up to two strings, with the second uncolored.

divx118:
> We could also change echo_color to take up to two strings, with the second uncolored.  

Yes, probably best so the syntax keeps a bit cleaner at least in most cases. I will prepare a pull request.

divx118:
Created a new pull request to support the second string uncolored. https://github.com/dnschneid/crouton/pull/2343

PR #1727, opened 2015-05-07
Change the license to 2015

tonyxue:
There was a pull request have done this but only for a single file and did not get updated and seems disappeared?
I'm picky so I just did this :P

I did this by using script to simply replace 2014 to 2015. Hopefully I did not break anything else.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-// Copyright (c) 2014 The crouton Authors. All rights reserved.
+// Copyright (c) 2015 The crouton Authors. All rights reserved.
 //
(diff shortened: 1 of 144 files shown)


dnschneid:
Thanks!  There's some 2013s lingering around too...

tonyxue:
Fixed

tonyxue:
@dnschneid Let's merge!! Half of 2015 will pass soon!!! Not very long this patch will work! :P

dnschneid:
Aww snap, you're right!  We're almost out of time!

PR #1753, opened 2015-05-16
Add color to the prompt asking for username

tonyxue:
Entering the username for UID 1000 is where the automated setup process end
normally. Some users may not see the prompt since it mixes with the large amount
of other output. Add a color here just in case they don't notice it.

Patched according to #1664


tonyxue:
https://github.com/dnschneid/crouton/issues/1770#issuecomment-103969410 proves we really need this one. See the last a few lines of that reply.

And I will wait for #1710 first.

firestar115:
I would like that :smile:

DennisLfromGA:
Me too but it's been awfully quiet on the home front of late, hope everyone and everyting is okay...

dnschneid:
Okay, #1710 is (finally) merged, so you can (finally) update this patch.

tonyxue:
Ok so sorry that it took me half a year to think of I still have this pull request here.:P It has been updated.

dnschneid:
Well, it took me half a year to notice a typo in the commit :)

PR #1754, opened 2015-05-16
Added a volume script with interface similar to brightness.

samueldr:
This script interfaces with same volume slider that Chrome OS uses.
It has been tested on a C720p, but not with HDMI output.

There is an issue currently with this script:
  * The status of the volume chrome-side is not updated. This means that lowering the volume in crouton, then switching to Chrome OS, then lowering it again will reset it at first to where it was the last time it was changed in Chrome OS.

This script is far from minimal, it uses some ugly command pipes to get status informations, but it works and should be solid.

It might be a good idea to build a native cras client that does this by itself instead of relying on the test utility and the dbus interface.

### DIFF
--- a/chroot-bin/volume
+++ b/chroot-bin/volume
@@ -3,8 +3,8 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 
-# This script uses amixer to present an interface similar to the
-# brightness script for crouton.
+# This script uses amixer to present an interface similar to the brightness
+# script for crouton.
 
 set -e
 set -u
@@ -15,6 +15,7 @@ APPLICATION="${0##*/}"
 USAGE="$APPLICATION [set] [0-100]
 $APPLICATION up|down [0-100]
 $APPLICATION mute [set|unset|toggle]
+$APPLICATION get [all|volume|mute]
 
 This script changes the volume of the current output
 device, as would changing it with the shortcut keys or
@@ -25,6 +26,7 @@ Shortcut invocations:
     $APPLICATION down   decreases by $DEFAULT_VOLUME_DELTA
     $APPLICATION mute   sets muted
     $APPLICATION 0-100  sets volume to given value
+    $APPLICATION get    shortcut for get all
 "
 # The control which controls the current output.
 ALSA_CONTROL="Master"
@@ -32,6 +34,18 @@ ALSA_CONTROL="Master"
 # Tiny sugar coating around amixer for common parameters.
 _amixer() {
     amixer -Dcras "$@"
+}
+amixer_get_value() {
+    # $1 is the control name to get the value for
+    _amixer cget "name=$1" | sed -n \
+        -e "/[[:space:]]\+:/s/.*=//p"
+    #       ^              ^      ^ Print the line.
+    #       ^              ^ Replace everything before the equal sign.
+    #       ^ Search for the line with spaces and a colon (the value).
+}
+
+get_volume() {
+    amixer_get_value "$ALSA_CONTROL Playback Volume"
 }
 
 set_volume() {
@@ -41,11 +55,7 @@ set_volume() {
 
 get_is_muted() {
     # Get the alsa state of the control.
-    local state="$(_amixer cget "name=$ALSA_CONTROL Playback Switch" | sed -n \
-        -e "/[[:space:]]\+:/s/.*=//p")"
-    #       ^              ^      ^ Print the line.
-    #       ^              ^ Replace everything before the equal sign.
-    #       ^ Search for the line with spaces and a colon (the value).
+    local state="$(amixer_get_value "$ALSA_CONTROL Playback Switch")"
 
     # Muted is off (control is off)
     if [ "$state" = "on" ]; then
@@ -54,6 +64,14 @@ get_is_muted() {
     else
         # Is muted, so is_muted is true
         return 0
+    fi
+}
+
+get_is_muted_as_text() {
+    if get_is_muted; then
+        echo "yes"
+    else
+        echo "no"
     fi
 }
 
@@ -152,6 +170,21 @@ set|[0-9]|[0-9][0-9]|100)
         exit 1
     fi
     set_volume "$amount"
+    ;;
+get)
+    action="${1-all}"
+    case "$action" in
+    mute) get_is_muted_as_text ;;
+    volume) get_volume ;;
+    all)
+        echo "Volume: $(get_volume)"
+        echo "Muted: $(get_is_muted_as_text)"
+        ;;
+    *)
+        error_help "Invalid action: $action for get."
+        exit 1
+        ;;
+    esac
     ;;
 *)
     error_help "Error: Unkown command $cmd"
--- a/chroot-bin/volume
+++ b/chroot-bin/volume
@@ -70,6 +70,13 @@ unmute() {
 
 relative_volume() {
     local delta="$1"
+
+    # Check that this is integer-ish enough.
+    if ! [ "$delta" -eq "$delta" ] 2>/dev/null; then
+        error_help "Error: $APPLICATION <up|down> needs a number [0-100]"
+        exit 1
+    fi
+
     if get_is_muted; then
         if [ "$delta" -lt 0 ]; then
             set_volume 0
@@ -78,10 +85,9 @@ relative_volume() {
     else
         # Is the volume going down or up?
         if [ "$delta" -lt 0 ]; then
-            # Down, then we need to make that number positive
-            : $(( delta=delta * -1 ))
+            # Going down, we strip the leading minus sign,
             # And suffix the minus sign.
-            set_volume "${delta}-"
+            set_volume "${delta#*-}-"
         else
             # Only suffix the plus sign.
             set_volume "${delta}+"
@@ -90,14 +96,16 @@ relative_volume() {
 }
 
 print_help() {
-    # Prints the usage message.
-    # Can also get a message to show first.
-    [ $# -gt 0 ] && echo "$@"
     echo "$USAGE"
+}
+error_help() {
+    # Prints an error message and the usage to stderr
+    [ $# -gt 0 ] && echo "$@" 1>&2
+    print_help 1>&2
 }
 
 if [ $# -lt 1 ]; then
-    print_help "Error: $APPLICATION needs at least a command." 1>&2
+    error_help "Error: $APPLICATION needs at least a command."
     exit 1
 fi
 
@@ -109,25 +117,19 @@ h*|-h*|--help)
     print_help
     ;;
 up)
-    relative_volume "${1-$DEFAULT_VOLUME_DELTA}"
+    relative_volume "${1:-$DEFAULT_VOLUME_DELTA}"
     ;;
 down)
-    relative_volume "-${1-$DEFAULT_VOLUME_DELTA}"
+    relative_volume "-${1:-$DEFAULT_VOLUME_DELTA}"
     ;;
 mute)
     action="${1-set}"
     case "$action" in
-    "toggle")
-        toggle_mute
-        ;;
-    set)
-        mute
-        ;;
-    unset)
-        unmute
-        ;;
+    toggle) toggle_mute ;;
+    set) mute ;;
+    unset) unmute ;;
     *)
-        print_help "Invalid action: $action for mute." 1>&2
+        error_help "Invalid action: $action for mute."
         exit 1
         ;;
     esac
@@ -136,7 +138,7 @@ set|[0-9]|[0-9][0-9]|100)
     amount="$cmd"
     # First, check that we received a value to set the volume to.
     if [ $# -eq 0 ] && [ "$cmd" = "set" ]; then
-        print_help "Error: $APPLICATION set needs an amount to set." 1>&2
+        error_help "Error: $APPLICATION set needs an amount to set."
         exit 1
     fi
     # The value /could/ have been a parameter or the command itself.
@@ -146,13 +148,13 @@ set|[0-9]|[0-9][0-9]|100)
     fi
     # Check that this is integer-ish enough.
     if ! [ "$amount" -eq "$amount" ] 2>/dev/null; then
-        print_help "Error: $APPLICATION set needs a number [0-100]" 1>&2
+        error_help "Error: $APPLICATION set needs a number [0-100]"
         exit 1
     fi
     set_volume "$amount"
     ;;
 *)
-    print_help "Error: Unkown command $cmd" 1>&2
+    error_help "Error: Unkown command $cmd"
     exit 1
     ;;
 esac
--- a/chroot-bin/volume
+++ b/chroot-bin/volume
@@ -1,0 +1,153 @@
+#!/bin/bash
+
+# This script uses cras_test_client to present an interface similar to the
+# brightness script for crouton.
+
+# The MIT License (MIT)
+# 
+# Copyright (c) 2015 Samuel Dionne-Riel
+# 
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+# 
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+# 
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+set -e
+set -u
+
+DEFAULT_VOLUME_DELTA=5
+
+get_current_output() {
+	cras_test_client --dump_server_info | grep -i 'Selected Output Node' | \
+		sed -e 's#[[:space:]]*##g' | sed -e 's#^[a-zA-Z]*:##'
+}
+
+get_current_output_infos() {
+	cras_test_client --dump_server_info | grep $(get_current_output) | grep '^	'
+}
+
+get_current_volume() {
+	get_current_output_infos | cut -d'	' -f3 | sed -e 's# *##' | cut -d' ' -f1
+}
+set_volume() {
+	local volume=$1
+	# Clamp volume to zero
+	# A negative amount of absolute volume sets it to 100%
+	if (( volume < 0 )); then
+		volume=0
+	fi
+	cras_test_client --set_node_volume $(get_current_output):$volume
+}
+
+get_is_muted() {
+	local counter=0
+	local is_muted=""
+	host-dbus dbus-send --print-reply --system \
+		--dest=org.chromium.cras --type=method_call \
+		/org/chromium/cras org.chromium.cras.Control.GetVolumeState \
+		| while read line ; do
+		# We want the fifth value.
+		if [[ "$counter" == "5" ]]; then
+			is_muted=$(echo $line | cut -d' ' -f2-)
+			echo $is_muted
+		fi
+		(( counter++ ))||:
+	done
+}
+
+toggle_mute() {
+	if [[ $(get_is_muted) == "true" ]]; then
+		unmute
+	else
+		mute
+	fi
+	exit 0
+}
+
+mute() {
+	cras_test_client --user_mute 1
+}
+unmute() {
+	cras_test_client --user_mute 0
+}
+
+relative_volume() {
+	local delta=$1
+	# FIXME : When muted, set to zero if relative is going down.
+	# We cannot currently get the proper mute state.
+	# FIXME : When unmuting, should only unmute and not adjust.
+	if [[ $(get_is_muted) == "true" ]]; then
+		unmute
+		if (( delta < 0 )); then
+			set_volume 0
+		fi
+	else
+		set_volume $(( $(get_current_volume) + $delta ))
+	fi
+}
+
+print_help() {
+	echo "Usage: $(basename $0) [set|up|down|mute|0-100]"
+	echo "   up/down can receive an amount value to increase/decrease."
+	echo "   mute can receive commands set|unset."
+	echo "   set take a value between 0 and 100."
+	echo ""
+	echo "Shortcut invocations:"
+	echo "  $(basename $0) up     increases by $DEFAULT_VOLUME_DELTA"
+	echo "  $(basename $0) down   decreases by $DEFAULT_VOLUME_DELTA"
+	echo "  $(basename $0) mute   sets muted"
+	echo "  $(basename $0) 0-100  sets volume to given value"
+}
+
+if [[ $# -lt 1 ]]; then
+	print_help
+	exit 0
+fi
+
+cmd=$1
+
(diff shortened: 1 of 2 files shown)


chroot-bin/volume
```diff
@@ -0,0 +1,153 @@
+#!/bin/bash
```
dnschneid: Could you make this /bin/sh compatible?

chroot-bin/volume
```diff
@@ -0,0 +1,153 @@
...
+
+# This script uses cras_test_client to present an interface similar to the
+# brightness script for crouton.
+
+# The MIT License (MIT)
```
dnschneid: This needs to be licensed like the rest of the project, and you'll need to sign the [CLA](https://cla.developers.google.com/clas/new?kind=KIND_INDIVIDUAL&domain=DOMAIN_GOOGLE) for the code to be accepted.

chroot-bin/volume
```diff
@@ -0,0 +1,153 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+set -e
+set -u
+
+DEFAULT_VOLUME_DELTA=5
+
+get_current_output() {
+	cras_test_client --dump_server_info | grep -i 'Selected Output Node' | \
```
dnschneid: You can combine the grep and the two seds all into one sed command.

chroot-bin/volume
```diff
@@ -0,0 +1,153 @@
...
+		sed -e 's#[[:space:]]*##g' | sed -e 's#^[a-zA-Z]*:##'
+}
+
+get_current_output_infos() {
+	cras_test_client --dump_server_info | grep $(get_current_output) | grep '^	'
```
dnschneid: You can combine the greps as well, and quote the `$(get_current_output)`

chroot-bin/volume
```diff
@@ -0,0 +1,153 @@
...
+}
+
+get_current_volume() {
+	get_current_output_infos | cut -d'	' -f3 | sed -e 's# *##' | cut -d' ' -f1
```
dnschneid: This seems really fragile.  Is there a better way to select the volume?

chroot-bin/volume
```diff
@@ -0,0 +1,153 @@
...
+}
+set_volume() {
+	local volume=$1
+	# Clamp volume to zero
+	# A negative amount of absolute volume sets it to 100%
+	if (( volume < 0 )); then
```
dnschneid: bashism

chroot-bin/volume
```diff
@@ -0,0 +1,153 @@
...
+
+get_current_output_infos() {
+	cras_test_client --dump_server_info | grep $(get_current_output) | grep '^	'
+}
+
+get_current_volume() {
+	get_current_output_infos | cut -d'	' -f3 | sed -e 's# *##' | cut -d' ' -f1
+}
+set_volume() {
+	local volume=$1
```
dnschneid: quoting

chroot-bin/volume
```diff
@@ -0,0 +1,153 @@
...
+	# Clamp volume to zero
+	# A negative amount of absolute volume sets it to 100%
+	if (( volume < 0 )); then
+		volume=0
+	fi
+	cras_test_client --set_node_volume $(get_current_output):$volume
```
dnschneid: quoting

chroot-bin/volume
```diff
@@ -0,0 +1,153 @@
...
+# THE SOFTWARE.
+
+set -e
+set -u
+
+DEFAULT_VOLUME_DELTA=5
+
+get_current_output() {
+	cras_test_client --dump_server_info | grep -i 'Selected Output Node' | \
+		sed -e 's#[[:space:]]*##g' | sed -e 's#^[a-zA-Z]*:##'
```
dnschneid: The rest of crouton uses 4-space indentation

chroot-bin/volume
```diff
@@ -0,0 +1,153 @@
...
+	cras_test_client --set_node_volume $(get_current_output):$volume
+}
+
+get_is_muted() {
+	local counter=0
+	local is_muted=""
+	host-dbus dbus-send --print-reply --system \
+		--dest=org.chromium.cras --type=method_call \
+		/org/chromium/cras org.chromium.cras.Control.GetVolumeState \
+		| while read line ; do
```
dnschneid: `read -r`

chroot-bin/volume
```diff
@@ -0,0 +1,153 @@
...
+		# We want the fifth value.
+		if [[ "$counter" == "5" ]]; then
```
dnschneid: bashism

chroot-bin/volume
```diff
@@ -0,0 +1,153 @@
...
+			is_muted=$(echo $line | cut -d' ' -f2-)
```
dnschneid: quoting (and the next line)

chroot-bin/volume
```diff
@@ -0,0 +1,153 @@
...
+			echo $is_muted
+		fi
+		(( counter++ ))||:
```
dnschneid: crazy bashism  :)

Also, you could probably do this entire block with a single sed or awk command.

chroot-bin/volume
```diff
@@ -0,0 +1,153 @@
...
+	done
+}
+
+toggle_mute() {
+	if [[ $(get_is_muted) == "true" ]]; then
```
dnschneid: bashism

chroot-bin/volume
```diff
@@ -0,0 +1,153 @@
...
+		unmute
+	else
+		mute
+	fi
+	exit 0
```
dnschneid: I believe the caller exits, so you probably don't want this here

chroot-bin/volume
```diff
@@ -0,0 +1,153 @@
...
+
+mute() {
+	cras_test_client --user_mute 1
+}
+unmute() {
+	cras_test_client --user_mute 0
+}
+
+relative_volume() {
+	local delta=$1
```
dnschneid: quoting

chroot-bin/volume
```diff
@@ -0,0 +1,153 @@
...
+	# FIXME : When muted, set to zero if relative is going down.
+	# We cannot currently get the proper mute state.
+	# FIXME : When unmuting, should only unmute and not adjust.
+	if [[ $(get_is_muted) == "true" ]]; then
```
dnschneid: quoting + bashism

chroot-bin/volume
```diff
@@ -0,0 +1,153 @@
...
+		unmute
+		if (( delta < 0 )); then
```
dnschneid: bashism

chroot-bin/volume
```diff
@@ -0,0 +1,153 @@
...
+			set_volume 0
+		fi
+	else
+		set_volume $(( $(get_current_volume) + $delta ))
```
dnschneid: quoting

chroot-bin/volume
```diff
@@ -0,0 +1,153 @@
...
+	fi
+}
+
+print_help() {
+	echo "Usage: $(basename $0) [set|up|down|mute|0-100]"
```
dnschneid: Could you please match the style of the rest of crouton for printing out usage text?

chroot-bin/volume
```diff
@@ -0,0 +1,153 @@
...
+	echo "   set take a value between 0 and 100."
+	echo ""
+	echo "Shortcut invocations:"
+	echo "  $(basename $0) up     increases by $DEFAULT_VOLUME_DELTA"
+	echo "  $(basename $0) down   decreases by $DEFAULT_VOLUME_DELTA"
+	echo "  $(basename $0) mute   sets muted"
+	echo "  $(basename $0) 0-100  sets volume to given value"
+}
+
+if [[ $# -lt 1 ]]; then
```
dnschneid: bashism

chroot-bin/volume
```diff
@@ -0,0 +1,153 @@
...
+	print_help
+	exit 0
+fi
+
+cmd=$1
+
+if [[ "$cmd" == "up" ]]; then
```
dnschneid: bashisms all the way down

chroot-bin/volume
```diff
@@ -0,0 +1,153 @@
...
+	relative_volume ${2-$DEFAULT_VOLUME_DELTA}
```
dnschneid: quoting to the finish

targets/audio
```diff
@@ -4,6 +4,7 @@
 # found in the LICENSE file.
 REQUIRES='core'
 DESCRIPTION='Support audio playback via Chromium OS'\''s audio system.'
+CHROOTBIN='volume'
```
dnschneid: excellent.  I always forget to do this :)

drinkcat:
Thanks for the PR.

I haven't tested recently, but isn't simpler to use `alsamixer` or `amixer` to change the volume? (that is, provided the audio target is installed in the chroot). The CRAS ALSA plugin should update the correct volume control. We suffer from the same synchronization problem with Chromium, though.

samueldr:
First of all, thank for the feedback, I'll go through the bashisms/code style issue today. I'll look into making the pipes shorter, but I'm no awk wizard, so expect the read|while to stay. For /bin/sh compatibility, is there anything else than /bin/dash that should be tested?

Secondly, about using alsamixer/amixer. I'm unable to get anything else than pulseaudio on my Debian jessie chroot with amixer and alsamixer. This is what I though should work, but it does not. Usually I can use alsamixer -c0 to target the "real" audio card on a full blown linux system, but somehow (I'm guessing because this is a chroot on a chromebook), the real (or cras virtual) devices will not show that way. It might be a separate issue to investigate, because it would be a much nicer, and especially, much more supported way to unify the volume control than a specialized script.

samueldr:
I was able to get by without using `awk`, and still, the `read` incantation is gone! This had a great payoff for me as it forced me to get more acquainted with `sed`.

I believe all the issues except the help text to be completely fixed.

The help text I did not really know what was wrong with it, but yet, I updated it as I found out it was a bit outdated. I tried getting help from what's in `/usr/local/bin/*`, but not many of them actually have help text, and they are not consistent either.

By the way the CLA has been agreed to. I did not think that this was a Google project.

chroot-bin/volume
```diff
@@ -0,0 +1,179 @@
...
+    cras_test_client --set_node_volume "$(get_current_output):$volume"
+}
+
+get_is_muted() {
+    local counter=0
+    local muted="$(host-dbus dbus-send --print-reply --system \
+        --dest=org.chromium.cras --type=method_call \
+        /org/chromium/cras org.chromium.cras.Control.GetVolumeState \
+        | sed -n -e '6s/.*boolean[[:space:]]\+//p')"
+    [ "$muted" = "true" ]
```
dnschneid: It's a little nicer to treat functions like functions, and explicitly return 1 or 0 depending on the outcome of that test.

chroot-bin/volume
```diff
@@ -0,0 +1,179 @@
...
+        unmute
+        if [ "$delta" -lt 0 ]; then
+            set_volume 0
+        fi
+    else
+        set_volume "$(( $(get_current_volume) + $delta ))"
+    fi
+}
+
+print_help() {
```
dnschneid: installer/main.sh has the crouton canonical way of printing usage.

chroot-bin/volume
```diff
@@ -0,0 +1,178 @@
+#!/bin/sh
```
dnschneid: See installer/main.sh for what the head of scripts should look like (especially the copyright text; no need to copy the entire license into each script).

Benefit of putting the USAGE text in the way the installer does it is that you don't need to repeat yourself with a header comment.

chroot-bin/volume
```diff
@@ -0,0 +1,178 @@
...
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+set -e
+set -u
```
dnschneid: By the way, I'm happy you did this.  I've been meaning to push crouton over to using -u as well, but never found enough time (see shu branch).

samueldr: Well, I actually do not want anything to do with anything that does not -u and -e! It's so much safer and much easier to see where you're wrong!

chroot-bin/volume
```diff
@@ -0,0 +1,178 @@
...
+
+if [ $# -lt 1 ]; then
+    print_help
+    exit 0
+fi
+
+cmd="$1"
+shift
+
+if [ "$cmd" = "help" ] || [ "$cmd" = "-h" ] || [ "$cmd" = "--help" ]; then
```
dnschneid: You may want to turn whole block into a case statement (against ${cmd#-}), so you can do nice things like `h*)` as a sloppy condition for help...

chroot-bin/volume
```diff
@@ -0,0 +1,178 @@
...
+    print_help
+elif [ "$cmd" = "up" ]; then
+    relative_volume "${2-$DEFAULT_VOLUME_DELTA}"
+elif [ "$cmd" = "down" ]; then
+    relative_volume "-${2-$DEFAULT_VOLUME_DELTA}"
+elif [ "$cmd" = "mute" ]; then
+    # Toggle cannot be used right now.
```
dnschneid: It can't?

chroot-bin/volume
```diff
@@ -0,0 +1,178 @@
...
+    action="${1-set}"
+    if [ "$action" = "toggle" ]; then
+        toggle_mute
+    elif [ "$action" = "set" ]; then
+        mute
+    elif [ "$action" = "unset" ]; then
+        unmute
+    else
+        echo "Invalid action: $action for mute."
```
dnschneid: 1>&2

chroot-bin/volume
```diff
@@ -0,0 +1,178 @@
...
+            set_volume 0
+        fi
+    else
+        set_volume "$(( $(get_current_volume) + delta ))"
+    fi
+}
+
+print_help() {
+    local script="$(basename "$0")"
+    echo "$script [set] [0-100]"
```
dnschneid: Also note: all of this should go to stderr...

chroot-bin/volume
```diff
@@ -0,0 +1,178 @@
...
+    fi
+    set_volume "$1"
+elif [ "$cmd" -eq "$cmd" ] 2>/dev/null; then
+    # Previous test tests for integer-ish value
+    set_volume "$cmd"
+else
+    print_help
+fi
+
+# vim: noai:ts=4:sw=4:expandtab
```
dnschneid: Please use something like YAIFA instead of adding vim configs to source files.

chroot-bin/volume
samueldr: You caught a dead comment! (Killing it)

dnschneid:
Have you tried `-c cras` for alsamixer?

Also, excellent commenting of the sed scripts.

samueldr:
It seems to be `-D cras` to access the CRAS device from alsamixer, which that now I know I can find some Google results about it.

Looks like I'll rebase the script around alsa, though I'll keep everything working as is, its external interface is stable as it is, except for the finer detail (volume level), you can make shortcuts behave exactly as they do in Chrome OS.

samueldr:
The script has been revamped to use `amixer`, which seems much, much less flimsier than using `cras_test_client`! The only issue I have is that getting the mute status is a bit flimsy, but is actually quite solid.

(The last three commits were written while trying to write this comment and lamenting on how `amixer` did not do 'X' and finally did.)

chroot-bin/volume
```diff
@@ -0,0 +1,158 @@
...
+    if get_is_muted; then
+        if [ "$delta" -lt 0 ]; then
+            set_volume 0
+        fi
+        unmute
+    else
+        # Is the volume going down or up?
+        if [ "$delta" -lt 0 ]; then
+            # Down, then we need to make that number positive
+            : $(( delta=delta * -1 ))
```
dnschneid: `delta="$((-delta))"`?

dnschneid: Or even `set_volume "${delta#*-}-"` :3

chroot-bin/volume
```diff
@@ -0,0 +1,158 @@
...
+            # Only suffix the plus sign.
+            set_volume "${delta}+"
+        fi
+    fi
+}
+
+print_help() {
+    # Prints the usage message.
+    # Can also get a message to show first.
+    [ $# -gt 0 ] && echo "$@"
```
dnschneid: print_help should always output to stderr, so you can pull the redirects into the function

chroot-bin/volume
```diff
@@ -0,0 +1,158 @@
...
+
+cmd="$1"
+shift
+
+case "$cmd" in
+h*|-h*|--help)
+    print_help
+    ;;
+up)
+    relative_volume "${1-$DEFAULT_VOLUME_DELTA}"
```
dnschneid: if $1 is set but empty, this will pass an empty string to relative_volume.  Is this what you intended?  Otherwise, you want `:-` instead of `-`

chroot-bin/volume
```diff
@@ -0,0 +1,158 @@
...
+    ;;
+down)
+    relative_volume "-${1-$DEFAULT_VOLUME_DELTA}"
+    ;;
+mute)
+    action="${1-set}"
+    case "$action" in
+    "toggle")
```
dnschneid: why the quotes?

chroot-bin/volume
```diff
@@ -0,0 +1,158 @@
...
+        toggle_mute
+        ;;
+    set)
```
dnschneid: feel free to compact these statements `set) mute;;`

dnschneid:
This is looking way nicer.

samueldr:
There is one thing left that may irritate you, when looking at what you asked: The usage message does go to stderr when it is a result of an error. Otherwise, when asking for help, the usage message goes to stdout. The rationale is that when asking for help, I expect that output to be in stdout. This tool does not have a big usage message, but for a heftier tool, I would expect to be able to pipe this to a pager or a tool to help manage the wall of text.

If for any reason this is completely against the policy of the crouton tools, I will make it always output to stderr, but as I explained, I believe that when asking for help, it should be what is outputted and not an error.

dnschneid:
That makes sense, and should probably propagate to the rest of crouton tools as well.

dnschneid:
Can you add "get" to get the current volume?

chroot-bin/volume
```diff
@@ -0,0 +1,160 @@
...
+    print_help
+    ;;
+up)
+    relative_volume "${1:-$DEFAULT_VOLUME_DELTA}"
+    ;;
+down)
+    relative_volume "-${1:-$DEFAULT_VOLUME_DELTA}"
+    ;;
+mute)
+    action="${1-set}"
```
dnschneid: Toggle would make a more reasonable default, wouldn't it?

dnschneid: Or are you trying to match the key behavior?

dnschneid: It matches key behavior very well, so I'm going to assume that and say it's fine :)

dnschneid:
Seems to behave properly.  I think adding a "get" command is all that's left before merging.

chroot-bin/volume
samueldr: Matches the key behaviour :smiley:

chroot-bin/volume
```diff
@@ -0,0 +1,160 @@
...
+set -u
+
+DEFAULT_VOLUME_DELTA=5
+
+APPLICATION="${0##*/}"
+USAGE="$APPLICATION [set] [0-100]
+$APPLICATION up|down [0-100]
+$APPLICATION mute [set|unset|toggle]
+
+This script changes the volume of the current output
```
dnschneid: Wrap to 80 characters

samueldr:
About the `get` command format, how about: `volume get` and `volume get [volume|mute]`, where get alone would get all the data at once, like

    $ volume get
    Volume: 42
    Muted: no

And the qualified `volume get [...]` outputting the data alone.

    $ volume get volume
    42
    $ volume get mute
    no

And, what is the most sensible output for the muted state? Alsa does it in an "inverted" way; it is not a *muted* state, but as *output active* state, which means on is unmuted and off is muted. I believe that asking for "mute" should be more like a *yes|no* question and not a quest for the meaning of *on* and *off*.

dnschneid:
That looks fine, since it's clear what the meaning is.
One option is using `=` instead of `: ` for the combined form so that you can eval the result if you really wanted to, but that's minor.

samueldr:
I believe that `eval`ing to get the values would be wrong in pretty much any cases. If you want the muted status, ask for mute, if you want the volume, ask for the volume. Also, I'd quite dislike using a tool that talks to me like I'm a machine. Using colons is succinct, but still feels human.

Not only that, but we're already wrapping the output of an app. I do not think that having another app wrapping a new level of indirection would be that great; if for any reason `amixer` outputs more than just the raw volume, the `eval`ed version would break for sure. In the other case, well, it depends what the caller does with the value.

In fact, I know I changed everything to alsa, but at the first hint of wonkyness with alsa, I think it might be better to go forward through dbus. I do not know, though, if the host dbus interfaces are stable enough through updates.

dnschneid:
OK, point taken :)

dnschneid:
Sorry, a bit slow at testing patches at the moment.  Did get an error, though:

```
$ sh volume get
amixer: Control cras element read error: Input/output error

Volume:
Muted: yes
$ amixer -Dcras cget name='Master Playback Volume'
numid=2,iface=MIXER,name='Master Playback Volume'
  ; type=INTEGER,access=rw------,values=1,min=0,max=100,step=0
amixer: Control cras element read error: Input/output error
```

amizer doesn't seem to return the right values regardless of the volume state.

samueldr:
This message worries me:

    amixer: Control cras element read error: Input/output error

I know next to nothing about the alsa/cras thing that's going on for crouton. In fact, I assume it's a crouton thing, but it might not be. (I know though that cras is a Chromium OS thing.)

The hardware I was able to test this on was a C720p, in Debian 8.0. `amixer` is version 1.0.28

What does `alsamixer -Dcras` do for you? I believe it will not be able to get (or maybe even set?) the values if amixer cannot. AFAIK, the message looks like alsa is able to find the right control, but cannot read from it.

[The actual place the error happens](http://git.alsa-project.org/?p=alsa-utils.git;a=blob;f=amixer/amixer.c;h=db1849333da3c09d8661dd0d4d7adb6c953230e5;hb=HEAD#l643), if it can be any help tracking down the issue.

dnschneid:
> Surely you tried updating your chroot, right?  [This](https://github.com/dnschneid/crouton/blob/master/CONTRIBUTING.md) says you should always update your chroot before claiming something's broken.

Oops, how embarrassing; you're right!  Updating my chroot fixed the issue with alsa.

It might be nice to wrap the amixer commands such that any error messages are consumed and instead, something generic like "update your chroot" is spat out instead.

samueldr:
I'll look into it.

I would implement an "amixer error handler", which would somehow read the error message. Then, would spit out an error message that would contain the original error message with a notice that most errors are due to out-of-date chroots, with a (shortened?) link to the update guide.

    The volume script hit an unexpected error:
        > [original amixer error message]
    Most of the time, this means the chroot is out of date.
    Update your chroot with this guide: [link].
    If the issue persists after an update, consider filing a bug report.

Though, this might be verbose. I don't expect to implement this before the week-end, if you have suggestion for a better error message, please do, I'm really not fond of this one.

(Maybe instead add an (hidden?) debug flag to spit out more infos like version of things used by this script to help debug issues).

dnschneid:
Keep it simple for now.  `2>/dev/null` and echo something and fail if the exit code is non-zero.

`Failed to communicate with the audio server. Please update your chroot.` is sufficient.

`croutonversion` will report the version of the chroot.  There's also an audio auto-update patch in the works to make things fix themselves (#1049)

samueldr:
It would be best to test in an out-of-date enough crouton chroot, but I tested by making it look for -Dfoobar and the sanity checks does seem to work properly.

dnschneid:
Thanks!

Issue #1558, opened 2015-03-29
xiwi keyboard target isn't working


mark-hahn:
I created a chroot with 

```
sudo sh -e ~/Downloads/crouton -t unity,xiwi,keyboard -r trusty -n mychroot
```

The chomebook keys (e.g. search-left for home) work fine in a command line when I enter the chroot like

```
sudo enter-chroot -n mychroot
```

However in either desktop unity or xfce I don't get those bindings at all.  For example the search key acts like Super_L and opens up the unity dash.

I've read the keyboard wiki and it says it should just work.  I haven't tried any xmodmap stuff.  xmodmap is new to me.

chromebook pixel 2015
Version 41.0.2272.102 (64-bit)
Platform 6680.78.0 (Official Build) stable-channel samus
Firmware Google_Samus.6300.102.0

**Edit:** I've been studying the keyboard on chromebook in general.  After reading other posts I was given the impression that the chromebook overlay definition is stored under `/usr/share/X11`.  But that directory doesn't exist on my chromebook.  Where are the definitions stored?

dnschneid:
Can you check the keyboard settings in Unity?  It should have "model: chromebook" set in there somewhere.

mark-hahn:
The unity gui has setting windows for`Keyboard Settings` and `Text Entry` (see screen caps below) but can't find anywhere to put in normal text-file config options.  Do you know how to change that setting?

I found the `/etc/X11/xkb` directory but it was empty.  I found the `/etc/default/keyboard` file  which contained

```
XKBMODEL="pc105"
XKBLAYOUT="us"
XKBVARIANT=""
XKBOPTIONS=""
```

I tried changing each of `XKBMODEL` and `XKBLAYOUT` to "chromebook" but neither change made any difference.

BTW, this problem is the same in unity and xfce so I doubt it is a unity setting.  Actually, my end goal is to have this work with the Atom editor in a xiwi window.  Atom works great now in a xiwi window but it has this same keyboard problem. 

I'm sorry to be such a pain.  I promise that after asking all these questions and I become an  expert (grin), I'll start helping others.


![screenshot 2015-03-30 at 5 02 07 pm](https://cloud.githubusercontent.com/assets/811455/6909767/85be3570-d6fe-11e4-9e93-5bc2a951adc3.png)

[...]

PR #1801, opened 2015-05-29
Fix for issue #1558 - Search key mappings in Freon

cribalik:
croutonfbserver now sends information to kiwi nacl module on startup regarding whether we are using Freon or not. The nacl module uses different key schemes depending on whether Freon is used or not, enabling SearchKey + otherkey mappings to work correctly for both freon and non-freon users.

### DIFF
--- a/host-ext/nacl_src/kiwi.cc
+++ b/host-ext/nacl_src/kiwi.cc
@@ -246,6 +246,12 @@ private:
         return false;
     }
 
+    /* Recieves and processes initialization information */
+    void SocketParseInitInformation(const char* data, int datalen) {
+        initinfo* info = (struct initinfo*) data;
+        freon_ = info->freon;
+    }
+
     /* Receives and handles a version request */
     bool SocketParseVersion(const char* data, int datalen) {
         if (connected_) {
@@ -421,6 +427,11 @@ private:
             return;
         }
 
+        if (data[0] == 'I') { /* Init information */
+            SocketParseInitInformation(data, datalen);
+            return;
+        }
+
         if (connected_) {
             switch (data[0]) {
             case 'S':  /* Screen */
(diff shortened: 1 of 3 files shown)


host-ext/nacl_src/kiwi.cc
```diff
@@ -246,6 +246,12 @@ class KiwiInstance : public pp::Instance {
         return false;
     }
 
+    /* Recieves and processes initialization information */
+    void SocketParseInitInformation(const char* data, int datalen) {
```
dnschneid: Is the idea that later revs would just make datalen longer, so you don't need to bother with versioning?

src/fbserver.c
```diff
@@ -491,6 +491,19 @@ int write_cursor() {
     return 0;
 }
 
+void write_init() {
+    struct stat _stat;
```
dnschneid: Is there a better way to name this variable, other than a leading _?

dnschneid:
This looks quite nice, thanks!  @drinkcat, any thoughts?

src/fbserver.c
cribalik: Well it's a buffer for the `stat` call.. so how about `stat_buf`?

host-ext/nacl_src/kiwi.cc
cribalik: I think that might present problems (adding a field to the `initinfo` struct doesn't necessarily increase `datalen`). 
I was thinking of using the server_version_ variable for versioning instead.

dnschneid:
To be clear, this is compatible with your other PR (#1832), right?

src/fbserver.c
```diff
@@ -491,6 +491,19 @@ int write_cursor() {
     return 0;
 }
 
+void write_init() {
+    struct stat stat_buf;
+    char raw[FRAMEMAXHEADERSIZE + sizeof(struct initinfo)];
+    struct initinfo* i = (struct initinfo*)(raw + FRAMEMAXHEADERSIZE);
+    i->type = 'I';
+    i->freon = 0;
+    if (stat("/sys/class/tty/tty0/active", &stat_buf) == -1) {
```
drinkcat: Using `access` is arguably a bit simpler...

drinkcat:
This looks awesome! I'm happy as it is, using `access` instead of `stat` makes `write_init` a tiny bit simpler, but, really, up to you.

cribalik:
Good call @drinkcat, changed it to `access`.

cribalik:
@dnschneid Yes, this PR is compatible and non-redundant with my other PR (#1832).

host-ext/nacl_src/kiwi.cc
```diff
@@ -257,7 +267,7 @@ class KiwiInstance : public pp::Instance {
 
         if (server_version_ != VERSION) {
             /* TODO: Remove VF1 compatiblity */
-            if (server_version_ == "VF1") {
+            if (server_version_ == "VF1" || server_version_ == "VF2") {
```
dnschneid: Actually, can we not pop up the warning for the update?  Most people probably won't care in this case...

dnschneid: Eh, I take that back.  Maybe it'll force people to update occasionally :)

PR #1808, opened 2015-05-30
fvwm-crystal

aurey2244:
I figured that this would excellent to have if you want an eye candy desktop that is also none power hungry such as unity and gnome.

 The things is you'd edit the ~/.xinitrc file and add "exec fvwm-crystal" into the file, which I use nano for.

I just don't know how to add this into the script if neccesary


dnschneid:
Please test your patches before submitting them.   There's no way the patch as submitted would actually work.

PR #1832, opened 2015-06-05
Input-related settings no longer blocked on Freon

cribalik:
Fix for issue #1558 where the `chromebook` keyboard model is not chosen at startup when using xiwi.

For non-Freon users using xiwi, no input settings (like the use of the chromebook keyboard model) are needed because the ChromeOS performs the keymapping before sending them to the window.

But for Freon users, ChromeOS does not perform this keymapping and instead sends the actual keypresses, again requiring the use of the input settings.

Changed the if statement so that it ignores setting input-related stuff only if using xiwi and not using freon.

### DIFF
--- a/chroot-bin/croutonxinitrc-wrapper
+++ b/chroot-bin/croutonxinitrc-wrapper
@@ -87,12 +87,17 @@ fi
 # Launch system-wide trigger daemon
 croutontriggerd &
 
-# Input-related stuff is not needed for kiwi
-if [ "$xmethodtype" != "xiwi" ]; then
+
+# Apply the Chromebook keyboard map. Not needed for xiwi with the exception of Freon.
+if [ "$xmethodtype" != 'xiwi' ] || [ ! -f "/sys/class/tty/tty0/active" ]; then
     # Apply the Chromebook keyboard map if installed.
     if [ -f '/usr/share/X11/xkb/compat/chromebook' ]; then
         setxkbmap -model chromebook
     fi
+fi
+
+# Input-related stuff is not needed for kiwi
+if [ "$xmethodtype" != "xiwi" ]; then
 
     # Launch X-server-local key binding daemon
     xbindkeys -fg /etc/crouton/xbindkeysrc.scm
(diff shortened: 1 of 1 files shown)


mark-hahn:
This did not work for me.  When using the new `croutonxinitrc-wrapper` and disabling all other fixes, the keyboard didn't work at all.  I.E. no output when typing.  This happened in xterm and the Atom editor.

cribalik:
A bunch of other stuff was done besides setting keyboard model to `chromebook`. Now the script is more reserved; it only set the keyboard model. We'll have to add more stuff if we realize it's causing an issue.

@mark-hahn, can you confirm that it works now  ?

mark-hahn:
Yes, it works.

chroot-bin/croutonxinitrc-wrapper
```diff
@@ -126,6 +126,15 @@ if [ "$xmethodtype" != "xiwi" ]; then
     fi
 fi
 
+# Apply the Chromebook keyboard map in xiwi if using Freon. 
+if [ "$xmethodtype" = 'xiwi' ] && [ ! -f "/sys/class/tty/tty0/active" ]; then
```
dnschneid: This should be an elif for the above block, and the comments above the previous block need to be tweaked.

dnschneid: Actually, the `setkbmap -model chromebook` in the above block could be factored out if you move this code up above.  Then the condition for setting the map is "xiwi or freon" (and the map exists).

dnschneid:
@drinkcat does this match your observations?

drinkcat:
Yes. That's the right thing to do. Thanks.

chroot-bin/croutonxinitrc-wrapper
cribalik: That would indeed be better. It's now factored out under the the condition `(!xiwi) || freon`.

dnschneid:
Fixed it.

Thanks a lot for these patches!!

cribalik:
My pleasure. Thank you for creating such an amazing application!

PR #1834, opened 2015-06-06
Force-yes when doing dist-upgrade

tonyxue:
If users installed some packages from PPA that cannot be authenticated, the
original ```apt-get -y dist-upgrade``` will stop working, returning an error: ```There
are problems and -y was used without --force-yes```.

I personally installed newer version of xfce4 from PPA and got this error.

### DIFF
--- a/targets/core
+++ b/targets/core
@@ -103,7 +103,7 @@ EOF
     apt-get -y update || true
 
     echo 'Ensuring system is up-to-date...' 1>&2
-    apt-get -y dist-upgrade
+    apt-get -y --force-yes dist-upgrade
 fi
 
 # On release upgrade, keyboard-configuration might be reconfigured.


dnschneid:
Is that really a good idea?  Package authentication is the only thing stopping you from getting MITM'd, since the package indexes are usually grabbed via http.

If you have a PPA that is reporting unauthenticated packages, you need to grab the PPA's key and add it to the keychain.

tonyxue:
@dnschneid You are right. That's a bad idea to directly ```---force-yes```.
How about I add one more check at ```apt-get update``` to ```grep``` any ```GPG Error``` (joining with ```ppa.launchpad.net``` ) then add the keys from ubuntu key server?

dnschneid:
I'd rather not blindly trust PPAs that are added to the sources.  If the user added the source, they should add the signing key as well.

PR #1903, opened 2015-06-28
Lookup release name correctly. Fixes #1880.

### LINKED ISSUE #1880: getrelease.sh fails if extra fields are in sources.list
installer/ubuntu/getrelease.sh prints $3 of the matching line in /etc/apt/sources.list. However, this isn't always the release name. For example, consider the following entry:

deb [arch=armhf] http://ports.ubuntu.com/ubuntu-ports/ trusty main restricted universe multiverse

$4 is required not $3.

(Serves me right for hacking the file in an attempt to run x86 binaries under qemu I suppose...)

tonyxue:
getrelease.sh now looks up release name after the URI of one of the types
specified in the [man page](http://manpages.debian.org/cgi-bin/man.cgi?sektion=5&query=sources.list&apropos=0&manpath=sid&locale=en).

Fixes #1880

### DIFF
--- a/installer/ubuntu/getrelease.sh
+++ b/installer/ubuntu/getrelease.sh
@@ -18,8 +18,11 @@ if [ ! -s "$sources" ]; then
     exit 1
 fi
 
-rel="`awk '/^deb .* main( .*)?$/ { print $3; exit }' \
-          "$sources" "$sources.d"/*.list 2>/dev/null`"
+rel="`awk '/^deb .* main( .*)?$/ \
+    { for ( i = 1; i <= NF; i++ ) \
+    { if ( $i~/http|file|cdrom|ftp|copy|rsh|ssh/ ) \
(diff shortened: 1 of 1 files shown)


tonyxue:
Noticed variable name choice right after committing.
But I think it doesn't matter if it is only effective in ```awk```

dnschneid:
Thanks, although I think this could be way simpler.  All mirror text will have :// in them, and I'm pretty sure that won't appear anywhere else in the line.  So something like this should work:

```sed -n 's|^deb .*://[^ ]* \([^ ]*\) main\( .*\)\?$|\1|p' "$sources" "$sources.d"/*.list | head -n 1```

tonyxue:
Oh that's a good idea. Updated.
I've been playing regex for only a while so I spent some time understand it and it works.

PR #1945, opened 2015-07-17
Respect armel for architecture flag

taterbase:


taterbase:
I needed to be able to create a chroot that used the `armel` architecture but the defaults file overwrites a `-a armel` as `armhf`. I was able to modify it locally and get it to work but I think this could be useful upstream.

dnschneid:
Good point.  I never imagined someone would actually want to do that :)

Issue #1954, opened 2015-07-21
Updated stretch chroot, can't enter xorg via starte17 anymore


yebyen:
```
$ sudo starte17
Entering /mnt/stateful_partition/crouton/chroots/stretch...
-su: 26: [[: not found

_XSERVTransmkdir: Owner of /tmp/.X11-unix should be set to root

X.Org X Server 1.17.2
Release Date: 2015-06-16
X Protocol Version 11, Revision 0
Build Operating System: Linux 3.16.0-4-armmp-lpae armv7l Debian
Current Operating System: Linux localhost 3.8.11 #1 SMP Mon Jul 13 03:38:36 PDT 2015 armv7l
Kernel command line: cros_secure console= console= loglevel=7 init=/sbin/init cros_secure oops=panic panic=-1 root=/dev/dm-0 rootwait ro dm_verity.error_behavior=3 dm_verity.max_bios=-1 dm_verity.dev_wait=1 dm="1 vroot none ro 1,0 2506752 verity payload=PARTUUID=5dd92286-c06c-814c-bca2-891543ddd628/PARTNROFF=1 hashtree=PARTUUID=5dd92286-c06c-814c-bca2-891543ddd628/PARTNROFF=1 hashstart=2506752 alg=sha1 root_hexdigest=444c5397278900fbeaf5fa57602a7ec01f9d2d4d salt=d07e2c6731681ccac9ccef5a6013ccbea7d07651f601396a6208bfb4c6eab3bd" noinitrd vt.global_cursor_default=0 kern_guid=5dd92286-c06c-814c-bca2-891543ddd628  
Build Date: 01 July 2015  05:24:31PM
xorg-server 2:1.17.2-1 (http://www.debian.org/support) 
Current version of pixman: 0.32.6
        Before reporting problems, check http://wiki.x.org
        to make sure that you have the latest version.
Markers: (--) probed, (**) from config file, (==) default setting,
        (++) from command line, (!!) notice, (II) informational,
        (WW) warning, (EE) error, (NI) not implemented, (??) unknown.
(++) Log file: "/tmp/Xorg.crouton.1.log", Time: Mon Jul 20 19:59:50 2015
(==) Using system config directory "/usr/share/X11/xorg.conf.d"
(EE) 
Fatal server error:
(EE) xf86OpenConsole: Cannot open /dev/tty0 (No such file or directory)
(EE) 
(EE) 
Please consult the The X.Org Foundation support 
         at http://wiki.x.org
[...]

PR #1973, opened 2015-07-28
freon: Intercept open64 calls; fixes Xorg on Debian stretch-armhf.

mkasick:
Intercept open64 calls the same as open.  Required on Debian stretch-armhf where Xorg calls open64 on /dev/tty0, /dev/tty7, and /dev/input/event.

### DIFF
--- a/src/freon.c
+++ b/src/freon.c
@@ -37,11 +37,13 @@ static int lockfd = -1;
 
 static int (*orig_ioctl)(int d, int request, void* data);
 static int (*orig_open)(const char *pathname, int flags, mode_t mode);
+static int (*orig_open64)(const char *pathname, int flags, mode_t mode);
 static int (*orig_close)(int fd);
 
 static void preload_init() {
     orig_ioctl = dlsym(RTLD_NEXT, "ioctl");
     orig_open = dlsym(RTLD_NEXT, "open");
+    orig_open64 = dlsym(RTLD_NEXT, "open64");
     orig_close = dlsym(RTLD_NEXT, "close");
 }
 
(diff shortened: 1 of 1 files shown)


src/freon.c
```diff
@@ -152,24 +154,44 @@ int ioctl(int fd, unsigned long int request, ...) {
     return ret;
 }
 
+static int _open(int (*origfunc)(const char *pathname, int flags, mode_t mode),
```
dnschneid: Why not just place the definition here?

src/freon.c
```diff
@@ -152,24 +154,44 @@ int ioctl(int fd, unsigned long int request, ...) {
...
+                 const char *origname, const char *pathname, int flags, mode_t mode);
+
 int open(const char *pathname, int flags, ...) {
     if (!orig_open) preload_init();
 
     va_list argp;
```
dnschneid: If you pass &flags to _open you can pull the shared vaargs stuff into the shared function as well

src/freon.c
```diff
@@ -152,24 +154,44 @@ int ioctl(int fd, unsigned long int request, ...) {
...
     va_start(argp, flags);
-    int mode = va_arg(argp, int);
+    mode_t mode = va_arg(argp, mode_t);
+    va_end(argp);
+
+    _open(orig_open, "open", pathname, flags, mode);
```
dnschneid: return?

dnschneid:
Nice catch!

src/freon.c
mkasick: I tried this, but two issues came up.  First, when trying to call va_start in _open, gcc complains:

> error: ‘va_start’ used in function with fixed args

So I had to modify _open's function signature to make it variadic, which is a bit misleading, but [here's the result](https://github.com/mkasick/crouton/commit/freon-test).

The second problem is that the va_start macro doesn't merely use the reference operator on the last argument, but instead calls __builtin_va_start passing the argument "directly".  Whatever built-in magic happens there doesn't obtain the right reference, so mode isn't fetched properly.

Thus, I'm not sure how to keep the va-args stuff in _open aside from passing a va_list, which also requires calling va_start in the separate functions.  Thoughts?

src/freon.c
mkasick: The intention was to preserve the forward reading of the original function across the split, which also happens to result in a smaller diff than placing the definition here.  That said, I'm happy with either way, so I've updated the pull request with the definition here (and the return fix).

src/freon.c
mkasick: Of course, should've tested with -Wall.

src/freon.c
dnschneid: Yeah, it's not worth banging heads with the compiler on this one :)

src/freon.c
dnschneid: `make croutonfreon.so` compiles the library with -Wall

yebyen:
This pull appears to solve my #1954 report.

dnschneid:
Thanks again!

PR #2061, opened 2015-09-02
Change the condition of applying vgem hack

arbuztw:
It seems that xserver will also check /sys/class/drm/card1/ for vgem existence,
so also apply the hack when the directory exists.

### DIFF
--- a/targets/x11-common
+++ b/targets/x11-common
@@ -12,7 +12,7 @@ ln -sfT '/etc/crouton/xserverrc' '/etc/X11/xinit/xserverrc'
 echo "$XMETHOD" > '/etc/crouton/xmethod'
 
 # Only apply hack if vgem card exists
-if [ -c '/dev/dri/card1' -o -d '/sys/class/drm/card1/' ]; then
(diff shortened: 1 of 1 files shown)


dnschneid:
Does this fix any issues?  On what platforms?

arbuztw:
It's an issue when we are running in containers. (PR #2063)
Since we only bind mount /dev/null, /dev/random, ... etc, it will not do the hack on Xorg.

dnschneid:
Makes sense.  Do we still need to check /dev then, or is it redundant?

arbuztw:
No, I think we don't need to check /dev.
From `strace`,  I get

    [pid 13153] readlink("/sys/class/drm/card1", "../../devices/platform/vgem/drm/"..., 1024) = 37
    [pid 13153] stat("/sys/devices/platform/vgem/drm/card1/uevent", {st_mode=S_IFREG|0644, st_size=4096, ...}) = 0

So I think it will always check vgem from /sys.

dnschneid:
Obsoleting this with #2097

PR #2063, opened 2015-09-03
Detect whether we are running in containers and act responsively.

arbuztw:
Detect whether we are running in containers by inspecting the environment
variable $container. If we are in container, we are not able to remount
or change the config in /proc/sys/*, so just skip them. We also have
to patch debootstrap since we are not able to mknod and we don't want
to start services in containers.

### DIFF
--- a/host-bin/enter-chroot
+++ b/host-bin/enter-chroot
@@ -47,7 +47,7 @@ chrootcmd() {
     # env may be overridden when running in the background; don't let it fork.
     local ret=0 oldtrap="$TRAP"
     TRAP=''
-    env -i container="$container" chroot "$CHROOT" su -s '/bin/sh' -c "$*" - root || ret=$?
+    env -i chroot "$CHROOT" su -s '/bin/sh' -c "$*" - root || ret=$?
     local pid="$!"
     # $pid might not be set if env has not been redefined yet
     if [ -n "$BACKGROUND" ] && [ -n "$pid" ]; then
@@ -323,7 +323,32 @@ bindmount() {
     fi
     mkdir -p "$target"
     mount --bind $bindopts "$1" "$target"
-    if [ -n "$3" ]; then
+    # Determine whether we really need to remount or not by checking the existing
+    # mount options. We are doing this since remount is not allowed in containers.
+    if [ -n "$3" ] && ! checkmountopt "$target" "$3"; then
+        mount -i -o "remount,$3" "$target"
+    fi
+}
+
+# Recursively bind-mounts $1 into $CHROOT/${2:-"$1"} if $2 is not already mounted
+# If $3 is specified, remounts with the specified options.
+# If $1 starts with a -, it's considered options to the bind mount, and the rest
+# of the parameters are shifted.
+rbindmount() {
+    bindopts=''
+    if [ ! "${1#"-"}" = "$1" ]; then
+        bindopts="$1"
+        shift
+    fi
+    local target="`fixabslinks "${2:-"$1"}"`"
+    if mountpoint -q "$target"; then
+        return 0
+    fi
+    mkdir -p "$target"
+    mount --rbind $bindopts "$1" "$target"
+    # Determine whether we really need to remount or not by checking the existing
+    # mount options. We are doing this since remount is not allowed in containers.
+    if [ -n "$3" ] && ! checkmountopt "$target" "$3"; then
         mount -i -o "remount,$3" "$target"
     fi
 }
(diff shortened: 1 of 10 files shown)


host-bin/unmount-chroot
```diff
@@ -267,6 +268,11 @@ unmount() {
             if [ -z "$printonly" ]; then
                 echo "Sending SIG$SIGNAL to processes under $CHROOTSRC..." 1>&2
             fi
+            # If there is no process running under $CHROOTSRC and we cannot
+            # unmount, then use lazy unmount.
+            # FIXME: This is a hack for running in containers, since we cannot
+            # unmount recursive bindmount in it.
+            arg='-l'
```
drinkcat: I believe this is a kernel bug. I'll try to reproduce on recent kernels and ask on LKML...

installer/functions
```diff
@@ -68,6 +68,41 @@ undotrap() {
...
+                    return 1
+                fi;;
+            *)
+                if ! awk -v dir="$1" -v opt="$opt(,|$)" \
+                      '$2 == dir && $4 !~ opt { exit 1 }' "/proc/mounts"; then
+                    return 1
+                fi;;
+         esac
+    done
+}
```
drinkcat: `return 0` for clarity

installer/ubuntu/bootstrap
```diff
@@ -31,7 +31,7 @@ Check your internet connection or proxy settings and try again.'
     fi
 fi
 
-# Patch debootstrap so that it retries downloading packages
+# Patch debootstrap so that it retries downloading packages and not mounting /sys
```
drinkcat: does not remount /sys

installer/functions
```diff
@@ -68,6 +68,41 @@ undotrap() {
     settrap "$TRAP"
 }
 
+# Check if we are able to mknod.
+capmknod() {
+    local tmp=`mktemp -d --tmpdir=/tmp 'crouton-mknod.XXX'`
+    if ! mknod "$tmp/test-mknod" c 0 0; then
+        return 1
+    fi
+    rm -rf "$tmp"
```
drinkcat: You leave the directory `$tmp` dangling if the test fails.

installer/functions
```diff
@@ -156,7 +193,9 @@ release() {
 # sure the kernel does not panic (this is the default configuration of a vanilla
 # kernel). See crbug.com/260955 for details.
 disablehungtask() {
-    echo 0 > /proc/sys/kernel/hung_task_panic
+    # Don't fail since we do not have permission to write /proc/sys/* if we are
+    # in container.
+    echo 0 > /proc/sys/kernel/hung_task_panic || true
```
drinkcat: 2>/dev/null

installer/functions
```diff
@@ -68,6 +68,43 @@ undotrap() {
     settrap "$TRAP"
 }
 
+# Check if we are able to mknod.
+capmknod() {
+    local tmp=`mktemp -d --tmpdir=/tmp 'crouton-mknod.XXX'`
+    local ret=0
+    if ! mknod "$tmp/test-mknod" c 0 0; then
```
drinkcat: 2>/dev/null

drinkcat:
`vboot_is_safe` is also noisy, because of `mount --bind / "$tmp"`

host-bin/enter-chroot
```diff
@@ -355,6 +390,10 @@ tmpfsmount /var/run/lock 'noexec,nosuid,nodev,size=5120k'
 bindmount /var/run/dbus /var/host/dbus
```
drinkcat: `2>/dev/null || true`. We probably won't bind mount /var/run/dbus in verified mode.

installer/functions
```diff
@@ -156,7 +193,9 @@ release() {
 # sure the kernel does not panic (this is the default configuration of a vanilla
 # kernel). See crbug.com/260955 for details.
 disablehungtask() {
-    echo 0 > /proc/sys/kernel/hung_task_panic
+    # Don't fail since we do not have permission to write /proc/sys/* if we are
+    # in container.
+    echo 0 2>/dev/null > /proc/sys/kernel/hung_task_panic || true
```
drinkcat: `echo 0 > /proc/sys/kernel/hung_task_panic 2>/dev/null` is a bit clearer.

arbuztw: I did this, but it doesn't work on my machine. It still printed out the error messages, and I'm not sure if it's the same on ChromeBook.
I guess what it did is something like this

    open("proc/sys/kernel/hung_task_panic")
    dup2()
    open("/dev/null")
    dup2()

and the error occurs (at first `open`) before we redirect the stderr.

drinkcat: Ah yeah that makes sense... Thanks.

Issue #1994, opened 2015-08-05
Keyboard and mouse get disabled during crouton install


TheStally:
Hi

I've recently power washed my chromebook and am looking to setup crouton again, which was previously running fine.

I downloaded the latest crouton and have been trying to install using the standard "-t xfce" to get ubuntu with xfce, however during the install process (around the time it's downloading/installing packages i think) my mouse and keyboard both stop working completely.

The install continues as far as entering the username, however by this point I cant do anything except hold the power button down and restart my chromebook.  Then the mouse/keyboard works fine.

I'm using an Acer Chromebook 11 (CB3-111), and have had this setup previously.

Anyone got any ideas?

tjcravey:
Same problem here.  Samsung XE303C12

I tried it again... on the second pass the mouse stopped working, but the keyboard worked, so I was able to complete the install.

Ephemerial:
I am experiencing a similar problem with a Toshiba Chromebook 2. Both the keyboard and touchpad stop responding to input part-way through the installation process. The issue is inconsistent and seems to occur roughly ~70% of the time, under the exact same circumstances.

tista500:
Greetings,

Same issue +1 on my Asus C200 with trusty fresh intall.
In my case during installation:
* internal keyboard - dead
* internal clickpad - works
* external USB keyboards - dead
* external Bluetooth keyboard - works

Regards.

danopia:
I posted about this on reddit: https://www.reddit.com/r/Crouton/comments/3i6axn/whenever_i_built_a_chroot_on_my_pixel_2015_it/

At one point running an install with a dead internal keyboard (from a previous chroot build) fixed the keyboard and broke the trackpad/touchscreen.

chrisware:
Same thing happened to me installing crouton/trusty/unity (which I've done several times on my Pixel 2015 LS in the last few days). Either the trackpad stopped and/or keyboard and/or touchscreen. The install gets to the last step but I can't complete the last step and enter the user name. Workaround was to power off the chromebook, restart then either start a fresh install or enter the chroot and when prompted to finish the install press Y. The freezing happens whether I'm using the chromebook during install, or not touching it.

sarbon15:
I am also experiencing this issue on my Acer C720 while trying to install trusty/unity. The first time through, the mouse was disabled, and then the keyboard. Then the second time, the keyboard and then the mouse.

mkvalor:
[...]

PR #2072, opened 2015-09-07
Deny auto start services.

arbuztw:
dpkg may start some services when installing, and we deny all of
them by return 101 in /usr/sbin/policy-rc.d. Users can still start
services manually by `start some-service` if they need it.

### DIFF
--- a/installer/ubuntu/prepare
+++ b/installer/ubuntu/prepare
@@ -230,6 +230,11 @@ if [ -r /debootstrap ]; then
     /debootstrap/debootstrap --second-stage
     # Clean contents of /var/run (since we'll always be mounting over it).
     rm -rf --one-file-system /var/run/*
+    # dpkg will auto start some services when installing, and it will ask
+    # policy-rc.d for permissions. We don't want any service to start, so always
+    # return denies.
+    echo exit 101 > /usr/sbin/policy-rc.d
+    chmod +x /usr/sbin/policy-rc.d
     # Request a script restart to refresh the mount environment
     relaunch_setup
 else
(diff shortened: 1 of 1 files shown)


dnschneid:
Does this fix any bugs in particular?

arbuztw:
I think this fixes issue #1994. I can reproduce that bug on samus many times, and it won't happen if I deny services to start. However, I am not quite sure, since that bug is not happening every time.

drinkcat:
And we'd have some autotests failing in the past (utopic IIRC?), because `dpkg-reconfigure` tries to restart (meaning, actually, start) services like udev, rsyslog, etc.

installer/ubuntu/prepare
```diff
@@ -230,6 +230,11 @@ if [ -r /debootstrap ]; then
     /debootstrap/debootstrap --second-stage
     # Clean contents of /var/run (since we'll always be mounting over it).
     rm -rf --one-file-system /var/run/*
+    # dpkg will auto start some services when installing, and it will ask
+    # policy-rc.d for permissions. We don't want any service to start, so always
+    # return denies.
+    echo exit 101 > /usr/sbin/policy-rc.d
+    chmod +x /usr/sbin/policy-rc.d
```
dnschneid: Why should this only apply to brand-new chroots?

arbuztw: Because this file only needs to be created once.

dnschneid: What if you're updating an old chroot?

arbuztw: Yes, I think that's right. It should be moved out of `if`.
(I can modified it, but I don't have a Chromebook to test it.)

dnschneid: Go ahead and modify it, and @drinkcat or I will test.

dnschneid:
Thanks again for these patches.

PR #2128, opened 2015-09-21
Don't add xephyr if system is using Freon

DennisLfromGA:
Checks if system employs Freon and removes xephyr target if installed.


DennisLfromGA:
Your definitely right, fixed I think...

PR #2130, opened 2015-09-22
targets/audio: Remove outdated hacks

drinkcat:
They have landed upstream a year ago. Also update minimum version to R39.


drinkcat:
Need to take care of that sbc patch as well.

Issue #2168, opened 2015-10-08
Add edit-chroot option to duplicate a chroot


ghost:
As a user I would like the ability to easily copy a chroot. Currently to achieve this I backup, then move, then restore from the backup.

I am not sure what option to use as `edit-chroot -c` and `edit-chroot -d` are already taken.

dnschneid:
If the chroot is unmounted, you can `sudo cp -a old_chroot new_chroot`.

You can also restore the backup using the crouton installer and specify `-n` to change the name.

But I guess we could add a command to edit-chroot as well.  Maybe `-C` or `-u`...

justinguy:
Great I will use `-u` as `-c` is already in use and submit a pull request.

tonyxue:
Maybe just write a simple function to ```cp``` and edit ```/etc/crouton/name```?

dnschneid:
/etc/crouton/name is updated on `enter-chroot`

Issue #2115, opened 2015-09-16
Add edit-chroot -u for updating chroots


dnschneid:
`edit-chroot -u` should download crouton (once) and updates the chroot(s).  Either explicitly list chroots or use -a to update everything.

DennisLfromGA:
pie???

dnschneid:
Possibly. Give it a shot and let me know :)

DennisLfromGA:
Will do.

DennisLfromGA:
I have a home-grown script that does someting similar but I'd have to incorporate it in to `edit-chroot`.

tedm:
how about an -s switch for "safety mode" - if -s is specified, copy the working chroot and crouton to ~/Downloads/backup/ and append the backup chroot name to chrootname_backup

then an -r restore mode, if the newly updated crouton and chroot give a blank screen or other odd behavior, -r reverts to a working system, then "echo, it almost worked, try next week..."

DennisLfromGA:
@tedm - `edit-chroot` already has a 'backup' and 'restore' option, why not use them?
```
    -b          Backs up the chroot to a tarball. Compression format is chosen
                based on the tarball extension. Backups always take place before
                other actions on a given chroot.
    -r          Restores a chroot from a tarball. The tarball path can be
                specified with -f or detected from name. If both are specified,
                restores to that name instead of the one in the tarball.
                Will not overwrite a chroot when restoring unless -r is
                specified twice.
```

PR #2170, opened 2015-10-08
Add copy option -u to edit-chroot

justinguy:
Adds the option to copy a chroot using `edit-chroot -u DEST`. See #2168

### DIFF
--- a/host-bin/edit-chroot
+++ b/host-bin/edit-chroot
@@ -58,12 +58,7 @@ Options:
                 If multiple chroots are specified, DEST must be a directory.
                 If you are moving a chroot to a SD card/USB drive, make sure the
                 storage is formatted to ext2/3/4.
-    -u DEST     Copies a chroot. Specify a new name to keep it in the same
-                directory, or an absolute path.
-                DEST can be a directory, in which case it must end in a slash.
-                If multiple chroots are specified, DEST must be a directory.
-                If you are copying a chroot to a SD card/USB drive, make sure the
-                storage is formatted to ext2/3/4.
+    -u DEST     Like -m, but copies instead of moves.
     -r          Restores a chroot from a tarball. The tarball path can be
                 specified with -f or detected from name. If both are specified,
                 restores to that name instead of the one in the tarball.
@@ -172,7 +167,7 @@ if [ $# -gt 1 -a -f "$KEYFILE" -a ! "$KEYFILE" = '-' ]; then
 elif [ $# -gt 1 -a -n "$MOVE" -a "${MOVE%/}" = "$MOVE" ]; then
     error 2 "Multiple chroots specified, but $MOVE is not a directory."
 elif [ $# -gt 1 -a -n "$COPY" -a "${COPY%/}" = "$COPY" ]; then
-      error 2 "Multiple chroots specified, but $COPY is not a directory."
+    error 2 "Multiple chroots specified, but $COPY is not a directory."
 elif [ $# -gt 1 -a -f "$TARBALL" ]; then
     error 2 "Multiple chroots specified, but $TARBALL is not a directory."
 fi
@@ -638,7 +633,7 @@ for NAME in "$@"; do
             error 1 "Unable to create directory for $target"
         fi
         echo "Copying $CHROOT to $target" 1>&2
-        cp -a "$CHROOT" "$target"
+        spinner 200 cp -av --one-file-system "$CHROOT" "$target"
     fi
 done
 
--- a/host-bin/edit-chroot
+++ b/host-bin/edit-chroot
@@ -58,7 +58,7 @@ Options:
                 If multiple chroots are specified, DEST must be a directory.
                 If you are moving a chroot to a SD card/USB drive, make sure the
                 storage is formatted to ext2/3/4.
-    -u DEST     Like -m, but copies instead of moves.
+    -C DEST     Like -m, but copies instead of moves.
     -r          Restores a chroot from a tarball. The tarball path can be
                 specified with -f or detected from name. If both are specified,
                 restores to that name instead of the one in the tarball.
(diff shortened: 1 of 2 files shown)


justinguy:
Just signed the Individual Contributor License Agreement per the contributor section of `README.md`.

host-bin/edit-chroot
```diff
@@ -57,6 +58,12 @@ Options:
                 If multiple chroots are specified, DEST must be a directory.
                 If you are moving a chroot to a SD card/USB drive, make sure the
                 storage is formatted to ext2/3/4.
+    -u DEST     Copies a chroot. Specify a new name to keep it in the same
```
dnschneid: `Like -m, but copies instead of moves.`

host-bin/edit-chroot
```diff
@@ -153,11 +166,13 @@ if [ -n "$BACKUP$RESTORE" -a -z "$TARBALL" -a "$PWD" = '/' \
...
 fi
 
-# If multiple chroots are listed, KEYFILE and MOVE must be empty or directories.
+# If multiple chroots are listed, KEYFILE and MOVE and COPY must be empty or directories.
 if [ $# -gt 1 -a -f "$KEYFILE" -a ! "$KEYFILE" = '-' ]; then
     error 2 "Multiple chroots specified, but $KEYFILE is not a directory."
 elif [ $# -gt 1 -a -n "$MOVE" -a "${MOVE%/}" = "$MOVE" ]; then
     error 2 "Multiple chroots specified, but $MOVE is not a directory."
+elif [ $# -gt 1 -a -n "$COPY" -a "${COPY%/}" = "$COPY" ]; then
+      error 2 "Multiple chroots specified, but $COPY is not a directory."
```
dnschneid: indentation

host-bin/edit-chroot
```diff
@@ -594,6 +615,31 @@ for NAME in "$@"; do
...
+        fi
+        if [ -e "$target" ]; then
+            # Can't tell if the destination is a directory or a chroot that
+            # already exists; be safe and assume it was a mistake.
+            error 2 "$target already exists"
+        fi
+        if ! mkdir -p "`dirname "$target"`"; then
+            error 1 "Unable to create directory for $target"
+        fi
+        echo "Copying $CHROOT to $target" 1>&2
```
dnschneid: Use the spinner approach that the cp portion of MOVE uses

justinguy:
Thanks for the feedback @dnschneid working on Sunday I see! Just pushed my updates. I also added the `--one-file-system` option because you had it in `-m`. I understand what the option does but curious as to what the reason was to use it. Cheers!

dnschneid:
Looking good.  `--one-file-system` is to protect the script from any weird bind mounts that people may be doing.  Don't leave home without it :)

Could you add tests to `test/tests/12-edit-chroot` to exercise the new feature?  Make sure to poke at the edge cases and error conditions.

justinguy:
I will work on the tests in the next couple days.

justinguy:
Added the tests. Traveling with my just my Chromebook so resorted to running the tests manually, let me know if any problems. Thanks!

dnschneid:
Oops, sorry, I just remembered that we want to add -u for updating chroots (#2115).  How about using -C (capital C)?

justinguy:
No problem, should have seen that  for -u. Changed to `-C` and added a test for trying to copy and move together.

justinguy:
Is there pointers somewhere for setting up a proper development environment so I can run the tests easier in the future? I usually don't work this far down the stack. Thanks!

dnschneid:
Thanks!  The tests are designed to be run straight from crosh on a Chromebook (not inside a chroot).  Copy the state of your repo to the chromebook (or use wget to grab the .tar.gz archive of your github repo), then use test/run.sh with some prefix of tests you want to run (such as 12 for 12-edit-chroot).

Other than that, it's up to you whether you want to develop in a chroot on the same machine you're testing on.

justinguy:
I figured that was the case. I am traveling and only carrying my Pixel. Combined with Crouton it is absolutely amazing. However I am reluctant to run tests (with things like filling the disk) because restoring everything from a powerwash takes forever. I guess if I want to contribute more I could just pick up a cheap Chromebook for testing.

Thanks for everything @dnschneid

dnschneid:
If you limit to just running one test (12 in this case), it won't run out of disk space, and the runner does a good job of cleaning up after itself on error.  Anyway, I ran the test:

```
515 [t] Expecting 'host edit-chroot -y -C /usr/local/test-20151028081933 prefix/12-edit-chroot.precise.1.Cwv/chroots.2 precise precise.2' to fail
515 [t] LAUNCHING: edit-chroot -y -C /usr/local/test-20151028081933 prefix/12-edit-chroot.precise.1.Cwv/chroots.2 precise precise.2
515 [e] Multiple chroots specified, but /usr/local/test-20151028081933 prefix/12-edit-chroot.precise.1.Cwv/chroots.2 is not a directory.
515 [t] FAIL with code 2: edit-chroot -y -C /usr/local/test-20151028081933 prefix/12-edit-chroot.precise.1.Cwv/chroots.2 precise precise.2
515 [t] PASS: 'host edit-chroot -y -C /usr/local/test-20151028081933 prefix/12-edit-chroot.precise.1.Cwv/chroots.2 precise precise.2' returned 2
515 [t] LAUNCHING: enter-chroot -n precise true
515 [e] /usr/local/test-20151028081933 prefix/12-edit-chroot.precise.1.Cwv/chroots/precise not found.
515 [t] FAIL with code 1: enter-chroot -n precise true
```

justinguy:
Went through the testing suite and got more comfortable with running them. Thanks for your patience. Just committed tests that are passing. This ended up being a deeper exercise than I expected, but it is awesome and I have learned a lot!

host-bin/edit-chroot
```diff
@@ -594,6 +610,31 @@ for NAME in "$@"; do
             mv "$CHROOT" "$target"
         fi
     fi
+
+    # Copy the chroot if requested
```
dnschneid: Actually, shouldn't this go above the MOVE section?  That way you can copy and move at the same time...

justinguy: I actually built it so you can't copy and move at the same time, and included testing for this. I can't see the use case of wanting to do both, and it seems it would be more likely to be a mistake. Do you feel this needs to be reworked so that you can move while copying?

dnschneid:
Finally merged.  Thanks!

Issue #1868, opened 2015-06-15
Kali Linux not downloading


movieboy711:
So I followed the instructions for trying to install Kali Linux and unfortunately I only get this:

Installing kali-amd64 chroot to /usr/local/chroots/kali
Downloading latest debootstrap...
2015-06-15 19:24:29 URL:http://anonscm.debian.org/gitweb/?p=d-i/debootstrap.git;a=snapshot;h=HEAD;sf=tgz [61841] -> "-" [1]
Patching debootstrap...
Downloading bootstrap files...
W: Cannot check Release signature; keyring file not available /usr/share/keyrings/debian-archive-keyring.gpg
I: Retrieving Release 
W: Retrying failed download of http://http.kali.org/dists/stable/Release
I: Retrieving Release 
W: Retrying failed download of http://http.kali.org/dists/stable/Release
I: Retrieving Release 
W: Retrying failed download of http://http.kali.org/dists/stable/Release
I: Retrieving Release 
W: Retrying failed download of http://http.kali.org/dists/stable/Release
I: Retrieving Release 
W: Retrying failed download of http://http.kali.org/dists/stable/Release
I: Retrieving Release 
W: Retrying failed download of http://http.kali.org/dists/stable/Release
I: Retrieving Release 
W: Retrying failed download of http://http.kali.org/dists/stable/Release
I: Retrieving Release 
W: Retrying failed download of http://http.kali.org/dists/stable/Release
I: Retrieving Release 
W: Retrying failed download of http://http.kali.org/dists/stable/Release
I: Retrieving Release 
W: Retrying failed download of http://http.kali.org/dists/stable/Release
W: http://http.kali.org/dists/stable/Release was corrupt
E: Failed getting release file http://http.kali.org/dists/stable/Release

tf7534:
I have the same problem bro i thinks it because the download link for kali got changed. If you open http.kali.org/dists/ there is no stable "option" the link in crouton needs to be changed to to http.kali.org/dists/kali i think

tonyxue:
Seems like the directory structure of http.kali.org has changed.

I've submitted a pull request to fix this. See #1870

dnschneid:
Should be fixed now.

Xxhackingxx:
i am having the same issue but it says 

Downloading latest crouton installer...
######################################################################## 100.0%
Installing kali-amd64 chroot to /usr/local/chroots/kali
Downloading latest debootstrap...
[...]

PR #2223, opened 2015-11-02
Kali-current is now sana

tonyxue:
Thus we only need sana now.

See https://github.com/dnschneid/crouton/issues/1868#issuecomment-152956650

### DIFF
--- a/installer/kali/releases
+++ b/installer/kali/releases
@@ -1,2 +1,1 @@
-kali
 sana
--- a/installer/kali/releases
+++ b/installer/kali/releases
@@ -1,1 +1,2 @@
+kali*
 sana


dnschneid:
This is a bit sticky, as there's no upgrade path.  Maybe just mark it as unsupported so that users get the appropriate "might be EOL" message?

tonyxue:
Updated. Like this?

dnschneid:
Yep, thanks.

cell13:
Oh,oh .So what can I do change the content ? The same problem.  " falied getting release file http://http.kali.org/kali/dists/sana.Release" .Thanks!

Issue #2260, opened 2015-11-24
wrong package name for kde install on kde wily/utopic


imakiro:
hello!
I've tried installing the kde-desktop target on Ubuntu Vivid release for my Dell CB 11 candy, 
the install fails with 

> Unable to locate package kde-plasma-desktop
> Failed to complete chroot setup.

The package name in wily is "plasma-desktop" not "kde-plasma-desktop"
I forked the project and tried to look for why the package name is wrong but I don't really get why because the process seems fully automated... 

Command run:

> sudo sh ~/Downloads/crouton -t xiwi,kde -r wily -e -P http://192.168.1.XXX:XXXX > ~/Downloads/installwilykde;log

(the  last 20 lines are coming, my log file doesn't seem to record what happened corectly and my bloody trackpad stops working each time the command crashes... mouse not recognised anymore too...)

natbett:
I was able to get this working. My solution was to extract the tar.bz file at the end of the /tmp/crouton-install-cache/crouton file:
```bash
cd /tmp/crouton-installer-cache
mkdir crouton.unbundled
line=`awk '/^###/ { print FNR+1; exit 0; }' crouton`
tail -n "+$line" crouton | tar xjC crouton.unbundled
```
Then I modified crouton.unbundled/targets/kde and replace kde-plasma-desktop with plasma-desktop. Then I built the tar.bz again:
```bash
cd crouton.unbundled && tar cvfj ../crouton.tar.bz *
```
and appended it in place of the original tar, I used vim to delete all lines after `### end of script; tarball follows` then ran:
```bash
cat crouton.tar.bz >> crouton
```

DennisLfromGA:
@natbett - Good job getting that working. :+1: 

----

@imakiro & @natbett,

Something similar to this cropped up when installing the vivid/gnome desktop, see https://github.com/dnschneid/crouton/issues/1706#issuecomment-104463165.

The problem is with the package specified in the desktop target files.
When it fails, you can enter the chroot as root and modify the `/prepare.sh` script to replace the correct package name and then re-run it. See the issue referenced above for more details.

For a more permanent fix, the '[kde target](https://github.com/dnschneid/crouton/blob/master/targets/kde)' would have to be modified via a pull request, similar to https://github.com/dnschneid/crouton/pull/1689 and using syntax similar to https://github.com/dnschneid/crouton/pull/1689#discussion_r29560691.

DennisLfromGA:
[...]

PR #2302, opened 2015-12-22
quickfix

imakiro:
trying to solve https://github.com/dnschneid/crouton/issues/2260 on the advices of @DennisLfromGA

### DIFF
--- a/targets/kde
+++ b/targets/kde
@@ -7,9 +7,13 @@ DESCRIPTION='Installs a minimal KDE desktop environment. (Approx. 600MB)'
 HOSTBIN='startkde'
 CHROOTBIN='crouton-noroot startkde'
 . "${TARGETSDIR:="$PWD"}/common"
-
+$TARGET=kde-plasma-desktop
+#quickfix for vivid package name 
+if [release -lt vivid]; then 
+    $TARGET=plasma-desktop 
+fi
 ### Append to prepare.sh:
-install --minimal kde-plasma-desktop -- network-manager
+install --minimal $TARGET -- network-manager
 
 TIPS="$TIPS
 You can start KDE via the startkde host command: sudo startkde


targets/kde
```diff
@@ -7,9 +7,13 @@ DESCRIPTION='Installs a minimal KDE desktop environment. (Approx. 600MB)'
 HOSTBIN='startkde'
 CHROOTBIN='crouton-noroot startkde'
 . "${TARGETSDIR:="$PWD"}/common"
-
+$TARGET=kde-plasma-desktop
```
divx118: This won't work, should be TARGET='kde-plasma-desktop' Also try to use quotes when declaring a string.

targets/kde
```diff
@@ -7,9 +7,13 @@ DESCRIPTION='Installs a minimal KDE desktop environment. (Approx. 600MB)'
...
+#quickfix for vivid package name 
+if [release -lt vivid]; then 
+    $TARGET=plasma-desktop
```
divx118: Idem see above

divx118:
Thanks for taking the effort to supply a patch. I am not sure however if we should want to use the global variable TARGET for this. Also IIRC it should be possible to add custom targets for releases when calling the install function.

imakiro:
haem, well first of all i don't use crouton anymore (took the chromebook and flashed it with a real kubuntu), and on the issue note you can see that I was completely lost on what was called where in the crouton script, so I thought that adding an condition here might be the solution.
On vivid, the package name is not 'kde-plasma-desktop' but only 'plasma-desktop' see here : http://packages.ubuntu.com/vivid/kde/plasma-desktop

So ok, I missed the quotes for the name, should I think I undestand that I also proposed the wrong place where to put the code? tell me and I will patch

targets/kde
```diff
@@ -7,9 +7,13 @@ DESCRIPTION='Installs a minimal KDE desktop environment. (Approx. 600MB)'
 HOSTBIN='startkde'
 CHROOTBIN='crouton-noroot startkde'
 . "${TARGETSDIR:="$PWD"}/common"
-
+$TARGET=kde-plasma-desktop
+#quickfix for vivid package name 
+if [release -lt vivid]; then
```
divx118: `if release -lt vivid; then`

divx118:
@imakiro No the place is good, indeed $TARGET is used and also capitols are used for global variables. Maybe use something like `installkdedesktop=´kde-plasma-desktop´` Also note when defining a string to a variable you mustn't use $ in front of the variable. Only when you want to expand the variable. I also noted another small error and that is the [] around the check. You should remove them. I added another comment to your code for that.

Just small things :)

Issue #2321, opened 2015-12-28
Crouton installer error when updating a chroot with x11 target


divx118:
Following error occurs when trying to update a trusty chroot with x11:
```
chronos@localhost / $ sudo sh ~/Downloads/crouton -u -n trusty -t x11
list of releases for  not found
Invalid target "x11".
```
This happens, because we check the targets if they are [valid](https://github.com/dnschneid/crouton/blob/master/installer/main.sh#L312). The release check in the x11 target will fail then, because the distro/release is not known at this moment. Possible solution is to wrap the check with -z TARGETNOINSTALL.

PR #2324, opened 2015-12-29
Prevent checking releases in x11 target when TARGETNOINSTALL is defined.

divx118:
This should fix https://github.com/dnschneid/crouton/issues/2321


dnschneid:
Thanks for the fix!

PR #2340, opened 2016-01-01
Adding xenial 16.04 next lts release.

divx118:
I tried it on a HP chromebook 14, with targets `xfce,extension,keyboard` it installs and starts just fine. Audio is working, clipboard sync not. I also tested unity. Normal unity target will fail to install, because of no unity-2d package anymore. unity-desktop will install. However on both targets, the unity desktop will fail to start. This will need some work.  
I think people can better start testing xenial, since it is going to be the next lts release and we do want to support that. Maybe remove test from vivid and wily?

### DIFF
--- a/installer/ubuntu/releases
+++ b/installer/ubuntu/releases
@@ -21,3 +21,4 @@ trusty
 utopic*
 vivid*|test
 wily*|test
+xenial*|test


idvoretskyi:
+

tista500:
I hope, too.

But on my some tests, unfortunately gnome-session 3.18.x failed to init with glib-critical error in startgnome (gnome-session-wrapper?) on my Flip. And agreed there's something to tweak.

But please do not drop Wily until the release day of XX beta1... :)

Regards.

dnschneid:
Sigh, maybe this time we'll be able to fully support a new release BEFORE it EOLs :)

DennisLfromGA:
A little off topic but could 'trusty' replace 'precise' as the default release yet?

divx118:
I correct my statement of clipboard sync not working. It does work, I created a new chroot with the same targets xfce,extension,keyboard so I could troubleshoot it, but it just works now.

PR #2347, opened 2016-01-04
Remove xephyr support

dnschneid:
Xephyr is not really needed anymore.  The only platforms that required xephyr (exynos and tegra) no longer do--exynos has been freonized, and tegra has the needed framebuffer fixes, so both work on xorg.

If tegra gets freonized in the future, we can remove all VT-related code as well, although that's not nearly so ugly.


chroot-bin/croutoncycle
```diff
@@ -101,42 +108,25 @@ if [ ! -f "/sys/class/tty/tty0/active" ]; then
...
+        if [ "${kiwidisp#:[0-9]}" != "$kiwidisp" ]; then
+            curdisp="$kiwidisp"
         fi
-    done
+    fi
 elif [ -z "$freonowner" ]; then
     # Poll the displays to figure out which one owns this VT
     curdisp="$tty"
-    for disp in $displist; do
+    for disp in ${displist#cros}; do
```
drinkcat: Can you separate the != cros test in the loop to make sure it does not bite us if we happen to change the order in displist?

chroot-bin/croutoncycle
```diff
@@ -101,42 +108,25 @@ if [ ! -f "/sys/class/tty/tty0/active" ]; then
...
-                kiwidisp="`cat $CRIATDISPLAY`"
-                if [ "${kiwidisp#:[0-9]}" != "$kiwidisp" ]; then
-                    curdisp="$kiwidisp"
-                fi
-            fi
-            break
+    # In Chromium OS or xiwi chroot
+    curdisp='cros'
+    if [ -n "$xiwiactive" -a -s "$CRIATDISPLAY" ]; then
+        kiwidisp="`cat $CRIATDISPLAY`"
```
drinkcat: "$(cat "$CRIATDISPLAY")"

drinkcat: huh, that's just re-indented code... Whatever, let's fix it ,-)

drinkcat:
So many negative lines, thanks!!

Minor nits only. And, for sure, needs testing...

chroot-bin/croutoncycle
dnschneid: Done!

jamesqo:
Awesome! Does this mean I'll finally be able to use Unity w/ Trusty on my ARM Chromebook?

jamesqo:
@drinkcat and @dnschneid, what is the progress on this? The reason I'm asking is because I'm looking to use Unity with 14.04 on my armhf Chromebook, but when I try running `crouton -t unity -r trusty` I get the error [here](https://github.com/dnschneid/crouton/blob/58b76a38ce4125373da8dc4023a9ab9570269482/targets/unity#L9) that Xephyr isn't supported.

PR #2371, opened 2016-01-13
Update targets/gnome

Linux-Distros:
Change gnome-session-fallback to gnome-session-flashback for Vivid And Up

### DIFF
--- a/targets/gnome
+++ b/targets/gnome
@@ -10,7 +10,7 @@ CHROOTBIN='crouton-noroot startgnome gnome-session-wrapper'
 
 ### Append to prepare.sh:
 install --minimal evolution-data-server gnome-control-center \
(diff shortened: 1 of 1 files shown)


Linux-Distros:
Yes But It Has Been Many Months Since, An Ideas When Its Going To Be Fixed?

DennisLfromGA:
Yes, When The Original PR Author Fixes The Syntax Glitches And Submits Them.

Simply Submitting A Copy Of The Same PR With The Same Glitches Isn't Going To Help - *At ALL*.

Linux-Distros:
Sorry

DennisLfromGA:
No problem. 
Adding a duplicate PR just adds to the developers already busy workload, that's all.

PR #2396, opened 2016-01-25
ubuntu/bootstrap: Patch debootstrap to not create devices under /dev

drinkcat:
More proper fix...

### DIFF
--- a/installer/ubuntu/bootstrap
+++ b/installer/ubuntu/bootstrap
@@ -23,7 +23,7 @@ if ! wget -O- --no-verbose --timeout=60 -t2 "$d"  \
 Check your internet connection or proxy settings and try again.'
     fi
     v="${f#*_}"
-    v="${v%.tar.xz}"
+    v="${v%.tar.gz}"
     echo "Downloading debootstrap version $v..." 1>&2
     if ! wget -O- --no-verbose --timeout=60 -t2 "$d$f" \
             | tar -C "$tmp" --strip-components=1 -zx 2>/dev/null; then
@@ -52,7 +52,7 @@ else
 fi
 
 # Patch debootstrap so that is does not create devices under /dev (issue #2387).
-sed -i -e 's/\(setup_devices () {\)/\1 return 0/' "$tmp/functions"
+sed -i -e 's/^setup_devices () {$/\0 return 0/' "$tmp/functions"
 
 # Add the necessary debootstrap executables
 newpath="$PATH:$tmp"
--- a/installer/ubuntu/bootstrap
+++ b/installer/ubuntu/bootstrap
@@ -51,6 +51,9 @@ else
     echo "Unable to patch debootstrap, moving on..." 1>&2
 fi
 
+# Patch debootstrap so that is does not create devices under /dev (issue #2387).
+sed -i -e 's/\(setup_devices () {\)/\1 return 0/' "$tmp/functions"
+
 # Add the necessary debootstrap executables
 newpath="$PATH:$tmp"
 cp "$INSTALLERDIR/$DISTRO/ar" "$INSTALLERDIR/$DISTRO/pkgdetails" "$tmp/"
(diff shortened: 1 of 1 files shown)


installer/ubuntu/bootstrap
```diff
@@ -54,6 +51,9 @@ else
     echo "Unable to patch debootstrap, moving on..." 1>&2
 fi
 
+# Patch debootstrap so that is does not create devices under /dev (issue #2387).
+sed -i -e 's/\(setup_devices () {\)/\1 return 0/' "$tmp/functions"
```
dnschneid: Drop the parens and use `\0` instead?

installer/ubuntu/bootstrap
```diff
@@ -9,24 +9,21 @@
...
+    f="`wget -O- --no-verbose --timeout=60 -t2 "$d" \
+            | sed -ne 's ^.*\(debootstrap_[0-9.]*.tar.gz\).*$ \1 p' \
+            | tail -n 1`"
+    if [ -z "$f" ]; then
+        error 1 'Failed to download debootstrap.
+Check your internet connection or proxy settings and try again.'
+    fi
     v="${f#*_}"
-    v="${v%.tar.gz}"
+    v="${v%.tar.xz}"
```
dnschneid: Should this actually be reverted?  This looked like a bug fix to me (a la #1926)

dnschneid:
Maybe the code to create the empty tarball should also be dropped.

installer/ubuntu/bootstrap
drinkcat: I did that because I could not find a reference to `\0` in the man page, nor POSIX. But, yeah, most probably ok ,-)

drinkcat:
> Maybe the code to create the empty tarball should also be dropped.

I'll double-check but I think it made deboostrap fail, even with the hack.

installer/ubuntu/bootstrap
drinkcat: Nope, that's right.

drinkcat:
> I'll double-check but I think it made deboostrap fail, even with the patch.

Yes, this (`debootstrap`), at the end of first stage:

```
	if ! am_doing_phase second_stage; then
...
		cp $DEBOOTSTRAP_DIR/devices.tar.gz	 "$TARGET/debootstrap/devices.tar.gz"
...
	fi
```

Now, I mean an empty file would also do the trick...

installer/ubuntu/bootstrap
drinkcat: Actually, the regex is missing checks for line beginning & end, so it could match another function.

drinkcat:
Alright, fixes in fixup commit.

rfc1036:
I am the author of the debootstrap change that apparently caused this regression. Do I understand correctly that debootstrap is being run in an environment which does not allow creating device nodes at all?

drinkcat:
> I am the author of the debootstrap change that apparently caused this regression.

Nice, thanks for stopping by ,-)

> Do I understand correctly that debootstrap is being run in an environment which does not allow creating device nodes at all?

Actually, we first call deboostrap first stage, then we bind mount `/dev` from outside the chroot (among other things), chroot, then call `deboostrap --second-stage`.

I did try unmounting /dev before running second-stage, but ended up with a "normal" /dev/null file with some content, as if some parts of deboostrap were expecting a valid /dev/null, then attempted to recreate it. I did not dig further into the issue...

At a minimum, it would be good to have an environment variable that could bypass the devices creation.

glaubitz:
The two-staged debootstrap process is also now broken in Debian unstable meaning it's no longer possible to debootstrap a chroot for a foreign architecture.

Previously, I could debootstrap --foreign onto a blank drive, hook that drive up to a single-board computer and let the machine boot with init=/bin/bash, then run ./debootstrap/debootstrap --second-stage.

With 1.0.75, second-stage worked out-of-the-box. With .76 and .77, debootstrap exits with error 1.

drinkcat:
@glaubitz : You should probably file a Debian bug about this...

glaubitz:
I will once I have verified the exact problem. I just noticed that downgrading debootstrap fixes my issue.

Issue #1706, opened 2015-05-03
Ubuntu 15.04 (Vivid) fails to install


justonian:
Hi there-

I realize I'm an early adopter of vivid, but I've tried twice with different parameters and can't get 15.04 with unity to install via crouton. Detailed notes regarding my requested install and error output is as follows. Appreciate any guidance on how to unblock the setup.

My command to invoke the chroot setup: sudo sh -e ~/Downloads/crouton -r vivid -t unity,keyboard,touch,xiwi -n vivid

Setup progresses as expected during retrieve/validate/extracting phases (outside of a keyring file not available warning presented a few times) until it gets to unpacking and configuring xkb-data where it mentions the python3 package may be at fault. Message there presented below:

I: Unpacking xkb-data...
W: Failure while installing base packages.  This will be re-attempted up to five times.
W: See //debootstrap/debootstrap.log for details (possibly the package /var/cache/apt/archives/python3_3.4.3-1_amd64.deb is at fault)

Next, it continues on and eventually fails again at ureadahead. Message show below:
I: Configuring ureadahead...
W: Failure while configuring base packages.  This will be re-attempted up to five times.
W: See //debootstrap/debootstrap.log for details (possibly the package /var/cache/apt/archives/python3_3.4.3-1_amd64.deb is at fault)

Finally, all GIDs get updated successfully then the final critical failure is relating to lsb-release, which ultimately fails installation and fails the chroot setup overall.

dpkg: dependency problems prevent configuration of lsb-release:
 lsb-release depends on python3; however:
  Package python3 is not installed.
 lsb-release depends on python3:any (>= 3.3.2-2~); however:

dpkg: error processing package lsb-release (--configure):
 dependency problems - leaving unconfigured
dpkg: dependency problems prevent configuration of ubuntu-minimal:
 ubuntu-minimal depends on lsb-release; however:
  Package lsb-release is not configured yet.
 ubuntu-minimal depends on python3; however:
  Package python3 is not installed.
 ubuntu-minimal depends on resolvconf; however:
  Package resolvconf is not installed.

dpkg: error processing package ubuntu-minimal (--configure):
 dependency problems - leaving unconfigured
dpkg: dependency problems prevent configuration of dh-python:
 dh-python depends on python3:any (>= 3.3.2-2~).

dpkg: error processing package dh-python (--configure):
 dependency problems - leaving unconfigured
Errors were encountered while processing:
 lsb-release
 ubuntu-minimal
 dh-python
[...]

PR #2429, opened 2016-02-09
Fix package management error during xenial, wily setup

StrawnPoint04:
"Error: Package gnome-session-fallback has no installation candidate" ― then a note that it has been replaced with gnome-session-flashback, so proposing a change to target script to reflect this

### DIFF
--- a/targets/gnome
+++ b/targets/gnome
@@ -10,7 +10,7 @@ CHROOTBIN='crouton-noroot startgnome gnome-session-wrapper'
 
 ### Append to prepare.sh:
 install --minimal evolution-data-server gnome-control-center \
-        gnome-screensaver gnome-session gnome-session-fallback \
+        gnome-screensaver gnome-session gnome-session-flashback \
         gnome-shell gnome-themes-standard gvfs-backends nautilus \
         unzip gtk2-engines-pixbuf
 
--- a/targets/gnome
+++ b/targets/gnome
@@ -8,9 +8,15 @@ HOSTBIN='startgnome'
 CHROOTBIN='crouton-noroot startgnome gnome-session-wrapper'
 . "${TARGETSDIR:="$PWD"}/common"
 
+if release -ge wheezy -ge kali -ge trusty; then
+  legacy_session_package="gnome-session-flashback"
+else
+  legacy_session_package="gnome-session-fallback"
+fi
+
 ### Append to prepare.sh:
 install --minimal evolution-data-server gnome-control-center \
-        gnome-screensaver gnome-session gnome-session-flashback \
+        gnome-screensaver gnome-session $legacy_session_package \
         gnome-shell gnome-themes-standard gvfs-backends nautilus \
         unzip gtk2-engines-pixbuf
 


Voronsky:
Greetings @StrawnPoint04 , seems we both have tried the same thing, a delayed response on my part to a referenced ticket, but I think you should give #1689 a look. Unfortunately gnome-session-flashback does not exist prior to trusty, wheezy or kali, so any distros that are mentioned on the crouton list, would break. In there I have one possible work-around pending.

StrawnPoint04:
Committed new code to the pull request to solve the Catch-22 that @Voronsky brought up

StrawnPoint04:
Since the only way I could possibly keep my fork up-to-date was to delete and recreate it, code to merge is [here](https://github.com/StrawnPoint04/crouton-fixes/commit/6d94044eec7899711cc068c6f6da89ef72c9977d).

drinkcat:
Just to clarify, this is the exact same fix as what was intended in #1689, right?

This one LGTM, I'm not sure what happened to the other (github interface says there are a lot of modified files)...

Voronsky:
@drinkcat seems the solution here is the same I have pending in #1689, both are showing what should be added to fix the issue.

targets/gnome
```diff
@@ -8,9 +8,15 @@ HOSTBIN='startgnome'
 CHROOTBIN='crouton-noroot startgnome gnome-session-wrapper'
 . "${TARGETSDIR:="$PWD"}/common"
 
+if release -ge wheezy -ge kali -ge trusty; then
```
dnschneid: I'm assuming all modern distros will be using -flashback, correct?  In that case, it might be better to make the if statement check `release -lt trusty`, since then wheezy+, kali+, and anything else we add will work without modification.

divx118: I just did a quick search and it seems wheezy and kali is using -fallback so it should be `release -lt jessie -lt sana -lt trusty` or `release -ge jessie -ge sana -ge trusty`

nahtnam:
I think this is the only issue that's stopping someone from using gnome on `xenial`. I tried using the new code in a custom target and it installed successfully but wasn't able to actually run it because it kept saying `No chroots with target 'gnome'`. If this is merged, I think it will work!

Thanks.

SoccerPie:
Can someone explain step by step how to change the fallback to flashback so I can install xenial gnome, I changed it on the zip file but I don't know where to go from there. @Voronsky 's post and this one don't make sense to me. Please and thank you guys!

StrawnPoint04:
No one has merged this pull request yet ― why? Bug still is causing problems.

DennisLfromGA:
@SoccerPie,

The idea is to replace gnome-session-fallback with gnome-session-flashback for newer releases.
However not all releases use it so conditional tests have to be done to apply it, the pull requests haven't been reviewed and merged yet.

However, you can run the crouton install script and when it fails, edit a script to fix it.
Take a look at https://github.com/dnschneid/crouton/issues/1706#issuecomment-104463165 for more details, it's using the 'vivid' release but the same applies to the 'xenial' release.

Once you've edited the script, just re-enter the chroot using `enter-chroot` and let the setup script finish.

*Hope this helps,*
-**DennisL**

dnschneid:
@StrawnPoint04 there's a [pending review item](https://github.com/dnschneid/crouton/pull/2429/files#r57680401) for you (not sure why it didn't show up in the log here).

targets/gnome
smibarber: +1 for `release -lt jessie -lt sana -lt trusty` and swap flashback/fallback below.

dnschneid:
Tracking in #2818

Issue #1321, opened 2015-01-07
Extension forgets HiDPI setting


dnschneid:
The extension forgets the HiDPI setting.  Which makes sense, because we don't actually save it anywhere.  We should change that.

PR #2462, opened 2016-02-23
Allow `enabled_` boolean to persist in extension

brycethorup:
I sometimes have two accounts logged in to Chrome OS at the same time. They both have the crouton extension installed. When I start crouton using xiwi it connects to whichever account happens to catch it first. I can avoid this by disabling one or the other account's crouton extension. The problem is that the option doesn't persist across logins. I have added the `storage` permission to the manifest to allow for the `enabled_` boolean to persist, and also calls to `chrome.storage.local.set` and `chrome.storage.local.get` to set the option when changed and to get the option when the extension is loaded.

### DIFF
--- a/host-ext/crouton/background.js
+++ b/host-ext/crouton/background.js
@@ -51,8 +51,9 @@ var notifications_ = {}; /* Map of notification id to function to be called when
 
 /* Check local storage for enabled option and set,
  * Otherwise default to true */
-chrome.storage.local.get("enabled", function(items){
+chrome.storage.local.get(null, function(items){
     enabled_ = (typeof items.enabled == "boolean" ? items.enabled : true);
+    hidpi_ = (typeof items.hidpi == "boolean" ? items.hidpi : false);
     refreshUI();
 });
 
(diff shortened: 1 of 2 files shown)


dnschneid:
Thanks for the PR!  (Ref: #1321)

Why not use localStorage?  This seems like something that we wouldn't want to sync between devices.

Could you also add the HiDPI setting while you're at it?

brycethorup:
I guess I chose `chrome.storage` because it is what I am familiar with. As long as the calls are made using the `chrome.storage.local` namespace instead of the `chrome.storage.sync` it won't sync to other devices. The option will only stay persistent on that device, for that user.

Yeah, I can add the HiDPI setting as well. I'll get it added and post back here when I am done.

brycethorup:
Ok, that should do it. Let me know if I messed anything up.

host-ext/crouton/background.js
```diff
@@ -49,6 +49,13 @@ var focus_win_ = -1; /* Focused kiwi window. -1 if no kiwi window focused. */
 var notifications_ = {}; /* Map of notification id to function to be called when
                             the notification is clicked. */
 
+/* Check local storage for stored options */
+chrome.storage.local.get(null, function(items){
+    enabled_ = (typeof items.enabled == "boolean" ? items.enabled : true);
+    hidpi_ = (typeof items.hidpi == "boolean" ? items.hidpi : false);
```
drinkcat: These duplicate the defaults above...

Can you do this instead?
```
if (typeof items.enabled == "boolean") enabled_ = items.enabled;
```

drinkcat:
One nit, otherwise LGTM, thanks!

brycethorup:
Thanks for the recommendation. That should do it.

dnschneid:
Thanks for this, and sorry for the delay!

PR #2504, opened 2016-03-28
add support for kali linux rolling releases

azet:
Kali recently switched to a rolling release model: https://www.kali.org/news/kali-linux-rolling-edition-2016-1/

(CLA signed, added myself to `CONTRIBUTORS`)

### DIFF
--- a/installer/kali/releases
+++ b/installer/kali/releases
@@ -1,2 +1,3 @@
 kali*
 sana
+kali-rolling
--- a/installer/kali/releases
+++ b/installer/kali/releases
@@ -1,3 +1,3 @@
 kali*
-sana
+sana*
 kali-rolling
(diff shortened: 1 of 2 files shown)


installer/kali/releases
```diff
@@ -1,2 +1,3 @@
 kali*
 sana
```
divx118: I think you should also make sana unsupported like kali already is. 

> Please note that the Kali sana repositories will no longer be updated and will be EOL’d on the 15th of April 2016.

azet: I wasn't sure about that as it's still supported. I will add a commit disabling sana.

dnschneid:
Thanks!

tonyxue:
Shouldn't the ```CONTRIBUTORS``` be alphabetical?

dnschneid:
It's actually auto-generated by the Makefile every time I go to release, so I didn't bother nitting it here :)

PR #2516, opened 2016-04-05
if netsurf-gtk not available, continue without it

zxvv:
Continue the installation without it if nesurf-gtk is not an available package.

Crouton fails to create a chroot if the netsurf-gtk package is not available.

netsurf-gtk is not available in certain debian releases, including jessie and stretch as of 4/1/16.

### DIFF
--- a/targets/gtk-extra
+++ b/targets/gtk-extra
@@ -8,12 +8,16 @@ DESCRIPTION='GTK-based tools including gdebi, gksu, and a simple browser.'
 
 ### Append to prepare.sh:
 install_dummy network-manager network-manager-gnome
-install gdebi gksu netsurf-gtk
+install gdebi gksu
 
-# Add netsurf-gtk to Debian-alternatives if not already there
-netsurf='/usr/bin/netsurf-gtk'
-for link in x-www-browser gnome-www-browser; do
-    if ! update-alternatives --query "$link" | grep -q "$netsurf"; then
-        update-alternatives --install "/usr/bin/$link" "$link" "$netsurf" 10
+for BROWSER in netsurf-gtk dillo hv3 ""; do
+    if install ${BROWSER}; then
+        EXECUTABLE="/usr/bin/${BROWSER}"
+        for link in x-www-browser gnome-www-browser; do
+            if ! update-alternatives --query "$link" | grep -q "$EXECUTABLE"; then
+                update-alternatives --install "/usr/bin/$link" "$link" "$EXECUTABLE" 10
+            fi
+        done
+        break
     fi
 done


zxvv:
I suspect there may be other ways to handle missing packages.
So please let me know if this proposal is the wrong approach.

I believe that netsuff-gtk might not be essential to a functional chroot, so I'm assuming that continuing to install without it would be reasonable.

Anyway, this is just an idea.

dnschneid:
Thanks for the patch!  One of gtk-extra's goals (as a base of all gtk-based environments) is to provide a graphical browser with a tiny footprint -- just functional enough so someone can search the internet for "how do I get out of my crouton chroot" and read the response.

If netsurf-gtk isn't available (and not just under a new name), we need to find a replacement to fill its tiny shoes.

divx118:
midori could be a good candidate http://pkg.kali.org/pkg/midori https://packages.debian.org/search?keywords=midori

dnschneid:
It's pretty lightweight as far as webkit goes, but it's *huge* size-wise.

divx118:
Oh yes you are right, didn't notice netsurf was so small.

zxvv:
Here is info on executable size of dillo, and netsurf for comparison.

-rwxr-xr-x 1 root root 1.6M May 25  2012 /usr/bin/netsurf-gtk
-rwxr-xr-x 1 root root 685K Jul 21  2013 /usr/bin/dillo

divx118:
+1 for dillo. It has even a smaller footprint then netsurf, less dependencies.

zxvv:
Here is a revised proposal that now does this:

if netsurf-gtk is unavailable, install dillo. 
If dillo is unavailable, continue.

The motivation here is that otherwise, one cannot create a chroot with xfce or lxde on certain recent debian and kali releases.

targets/gtk-extra
```diff
@@ -8,12 +8,23 @@ DESCRIPTION='GTK-based tools including gdebi, gksu, and a simple browser.'
...
+    netsurf='/usr/bin/netsurf-gtk'
+    for link in x-www-browser gnome-www-browser; do
+        if ! update-alternatives --query "$link" | grep -q "$netsurf"; then
+            update-alternatives --install "/usr/bin/$link" "$link" "$netsurf" 10
+        fi
+    done
+else
+    if install dillo; then
+       dillo='/usr/bin/dillo'
+        for link in x-www-browser gnome-www-browser; do
```
dnschneid: is this needed for the dillo install (i.e., it doesn't do it automatically)?  If so, please restructure the code to avoid duplication.

dnschneid:
I initially declined to use Dillo; I forget if this was due to its unusual UI or its inability to render google.com.  As a fall-back for Netsurf it's probably OK, but if there's another option it'd be nice to find it.

zxvv:
Yea, I was unable to find any other browsers with a small install and also in the repositories.

So I surveyed all the browsers available in recent ubuntu releases, below.

There are small text mode ones, but it sounds that's not is desired.
There's netsurf and dillo at ~3MB installed.
There's hv3 and qupzilla at 15 to 20 MB installed.
Then there are the other webkit based ones which are much larger.

firefox - big
lynx - text mode
arora - QT + webkit = big
chromium-browser
dillo - 3MB
epiphany-browser - GTK, XML cairo sqlite pango, etc.
hv3 - 18MB
konqueror - big
midori - big
netsurf-gtk 3MB
qupzilla - 15MB for QT plus lots moe.
surf - 40MB for webkit
xombrero - 40MB for webkit

So, nothing graphical, small, and available in recent debian and kali releases other than dillo, as far as I can see.'

Not that I want to advocated dillo in any way. -- it's simply the only thing I could identify that fits.

If 40MB was under consideration, some of the webkit based options might provide better rendering.

dnschneid:
hv3 would be the way to go, except 
> ECMAScript (a.k.a. javascript) is not supported since the corresponding library (SEE) is not available in Debian yet.

Sid has netsurf-gtk, so I'm not really sure why stretch is missing it.  Perhaps this is a transient issue and we can live with a subpar fallback.  In that case, I'd err on the side of small...but Dillo is a bit painful.  Maybe hv3 is a good compromise?

zxvv:
Yes, whichever you prefer.  

Yea, since netsuft builds on sid, it ought to be a matter of time till it makes it to the other recent release branches.  But that's speculation, it could be months.

I just took hv3 for a spin.  Apt-get said the installed size with dependencies was 27.6 MB.
It doesn't do https. (or javascript of course)
That may be an obstacle for any sites that require login.

so I tried a couple of the smaller webkit based:

surf installed size was 40MB, but it failed in libGL looking for fbConfigs and swrast driver.
That's probably a misconfiguration of webkit, since the others had similar error messages, but succeeded in rendering.

xombrero was 43.2 MB installed, and appears to render quite nicely.

midori was 48.7 MB installed, and renders nicely likewise.

epiphany-browser was 48 MB installed, and renders nicely likewise.

qupzilla was 100 MB installed.

Sigh, webkit is desirable, but 40MB is quite an increment.
Vs 27MB, it's certainly worth the extra 13MB though, so perhaps better than hv3.

zxvv:
BWT, sorry for the changing estimates.
My initial estimates were from reported sizes searching packages on packages.debian.org site.
The later estimates were sizes reported by 'apt-get install'.

divx118:
IMO if you want a small footprint go with dillo, it is just to google your way out of a chroot. At least I don't use netsurf either in any of my chroots.  
If you want a descent browser that is usable and also available in almost all major linux distros through their package managers then I would choose midori.

dnschneid:
Yeah, it sounds like Dillo is the way to go.  It also looks like Dillo is added to x-www-browser, but not gnome-www-browser.

zxvv:
I believe it updates links for both:
     "for link in x-www-browser gnome-www-browser; do ..."

dnschneid:
Sorry, I meant the Dillo debian package.  It already does x-www-browser.

zxvv:
I removed the the duplicate creation of link 'x-www-browser'.

Sorry for the delay.  I somehow didn't see your inline comment about it till today.

dnschneid:
Thanks.  It'd be nicer if the code for the two branches wasn't mostly just a copy-paste.  Perhaps set some variables in the if blocks and then have a common for loop afterwards?

zxvv:
OK, I'll clean this up, as well as, based on your feedback, LXDE and Mate.

zxvv:
I condensed the two 'if/then/else' branches into a single loop.
I squashed the commits.
I've tested it in a debian stretch chroot.

Just let me know if I can do anything else!

targets/gtk-extra
```diff
@@ -8,12 +8,16 @@ DESCRIPTION='GTK-based tools including gdebi, gksu, and a simple browser.'
...
 install_dummy network-manager network-manager-gnome
-install gdebi gksu netsurf-gtk
+install gdebi gksu
 
-# Add netsurf-gtk to Debian-alternatives if not already there
-netsurf='/usr/bin/netsurf-gtk'
-for link in x-www-browser gnome-www-browser; do
-    if ! update-alternatives --query "$link" | grep -q "$netsurf"; then
-        update-alternatives --install "/usr/bin/$link" "$link" "$netsurf" 10
+for BROWSER in netsurf-gtk dillo; do
```
dnschneid: Very nicely structured.  If no browsers succeed, can you fail out?  One easy way is like this:

```
for BROWSER in netsurf-gtk dillo ""; do
    test -n "$BROWSER"
    if install .....
```

Then if dillo doesn't succeed, it'll try the blank string which will fail the test.

targets/gtk-extra
```diff
@@ -8,12 +8,16 @@ DESCRIPTION='GTK-based tools including gdebi, gksu, and a simple browser.'
...
+    if install ${BROWSER}; then
+        EXECUTABLE="/usr/bin/${BROWSER}"
+        for link in x-www-browser gnome-www-browser; do
+            if ! update-alternatives --query "$link" | grep -q "$EXECUTABLE"; then
+                update-alternatives --install "/usr/bin/$link" "$link" "$EXECUTABLE" 10
+            fi
+        done
+	break
```
dnschneid: indentation

zxvv:
I believe I fixed the indentation, and aded a "" to the loop so that it fails out if no browser is installed.

One additional concern is that if no browser is installed, users may not understand why xfce, lxde or other desktops fail to install on recent releases such as xenial, and report this as an issue.  I added a third browser, hv3, which you had mentioned as a possible next alternative, to try to mitigate this.

Just let me know if there is anything else needed!

zxvv:
Is there any alternative to failing installs for xfce, lxde, lxqt, mate, etc, when no browser is available?  

We could consider extending the list of browsers, possibly including xombrero, midori, epiphany-browser and/or qupzilla (in order of increasing install size).

ryanpwest:
Any update on this? I am currently unable to install crouton because of this error.

zxvv:
I've tried to address all the issues dnschneid raised, so we're waiting for him to review.

dwks:
I second this patch, had to install xfce manually because of this error. Maybe the scripts could also have a -k "keep-going" mode that ignores certain missing packages in case something like this breaks in the future.

dnschneid:
Sorry for the delay!  This looks great.

aivijay:
Has the issue with "unable to locate netsurf-gtk" in stretch?
Had been trying to install debian stretch on Asus flip and things stop with netsurf-gtk.

DennisLfromGA:
@aivijay,

It's a work in progress, it has to have a final-review and get merged.

-**DennisL**

dnschneid:
So, ran tests on this.  It doesn't seem to work properly:

```
3652 [e] E: Unable to locate package netsurf-gtk
3652 [e] update-alternatives: error: no alternatives for x-www-browser
3652 [e] update-alternatives: error: alternative path /usr/bin/netsurf-gtk doesn't exist
```

Without looking into it further, that suggests to me that `install` is always returning success.  Has the current version of the code worked for you?

(Again, very sorry for the long turnaround times)

zxvv:
Hey, no worries, and thanks so much this and for the Mate and LXQT testing logs.  

This has been working for me for the certain distros I tested, but that's a limited subset at best.  Let me reproduce this, and study the autotest logs to understand the issue better.  I should be able to respond this weekend about the netsuft-gtk, Mate and LXQT PRs.

I've been wishing I had the ability to test more distros, and seeing this, I'd love run the autotest myself, but I can't tell from reading cruton and test/* whether it's needs some infrastructure I don't have.  Maintaining some extra VMs to support it is no trouble.  I'll look closer at autotest at a later date.

dnschneid:
You can create tests like I did in 32c3f99c5f3dbc770fff6b12ef404f8c8ec7cc26.  Then you just need to download a snapshot of the repo onto your chromebook, and run the tests/run.sh script from crosh (not inside the chroot).  You can run `./run.sh -h` to see the parameters, since you'll probably want to specify specific tests to run.

zxvv:
Awesome.  Thanks so much for the pointers. This helps tremendously.

Elowan:
hey, any news on this? Install still fails, because of missing netsufe. did the xfce4 install instead, but want gnome3 DE... is there a chance? I even did install Gnome3 from the xfce4 terminal, but starting with "sudo startxiwi -T gnome-session just brings up a blank/black screen with a mouse cursor... even when I got "apt-get install gnome-core kali-defaults kali-root-login desktop-base" installed.

bpd1069:
Perhaps a new target, nobrowser, which bypasses the installation of a browser by default, skipping over the apt-get install netsurf-gtk directive as well as the addition to alternatives for browsers.
This would not solve the underlying issue, but would give end-users an avenue to install the chroot and add a browser later, or use the chrome target during setup/upgrade.

Currently I must wait for the installation to fail, then: enter-chroot -u root, then edit /prepare.sh to comment the above directives.

hgrubbs:
Any chance of this getting merged soon? Still have to manually edit prepare.sh after failed setup, then re-enter chroot and let it finish to get a xfce/lxde target working with stretch.

dnschneid:
Woohoo!  Thanks for the patch.

DmitriyPlatonov:
Very well, but now:
E: Package 'gnome-session-fallback' has no installation candidate

dnschneid:
#2429

PR #2525, opened 2016-04-10
LXQT

zxvv:
This adds the target: lxqt

It consists of lxqt, lxqt-common and dependent packages which are ~32MB installed.

These packages are available only on ubuntu vivid and xenial currently.

It depends on gtk-extra for a browser.

### DIFF
--- a/targets/lxqt
+++ b/targets/lxqt
@@ -1,0 +1,29 @@
+#!/bin/sh -e
+# Copyright (c) 2016 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+if [ "${TARGETNOINSTALL:-c}" = 'c' ]; then
+   if release -lt trusty; then
+       error 99 "LXQT packages are not available for Ubuntu releases prior to xenial."
+   fi
+   if release -lt stretch; then
+       error 99 "LXQT packages are not available for Debian releases prior to stretch."
+   fi
+   if release -lt kali-rolling; then
+       error 99 "LXQT packages are not available for Kali Linux prior to kali-rolling."
+   fi
+fi
+
+REQUIRES='x11'
+DESCRIPTION='Installs the LXQT desktop environment. (Approx. 1GB)'
+HOSTBIN='startlxqt'
+CHROOTBIN='crouton-noroot startlxqt'
+. "${TARGETSDIR:="$PWD"}/common"
+
+### Append to prepare.sh:
+install lxqt lxqt-common openbox
+
+TIPS="$TIPS
+You can start LXQT via the startlxqt host command: sudo startlxqt
+"
--- a/chroot-bin/startlxqt
+++ b/chroot-bin/startlxqt
@@ -1,0 +1,1 @@
+crouton-noroot
--- a/host-bin/startlxqt
+++ b/host-bin/startlxqt
@@ -1,0 +1,18 @@
+#!/bin/sh -e
+# Copyright (c) 2016 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+set -e
+
+APPLICATION="${0##*/}"
+
+USAGE="$APPLICATION [options]
+
+Wraps enter-chroot to start an LXQT session.
+By default, it will log into the primary user on the first chroot found.
+
+Options are directly passed to enter-chroot; run enter-chroot to list them."
+
+exec sh -e "`dirname "\`readlink -f "$0"\`"`/enter-chroot" -t lxqt "$@" "" \
+    exec xinit /usr/bin/startlxqt
--- a/test/tests/x0-alltargets
+++ b/test/tests/x0-alltargets
@@ -26,7 +26,7 @@ for target in "$SCRIPTDIR/targets/"*; do
     # Some other targets do not require testing in this context,
     # or have their own w* tests
     for blacklist in audio core x11 xephyr xiwi xorg \
-                     e17 gnome kde lxde unity xbmc xfce; do
+                     e17 gnome kde lxde lxqt unity xbmc xfce; do
         if [ "$target" = "$blacklist" ]; then
             break
         fi


zxvv:
I removed the gtk-extra dependency, because lxde includes qupzilla, which is webkit based.

The installed size is ~1GB, due to the included apps.

The memory usage is relatively lean. Roughly 10MB more than XFCE.
The default theme has a flat color palette.

In addition to ubuntu vivid and xenial, it's also available in debian stretch and sid.

DennisLfromGA:
@zxvv - I may try this out, lxde is currently my go-to desktop so this should work out well for me.

dnschneid:
Thanks!  Can you please rebase (or squash) this branch to get rid of the extra commits and reverts?

Also, at a minimum this target should depend on x11.

The currently supported releases in crouton are: wheezy jessie stretch sid kali-rolling precise trusty.  If any of these don't support lxqt, you need to add some checking code like in the e17 target.

zxvv:
I squashed the commits,
added x11 as a dependency,
and validate the release prior to install.

Just let me know if anything needs attention!

targets/lxqt
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh -e
+# Copyright (c) 2016 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+if [ "${TARGETNOINSTALL:-c}" = 'c' ]
+   if [ "${ARCH#arm}" != "$ARCH" ]; then
+       error 99 "Mate packages are not available for ARM."
```
dnschneid: Are these checks correct?  First of all, it's LXQT.  Secondly, some of the text doesn't match the check (Xenial/Trusty).  Thirdly, are you sure there's no ARM builds of LXQT?  It's not like you're adding a repo or anything...

zxvv:
My intention was for the test to succeed for trusty, xenial and later, as well as, stretch and later.
Those are the releases that have LXQT in the bundled repos.
If that's not what the tests are doing then I've misunderstood what 'release -lt' means.

There is are arm64 builds, but there are none for arm. 

Thanks for catching the "Mate", that was wrong and should have been LXQT.
The only change I've made in the last commit was to fix the "LXQT" in the error messages.
Please let me know if I've misunderstood something about the tests!

zxvv:
I have misunderstood the names for the various arm architectures.

xfce4 is available for arm64, armel and armhf.
likewise, lxde, lxqt and mate are available for arm64, armel and armhf.

I think this means lxde, etc., should be available on ARM just as xfce4 is.
So I'll remove the code that exclude install on "ARM" target for LXQT and Mate.

zxvv:
OK, with this commit, it installs on ARM now as discussed above.
Please let em know if you see any issues!

targets/lxqt
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh -e
+# Copyright (c) 2016 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+if [ "${TARGETNOINSTALL:-c}" = 'c' ]; then
+   if release -lt trusty; then
+       error 99 "LXQT packages are not available for Ubuntu releases prior to xenial."
```
dnschneid: trusty or xenial?

targets/lxqt
```diff
@@ -0,0 +1,29 @@
...
+   fi
+   if release -lt kali-rolling; then
+       error 99 "LXQT packages are not available for Kali Linux prior to kali-rolling."
+   fi
+fi
+
+REQUIRES='x11'
+DESCRIPTION='Installs the LXQT desktop environment. (Approx. 1GB)'
+HOSTBIN='startlxqt'
+CHROOTBIN='crouton-noroot startlxqt'
```
dnschneid: crouton isn't responsible for installing startlxqt inside the chroot since you're using the distro-provided script/binary.

dnschneid: Actually, sorry, startlxqt should be installed.  In chroot-bin, add a symlink to crouton-noroot called startlxqt.

targets/lxqt
zxvv: in this case, trusty.  
Prior to trusty, lxqt is not available at all as far as I've been able to find.

targets/lxqt
zxvv: I've added the symlink: startlxqt -> crouton-noroot
Just let me know if/when you want the commits squashed.

targets/lxqt
dnschneid: Cool. I think the text needs to be fixed, then.

tsimonq2:
Bump, I'm the person working on Lubuntu's LXQt transition, it would be good to have this for my own testing and for when we actually make the transition.

dnschneid:
@tsimonq2 if you fork @zxvv's branch and fix the testing errors, you could submit a PR with your fixups on top of @zxvv's commits.

tsimonq2:
@dnschneid Sure, I'll fork it over the weekend. :)

I'm not entirely sure what the issues are here, even after reading it over.

And it's `LXQt`, not `LXQT`.

dnschneid:
>I'm not entirely sure what the issues are here, even after reading it over.

that's because the issues are confusingly described [here](https://github.com/dnschneid/crouton/pull/2532#issuecomment-227546165).

> And it's LXQt, not LXQT.

this is why i just leave everything lowercase

dnschneid:
(extra patches)
[0001-Fix-LXQT-target-error-message.txt](https://github.com/dnschneid/crouton/files/520410/0001-Fix-LXQT-target-error-message.txt)
[0002-Add-Mate-and-LXQT-w-tests.txt](https://github.com/dnschneid/crouton/files/520409/0002-Add-Mate-and-LXQT-w-tests.txt)

DennisLfromGA:
I ran into a conflict with a package that the `lxqt` target installs: 'connman' 
with one that `gtk-extra` installs: 'network-manager'

`gtk-extra:install_dummy network-manager network-manager-gnome`

See this [thread and comment](https://github.com/dnschneid/crouton/issues/2883#issuecomment-260160961) for a few more details.

In the `installer/prepare.sh` script, the 'install' function can be used to *avoid* packages too, here's a snippet from the comments:

> \# install [--minimal] [--asdeps] <packages> \-\- \<avoid packages\>
\# For the specified crouton-style package names (see distropkgs()), installs
\# \<packages\>, while avoiding installing \<avoid packages\> if they are not
\# already installed.

There are a number of targets that *avoid* installing 'network-manager' to eliminate these conflicts, I've listed a few below -
```
chronos@localhost ~/Downloads/crouton.unbundled $ sudo grep network-manager targets/*
targets/kde:install --minimal kde-baseapps kde-runtime plasma-desktop pulseaudio \
targets/kde:        -- network-manager
targets/kde-desktop:install ubuntu=kubuntu-desktop,task-kde-desktop -- network-manager
targets/lxde:install lxde -- dictionaries-common chromium-browser network-manager wicd
targets/lxde-desktop:install ubuntu=lubuntu-desktop,task-lxde-desktop -- network-manager
targets/unity-desktop:install ubuntu-desktop -- network-manager xorg
targets/xfce-desktop:install ubuntu=xubuntu-desktop,task-xfce-desktop -- network-manager xorg
```

I don't know if I understand the 'install' function well enough to make any suggestions but your `lxqt` target could possibly drop the 'connman' package requirement by using something like:

`install lxqt lxqt-common openbox -- connman`

I'm not sure that's the correct syntax though.

I suppose I could have avoided this in my case by just not installing `gtk-extra` but others may want this and other desktops along with lxqt so I thought I'd meniton it.

-**DennisL**

Serkan-devel:
Is this still being worked on?

Issue #2520, opened 2016-04-06
Will there ever be a mate target?


JaxonWright:
I would love to have mate on my chromebook, as it is what I am most used to.

Polygon-6789:
You can install MATE by using the x11 target and then installing the MATE desktop packages yourself.

zxvv:
I can take a stab at it.

Based on what I see recommended, the gist of it would be:

sudo apt-get install --no-install-recommends ubuntu-mate-core ubuntu-mate-desktop

zxvv:
The pull request supports mate on most ubuntu, and a few recent debian releases.

There is one caveat when running mate in a window using xiwi. Xiwi runs the i3 window manager.  Due to differences betwen mate's 'marco' window manager and i3, some desktop items display error dialogs. This is not an issue when it is started using the xorg target.

gardwired:
working for me -- sourced from various threads here:
crouton -r jessie -t audio,x11,xorg  (not actually sure you need targets at all)
then in chroot:  apt-get  install task-mate-desktop

create .xinitrc
exec mate-session

to start x (inside chroot): xinit  
startx doesn't work

PR #2532, opened 2016-04-15
Mate

zxvv:
this adds the target 'mate' consisting of the mate-desktop-environment package, available on most ubuntu releases, and debian jessie and newer.

### DIFF
--- a/targets/mate
+++ b/targets/mate
@@ -21,10 +21,19 @@ HOSTBIN='startmate'
 CHROOTBIN='crouton-noroot startmate'
 . "${TARGETSDIR:="$PWD"}/common"
 
+EXTRAPACKAGES=
+
 if release -le precise ; then
     install software-properties-common
     sudo apt-add-repository -y "deb http://packages.mate-desktop.org/repo/ubuntu precise main"
     sudo apt-get update
+fi
+
+if release -ge xenial ; then
+    install software-properties-common
+    sudo apt-add-repository -y ppa:ubuntu-mate-dev/xenial-mate
+    sudo apt-get update
+    EXTRAPACKAGES="${EXTRAPACKAGES} caja-actions ubuntu-mate-artwork"
 fi
 
 if release -eq trusty && [ "$ARCH" = 'amd64' -o "$ARCH" = 'i386' ]; then
@@ -34,7 +43,8 @@ if release -eq trusty && [ "$ARCH" = 'amd64' -o "$ARCH" = 'i386' ]; then
     sudo apt-get update
 fi
 
-install mate-desktop-environment
+
+install mate-desktop-environment ${EXTRAPACKAGES}
 
 TIPS="$TIPS
 You can start mate via the startmate host command: sudo startmate
--- a/targets/mate
+++ b/targets/mate
@@ -1,0 +1,41 @@
+#!/bin/sh -e
+# Copyright (c) 2016 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+if [ "${TARGETNOINSTALL:-c}" = 'c' ]
+   if release -lt trusty; then
+       error 99 "Mate packages are not available for Ubuntu releases prior to trusty."
+   fi
+   if release -lt jessie; then
+       error 99 "Mate packages are not available for Debian releases prior to jessie."
+   fi
+   if release -lt sana; then
+       error 99 "Mate packages are not available for Kali Linux prior to sana."
+   fi
+fi
+
+REQUIRES='x11'
+DESCRIPTION='Installs the mate desktop environment. (Approx. 205MB)'
+HOSTBIN='startmate'
+CHROOTBIN='crouton-noroot startmate'
+. "${TARGETSDIR:="$PWD"}/common"
+
+if release -le precise ; then
+    install software-properties-common
+    sudo apt-add-repository -y "deb http://packages.mate-desktop.org/repo/ubuntu precise main"
+    sudo apt-get update
+fi
+
+if release -eq trusty && [ "$ARCH" = 'amd64' -o "$ARCH" = 'i386' ]; then
+    install software-properties-common
+    sudo apt-add-repository -y ppa:ubuntu-mate-dev/ppa
+    sudo apt-add-repository -y ppa:ubuntu-mate-dev/trusty-mate
+    sudo apt-get update
+fi
+
+install mate-desktop-environment
+
+TIPS="$TIPS
+You can start mate via the startmate host command: sudo startmate
+"
--- a/targets/mate
+++ b/targets/mate
@@ -36,7 +36,7 @@ if release -ge xenial ; then
     EXTRAPACKAGES="${EXTRAPACKAGES} caja-actions ubuntu-mate-artwork"
 fi
 
-if release -eq trusty && [ "$ARCH" = 'amd64' -o "$ARCH" = 'i386' ]; then
+if release -eq trusty ; then
     install software-properties-common
     sudo apt-add-repository -y ppa:ubuntu-mate-dev/ppa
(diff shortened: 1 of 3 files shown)


zxvv:
One can start mate natively using:
startmate

Or, one can start mate in xiwi using:
sudo startxiwi -F mate-session

However, when using xiwi, some desktop items display error dialogs due to differences from mate's 'marco' window manager.

This is not an issue when it is started natively using 'startmate'.

zxvv:
This addresses the request for a mate target: #2520

JaxonWright:
Thanks for this. Hopefully it gets added soon!

targets/mate
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh -e
+# Copyright (c) 2016 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+DESCRIPTION='Installs the mate desktop environment. (Approx. 205MB)'
+HOSTBIN='startmate'
+CHROOTBIN='crouton-noroot startmate'
+. "${TARGETSDIR:="$PWD"}/common"
+
+if release -lt jessie ; then
```
dnschneid: This should check should be done prior to installation, like in the e17 target.  It looks like you might also need a check for ARM + PPA?

targets/mate
```diff
@@ -0,0 +1,32 @@
...
+    echo mate is not available in the repositories for this release.
```
dnschneid: use the "error" function and quote the text

dnschneid:
Thanks again.  Similar to the LXQT addition, if all the supported releases (wheezy jessie stretch sid kali-rolling precise trusty) and arches (i386, x86_64, arm) aren't covered, there needs to be tests prior to installation.  You already have some of the tests, but please make sure everything is there.

zxvv:
I've added the checks, prior to install, for unsupported arch (arm) and unsupported releases.

I tried and failed to squash.  I'll try to resolve that.

zxvv:
OK, I believe I've squashed the commits into a single commit.
Please let me know if there's anything else that needs attention!

targets/mate
```diff
@@ -0,0 +1,44 @@
+#!/bin/sh -e
+# Copyright (c) 2016 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+if [ "${TARGETNOINSTALL:-c}" = 'c' ]
+   if [ "${ARCH#arm}" != "$ARCH" ]; then
+       error 99 "Mate packages are not available for ARM."
```
dnschneid: Ooh, sorry, I should have realized earlier that Mate isn't available for ARM.
Historically I haven't merged DEs unless they're available on a vast majority of Chromebooks (there's quite a lot of ARM) and on a majority of the supported releases.

I think Mate now squeaks by on the latter point, but I'm extremely hesitant to include something that only works on certain architectures.

@drinkcat @DennisLfromGA thoughts?

dnschneid: Or is Mate available on ARM on later releases?

zxvv:
It's available on ARM64 but not ARM.  The same is true for lxqt.

But "${ARCH#arm}" != "$ARCH" was wrong.
I changed it to "${ARCH}" = "arm" so that it excludes only arm and will install on arm64.

As an aside, a few others such as chromium are available on arm64 in debian jessie and stretch.
It might be worth considering installing on arm64 for those.

zxvv:
I reread the xfce4 target, and see that xfce4 is available for arm64, armel and armhf, and lxde, lxqt and mate are available for the same -- arm64, armel and armhf.

So, the above commit removes the code that excludes the ARM target5.
it installs on ARM now. (I discussed this in the LXQT PR as well).

targets/mate
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh -e
+# Copyright (c) 2016 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+if [ "${TARGETNOINSTALL:-c}" = 'c' ]
```
Morgul: This should be `if [ "${TARGETNOINSTALL:-c}" = 'c' ]; then`; bash complains about an unmatched `fi` otherwise.

targets/mate
```diff
@@ -0,0 +1,41 @@
...
+fi
+
+if release -eq trusty && [ "$ARCH" = 'amd64' -o "$ARCH" = 'i386' ]; then
+    install software-properties-common
+    sudo apt-add-repository -y ppa:ubuntu-mate-dev/ppa
+    sudo apt-add-repository -y ppa:ubuntu-mate-dev/trusty-mate
+    sudo apt-get update
+fi
+
+install mate-desktop-environment
```
Morgul: This does not work in xenial; you will need to use `ppa:ubuntu-mate-dev/xenial-mate`.

zxvv: Are you saying that mate-desktop-environment is broken on xenial? (I just want to be sure I understand, since mate-desktop-environment is available to install from the bundled repos and is able to start a session.)

zxvv:
On xenial, the ppa:ubuntu-mate-dev/xenial-mate repo is added, and
the caja-actions ubuntu-mate-artwork that it provides are installed.

Thanks for the help. Please let me know if there are remaining issues.

targets/mate
```diff
@@ -0,0 +1,51 @@
...
+DESCRIPTION='Installs the mate desktop environment. (Approx. 205MB)'
+HOSTBIN='startmate'
+CHROOTBIN='crouton-noroot startmate'
+. "${TARGETSDIR:="$PWD"}/common"
+
+EXTRAPACKAGES=
+
+if release -le precise ; then
+    install software-properties-common
+    sudo apt-add-repository -y "deb http://packages.mate-desktop.org/repo/ubuntu precise main"
```
dnschneid: The script is already run as root, so sudo isn't necessary anywhere in here.

targets/mate
```diff
@@ -0,0 +1,51 @@
...
+if release -ge xenial ; then
+    install software-properties-common
+    sudo apt-add-repository -y ppa:ubuntu-mate-dev/xenial-mate
+    sudo apt-get update
+    EXTRAPACKAGES="${EXTRAPACKAGES} caja-actions ubuntu-mate-artwork"
+fi
+
+if release -eq trusty && [ "$ARCH" = 'amd64' -o "$ARCH" = 'i386' ]; then
+    install software-properties-common
+    sudo apt-add-repository -y ppa:ubuntu-mate-dev/ppa
```
dnschneid: Both ppa and trusty-mate are required?

Also, does this work with ARM on trusty?

targets/mate
dnschneid: Will this work with arm on precise?

dnschneid:
Sorry for the long delay in replying!

targets/mate
zxvv: Yes, there are arm binaryies for trusty:
http://ppa.launchpad.net/ubuntu-mate-dev/trusty-mate/ubuntu/dists/trusty/main/

So that needs to be changed.  It shouldn't be only amd64 and i386.

zxvv:
On trusty, it allows install on arm now.
If/when you want the commits squashed just let me know!

dnschneid:
I ran this and #2525 through the autotester, and both had lots of problems across different releases.

Here's the [test output](https://github.com/dnschneid/crouton/files/326427/2016-06-21_01-36-49_dnschneid_crouton_mergeboat_w8_w9.zip).  If you extract it somewhere and open status.html, you'll see a bunch of red logs, which you can open up and look for more details as to what failed.

trusty, for example, seems to be missing /usr/bin/mate-session.

JaxonWright:
Any progress on this? I would love a simple way to get Mate

farinc:
I know this is old, but there any more progress on this? Like @JaxonWright, I would love Mate in my chroot.

Issue #2181, opened 2015-10-12
xiwi screen freezes and repeats last key stroke. Error: find_shm: Error running helper.


hsuijs:
When I start my crouton session with xiwi in a window, after some time running I get screen freezes after which key strokes are repeated multiple times and sometimes the screen turns black for a second.

After the first freeze (which can take some time to appear) it gets worse really quickly with freezes every 10-30 seconds.

Most common errors around this time:
popen2: child exited with status 1
find_shm: Error running helper.
write_image: Cannot find shm, moving on...

See the complete log of a session:
X.Org X Server 1.15.1
Release Date: 2014-04-13
X Protocol Version 11, Revision 0
Build Operating System: Linux 3.2.0-76-generic x86_64 Ubuntu
Current Operating System: Linux localhost 3.14.0 #1 SMP Fri Sep 18 00:42:28 PDT 2015 x86_64
Kernel command line: cros_secure console= loglevel=7 init=/sbin/init cros_secure oops=panic panic=-1 root=/dev/dm-0 rootwait ro dm_verity.error_behavior=3 dm_verity.max_bios=-1 dm_verity.dev_wait=1 dm="1 vroot none ro 1,0 2506752 verity payload=PARTUUID=92ff718b-b6af-e344-ba30-7a3f500db8fd/PARTNROFF=1 hashtree=PARTUUID=92ff718b-b6af-e344-ba30-7a3f500db8fd/PARTNROFF=1 hashstart=2506752 alg=sha1 root_hexdigest=f9baec8f94f6518ad147cd0f0a28f37fd961cff2 salt=6f27c81451b6ac1d2c19b083045b515b6946c059cc8743bffcf8bdb36b21160d" noinitrd vt.global_cursor_default=0 kern_guid=92ff718b-b6af-e344-ba30-7a3f500db8fd add_efi_memmap boot=local noresume noswap i915.modeset=1 tpm_tis.force=1 tpm_tis.interrupts=0 nmi_watchdog=panic,lapic  
Build Date: 12 February 2015  02:49:29PM
xorg-server 2:1.15.1-0ubuntu2.7 (For technical support please see http://www.ubuntu.com/support) 
Current version of pixman: 0.30.2
        Before reporting problems, check http://wiki.x.org
        to make sure that you have the latest version.
Markers: (--) probed, (**) from config file, (==) default setting,
        (++) from command line, (!!) notice, (II) informational,
[...]

PR #2621, opened 2016-06-11
Pass error messages to findnacld. (refs #2181)

monaka:
For maintainability. This will help to see the reason why croutonfindnacl was failed.

### DIFF
--- a/chroot-bin/croutonfindnacl
+++ b/chroot-bin/croutonfindnacl
@@ -20,6 +20,8 @@
 #              Space-separated list of PIDs to scan for.
 #
 # On success, prints "pid:filename" and exits with code 0.
+# On error (flie not found, duplicate files found),
+#    prints "-1:cause" and exits with code 0.
 # On error (shm not found, invalid parameters), exits with code >0.
 
 set -e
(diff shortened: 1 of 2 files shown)


dnschneid:
Why exit with code 0 though?  It did fail...

monaka:
On current code, `findnacld` (and developers) has no way to get the reason why `croutonfindnacl` was terminated.

dnschneid:
Yeah, I like that it outputs text, but shouldn't the exit code still be 1?  Or are you saying that's bad because it causes issues with `set -e` and `$(findnacld)`?

dnschneid:
I think scripts that use `set -e` `findnacld` should ignore the return value (`|| true`) if they don't want a `$(findnacld)` failure to exit.  Scripts that don't need to display the error message shouldn't have to parse the output just to figure out if the script succeeded.

monaka:
After fixed #2622, it will be possible to get "-1:no match" error.
I think this error will be very rare case but be possible.

dnschneid:
Thanks for the patch, but I don't think it's worth changing the semantics of findnacld for this.

monaka:
I see. Thanks for your consideration and decision.

PR #2622, opened 2016-06-12
Fix abnormal termination at calling readlink in croutonfindnacl.

monaka:
This patch is for fixing #2181.
In some situations, grabbed files in `/proc` are deleted before `readlink` calls.
In such case, `croutonfindnacl` is exited with its error code 1.
They may be caused by heavy CPU load, many file accesses.

The patch is adding `set +e / set -e` and explicit error checking around `readlink`.

### DIFF
--- a/chroot-bin/croutonfindnacl
+++ b/chroot-bin/croutonfindnacl
@@ -52,7 +52,7 @@ for pid in $PIDS; do
 
     # Iterate over mappings, and check signature
     for fd in "/proc/$pid/fd"/*; do
-        link="`readlink -- "$fd"`"
+        link="$(readlink -- "$fd" || true)"
         link="${link% (deleted)}"
         if [ "$link" = "$file" ]; then
             # Check if signature matches


monaka:
In some rare cases, `croutonfindnacl` may fail to get the file. So #2621 also should be applied, IMO

chroot-bin/croutonfindnacl
```diff
@@ -52,7 +52,10 @@ for pid in $PIDS; do
 
     # Iterate over mappings, and check signature
     for fd in "/proc/$pid/fd"/*; do
+	set +e
         link="`readlink -- "$fd"`"
```
dnschneid: I think you can just add || true and the if statement below will naturally fall through.

`link="$(readlink -- "$fd" || true)"`

dnschneid:
Thanks for debugging this!

monaka:
@dnschneid I see. Your suggestion looks better than my patch.
I'll test it on my environment to make sure. And I'll re-push it this weekend.
Please stay tune.

monaka:
I'm using Crouton for a week with the patch re-pushed now. It works fine for me.

dnschneid:
Thanks again for figuring this out, and apologies for the delay in merging.

PR #2635, opened 2016-06-19
Nothing that has passed final release should be marked "test".

StrawnPoint04:


dnschneid:
Actually, the `test` tag forces our autotester to run tests on that release, even if the release is marked "unsupported" (the `*`).  The idea being that releases we want to eventually support should be tested even if we haven't done the push to fully get everything working (which hasn't happened successfully since trusty, unfortunately).

So those releases should be flagged for testing until one of two situations:
1. We get the release stable to the point where we can claim it's fully supported and remove both the `*` and the `|test`.
2. The release goes EOL, in which case there's no point in doing automated testing on it (keep the `*`, remove the `|test`).

PR #2646, opened 2016-06-25
`sudo bash enter-chroot` doesn't mount Downloads

mjwhitta:
For some reason, `startlxde` (and I assume others) properly substitute
`~` before mounting Downloads, however just calling `enter-chroot`
does not. After escaping the `~`, both `startlxde` and `enter-chroot`
work as expected.

### DIFF
--- a/host-bin/enter-chroot
+++ b/host-bin/enter-chroot
@@ -502,7 +502,7 @@ if [ -z "$NOLOGIN" -a -f "$shares" ]; then
                 continue;;
         esac
         # Expand dest for homedirs
-        if [ "${dest#~}" != "$dest" ]; then
+        if [ "${dest#"~"}" != "$dest" ]; then
             destuser="${dest%%/*}"
             if [ "$destuser" = '~' ]; then
                 if [ -z "$CHROOTHOME" ]; then
@@ -511,7 +511,7 @@ if [ -z "$NOLOGIN" -a -f "$shares" ]; then
                 fi
                 dest="$CHROOTHOME/${dest#*/}"
             else
-                dest="/home/${destuser#~}/${dest#*/}"
+                dest="/home/${destuser#"~"}/${dest#*/}"
             fi
         fi
         # Do the bindmount


DennisLfromGA:
@mjwhitta - This *may* not pertain to the issue but the invocation of 'sh' in your example is not needed. Just use `sudo enter-chroot` instead.

mjwhitta:
So in the ~3 years I've had my Acer C720, I've never been able to execute any shell script directly.

```
$ sudo ./enter-chroot
sudo: unable to execute ./enter-chroot: Permission denied
```

This has always seemed weird to me as I've seen comments here that imply it works for others. I've powerwashed my chromebook numerous times, and this behavior has never changed. Regardless, the `startlxde` script uses `sh` to call `enter-chroot`, so I'm pretty sure it doesn't pertain to my issue.

DennisLfromGA:
@mjwhitta,

*[ I am by no means qualified to answer this but I thought I'd take a stab at it. ]*

I'm not saying your PR is not warranted but, since you're having problems with normal execution, maybe that's why this 'fix' is needed for you, I don't think others have run into this problem.

The tilde (~) character is used/expanded in the bash shell but not the normal posix shell so that may be part of the problem. In your crosh shell session, you're using 'bash' so the tilde is expanded.
`enter-chroot` invokes the posix shell on line 1: `#!/bin/sh -e` so the tilde has to be handled differently. 

----
Having to use 'sh' in every case does indeed sound strange, it sounds like `/usr/local/bin` is not 'executable'. Are you using the internal SSD and have your chroot(s) installed in the default location?

Here's what mine looks like -
```
chronos@localhost ~ $ mount|grep local
/dev/sda1 on /usr/local type ext4 (rw,nodev,relatime,commit=600,data=ordered)
chronos@localhost ~ $ ls -ld /usr/local/bin
drwxr-xr-x 2 root root 4.0K Jun 21 17:10 /usr/local/bin
```

Maybe some of the gurus lurking about can jump in and explain what might be going on with your system and why... 

-----
**Additional info.** *( please disregard if you like )* -
When I first got my Cr-48, I used to build my own Chrome OS images and also used some from *real* developers. One of them (tvierling) gave me a routine that I've used in my `.bashrc` script on every Chromebook I've had since. It creates a special 'dev_image' and uses it for `/usr/local`.
I've included it below for your perusal and to maybe gain some perspective about what's going on in Chrme OS.
Here's a link where he explains what's going on, 
scroll down to ['[Cr-48]-CHROME-OS Experimental ADD-ON PKGS'](https://www.diigo.com/list/kim122079/chrome_os)
```
##
## Open a shell and run:
#   sudo mkdir /mnt/stateful_partition/dev_image
##
## This will create a new directory in the stateful partition that represents the (modifiable!) /usr/local area.
## The name "dev_image" is semi-magic -- this is the same name that non-official (developer) builds use to represent /usr/local.
## You can actually pick any name, even "local", but the instructions here assume "dev_image".
##
## Mount /usr/local by hand with the following commands:
#   sudo mount -n --bind /mnt/stateful_partition/dev_image /usr/local
#   sudo mount -n -o remount,exec,suid /usr/local
##
## At this point, you should be able to verify that /usr/local is mounted by:
#   mount|grep local
## which should show:
##  /dev/sda1 on /usr/local type ext3 (rw,nodev,relatime,errors=continue,commit=600,data=ordered)
## In particular, the mount should not show the flags noexec or nosuid.
##
## If you're using any package that has a "requires: ..." note below, you need one more command to
## ensure the extra libraries are picked up at runtime.
## To do this, also run:
##  export LD_LIBRARY_PATH=/usr/local/lib
export LD_LIBRARY_PATH=/usr/lib:/usr/local/lib:/usr/local/lib64
##
## The steps above have to be performed on every boot, but to save hassle,
## edit ~/.bashrc and add the following lines to the bottom to mount it as soon as any shell is invoked
## (vt2 or via Ctrl-Alt-T plus "shell" command):
##
   if ! grep -q /usr/local /etc/mtab; then
      [ ! -d /mnt/stateful_partition/dev_image/ ] && sudo mkdir /mnt/stateful_partition/dev_image 2>/dev/null
      sudo mount -n --bind /mnt/stateful_partition/dev_image /usr/local
      sudo mount -n -o remount,exec,suid /usr/local
      echo "...local mounted r-w-x..."
   fi

##
```

*Hope at least some of this helps,*
-**DennisL**

divx118:
@mjwhitta It seems to me that you don't have the execution bit for enter-chroot set.
To be sure post the output of `ls -al` in the directory where you have enter-chroot. Normally it should be in /usr/local/bin.

mjwhitta:
@DennisLfromGA I use a 64GB USB for my crouton chroots. It is indeed mounted noexec so that's likely the reason. I've only ever written bash scripts in `~/Downloads` which is also noexec which is likely the reason it's never worked for me. The mount points being noexec really don't concern me as I can just call `sh` or `bash` directly which should be no different than the shell invoking the interpreter via `#!/usr/bin/env sh` or whatever.

@divx118 The scripts in `/mount/removable/64GB_USB/crouton/bin` are executable, however my USB is mounted noexec.

The problem here is that the `~` needs to be escaped in the regex (or maybe pattern is a better name for it as I not sure it's a proper regex). The `${dest#~}` does not always replace the `~`. In this case, the shell is not doing the `~` expansion, but rather the script is doing a substitution via this line:

```
dest="/home/${destuser#~}/${dest#*/}"
```

My hunch is that I actually call `sudo bash startlxde` which then calls `exec sh -e ... enter-chroot` (which works), but when only wanting a cli, I call `sudo bash enter-chroot` (which doesn't work). This is likely a `bash` vs `sh` thing, but I'm also no expert. However I feel escaping the `~` to work in both `bash` and `sh` shouldn't cause problems and I would be interested to know if my changes break anything for other users.

DennisLfromGA:
@mjwhitta,

So, it turns out you know a *whole* lot more about this than I do and I believe your 'hunch' just might be correct.

Maybe one of the devs will get a chance to look at your PR and chime in.

-**DennisL**

mjwhitta:
Just did a quick test, `sudo sh enter-chroot` works as expected and `sudo bash enter-chroot` fails to mount Downloads. I updated the title to be accurate. I had an alias I set up a long time ago which I thought was using `sh` but was actually using `bash`.

DennisLfromGA:
@mjwhitta - *Great, mystery solved!* You can probably close this issue if you like.

mjwhitta:
If you would like to close it, that's fine. I feel it would still be worth merging to prevent issues in the future, although I'm probably, somehow, the only one that's had this problem (story of my life).

DennisLfromGA:
@mjwhitta,

Nope, I'm sorry, I had totally forgotten about your PR, I was mistakenly thinking this was an 'issue' - my bad.

We'll let the devs weigh in first at least.

-**DennisL**

dnschneid:
crouton is *supposed* to be able to handle being run with bash (and even busybox's sh) as the interpreter, although it primarily targets dash's /bin/sh.  This is usually achieved by carefully only using POSIX features.  That said, weird things like this pop up from time to time, so good job catching another one :)

test.sh:
```
x="$HOME/arst"
y="~/neio"
echo "~: ${x#~}"
echo "~: ${y#~}"
echo "\\~: ${x#\~}"
echo "\\~: ${y#\~}"
echo "\"~\": ${x#"~"}"
echo "\"~\": ${y#"~"}"
```
running it against dash and bash:
```
chronos@localhost ~/Downloads $ /bin/sh test.sh 
~: /home/chronos/user/Downloads/arst
~: /neio
\~: /home/chronos/user/Downloads/arst
\~: /neio
"~": /home/chronos/user/Downloads/arst
"~": /neio
chronos@localhost ~/Downloads $ /bin/bash test.sh 
~: /Downloads/arst
~: ~/neio
\~: /home/chronos/user/Downloads/arst
\~: /neio
"~": /home/chronos/user/Downloads/arst
"~": /neio
```
busybox behaves the same as dash, and zsh behaves the same as bash (for what it's worth).

I prefer the quoting rather than the backslash, though.  Could you please switch to quoting the ~?

mjwhitta:
Done. I amended the commit since I didn't know the cause of the bug when initially making the change. Let me know if there's anything else you need me to do.

dnschneid:
Looks great. Thanks!

PR #2717, opened 2016-08-10
Stop testing Wily; it's EOL

brmbrmcar:
It was done with Vivid. Just a simple fix.


dnschneid:
Goal: fully support xenial before the LTS EOLs 😢

PR #2724, opened 2016-08-14
Removed redundancies in shell scripts

mmattioli:
Removed several uses of `set` which specified options that were already enabled
in the shebang. One instance in which `set` was called on two separate lines was
moved to the shebang for uniformity.


dnschneid:
Thanks for the cleanup, but those are actually necessary.  They were added in 7132db10e9aab456d246a0c21a525a354033e6ce to handle the cases where people run the scripts explicitly with shell (which is necessary on noexec media): `sh enter-chroot ...`

mmattioli:
Ah, didn't see that; those commits were from 2013 so it was a while back in the commit history 😁

Issue #2706, opened 2016-08-01
Google Play crashes on Crouton chroot mount


sshivaji:
```
Please paste the output of the following command here: sudo edit-chroot -all
```
sudo edit-chroot -all
name: trusty
encrypted: no
Entering /media/removable/Crouton/chroots/trusty...
crouton: version 1-20160620130246~master:c39102d9
release: trusty
architecture: amd64
xmethod: xorg
targets: x11,xorg,xfce,touch,unity,extension,core,audio,xiwi
host: version 8530.35.0 (Official Build) dev-channel samus 
kernel: Linux localhost 3.14.0 #1 SMP PREEMPT Thu Jul 28 17:43:58 PDT 2016 x86_64 x86_64 x86_64 GNU/Linux
freon: yes
Unmounting /media/removable/Crouton/chroots/trusty...

#### Please describe your issue:
Latest developer channel chromebook pixel 2 has google play support. However, when I mount a chroot (even do "sudo enter-chroot" in a terminal), google play and many android services stop.

#### If known, describe the steps to reproduce the issue:

1. Update chromebook to google play support (possible with certain chromebooks like Chromebook Pixel 2).
2. Execute "sudo enter-chroot"
3. Google play services will all stop and there will be several error messages popping up.

pjchamberlain:
same issue on Acer r11 on developer channel. Google Play appeared after the latest chrome update. Google play was working and so were the 3 Android apps I installed, but could no longer start Xfce , stuck in a loop requiring a password.

 Fixed crouton by wiping it out and reinstalling from scratch. 

Google Play no longer works at all , fails to start. Error message " unfortunately Google Play has stopped"

tedm:
a bit off topic, but if I update my Toshiba Chromebook2 (2015) to dev channel, can I access the play store?

or is it only 3 systems now, and others on the compatibility list later in the year?

DennisLfromGA:
@tedm - 

Not yet, only the 3 systems have it working on the beta channel now.
A lot of us are waiting for the roll out to other systems but [the site](https://sites.google.com/a/chromium.org/dev/chromium-os/chrome-os-systems-supporting-android-apps?hl=en&source=genius-rts&rd=1) now says: 2016/2017 so we'll have to keep our fingers crossed and see if it arrives this year.

-**DennisL**

tedm:
[...]

PR #2735, opened 2016-08-18
Fix ARC / Google Play for Chrome OS crash when starting chroot

greymouser:
This fixes issue #2706 .

* Remounting the selinuxfs mountpoint /sys/fs/selinux in the chroot as
  read-only doesn't seem to offer the protection one would expect.
* This fix assumes there is no direct need for use of SELinux within
  the chroot (and besides, wouldn't /etc/selinux need to be bind
  mounted, too?)
* TL;DR This fixes the ARC / Google Play breakage when using crouton.

### DIFF
--- a/host-bin/enter-chroot
+++ b/host-bin/enter-chroot
@@ -528,10 +528,10 @@ if ! mountpoint -q "$CHROOT/sys"; then
     mount --make-rshared /sys
     mount --rbind /sys "$CHROOT/sys"
     mount --make-rslave "$CHROOT/sys"
-    # Remount selinux as ro
-    if [ -d "$CHROOT/sys/fs/selinux" ]; then
-        mount -o remount,ro "$CHROOT/sys/fs/selinux"
-    fi
+    # Mounting selinuxfs, remounted as RO or not, causes
+    # ARC / "Google Play for Chrome OS" to crash. The break
+    # is permanent until a clean reboot.
+    umount $CHROOT/sys/fs/selinux
 fi
 
 # Modify chroot's /sys/class/drm to avoid vgem
--- a/host-bin/enter-chroot
+++ b/host-bin/enter-chroot
@@ -531,7 +531,9 @@ if ! mountpoint -q "$CHROOT/sys"; then
     # Mounting selinuxfs, remounted as RO or not, causes
     # ARC / "Google Play for Chrome OS" to crash. The break
     # is permanent until a clean reboot.
-    umount "$CHROOT/sys/fs/selinux"
+    if [ -d "$CHROOT/sys/fs/selinux" ]; then
+    	umount "$CHROOT/sys/fs/selinux" 2>/dev/null
+    fi
 fi
 
 # Modify chroot's /sys/class/drm to avoid vgem
--- a/host-bin/enter-chroot
+++ b/host-bin/enter-chroot
@@ -531,7 +531,7 @@ if ! mountpoint -q "$CHROOT/sys"; then
     # Mounting selinuxfs, remounted as RO or not, causes
     # ARC / "Google Play for Chrome OS" to crash. The break
     # is permanent until a clean reboot.
-    umount $CHROOT/sys/fs/selinux
+    umount "$CHROOT/sys/fs/selinux"
 fi
 
 # Modify chroot's /sys/class/drm to avoid vgem


DennisLfromGA:
I wonder if a check could be made to see if the system has enabled Google Play and, if so, then apply this fix. I don't know where exactly to check though I'm afraid, I don't know if ARC++ will show up anywhere in the `crossystem` output or someplace else.

No biggey I guess, just thinking out loud.

greymouser:
@DennisLfromGA I was curious, so I checked. `crossystem` doesn't seem to afford any information to us, but it's easy enough to detect ala `mount |grep "/run/containers/android"` (seems to return a lot of matches if it's enabled). Knowing the full ramifications of not bind-mounting the selinux mount in the changeroot (why it was there to begin with) would put me at ease, for sure, but ... well, I'm happy enough that this bug seemingly is crushed now.

DennisLfromGA:
@greymouser,

I agree and am happy you found a way around this.

ARC++ is still a work in progress so maybe things will change (again) before it hits stable.

I don't know all the ramifications of SELinux either but a [renowned Google developer commented](https://github.com/dnschneid/crouton/issues/2706#issuecomment-237737686) that it might be an SELinux interference (he hit the nail on the head) and he'd investigate - fingers crossed.

-----

I'm also looking forward to some file system integration between Chrome OS and Android, right now, [you have to do things the hard way](https://productforums.google.com/d/msg/chromebook-central/T7EgfwQAibo/rqCh8DLLCAAJ).

-**DennisL**

greymouser:
@DennisLfromGA ARC has some SELinux misconfiguration I discovered, but it's innocuous on the host side. On the chroot side, the misconfiguration (frankly, the configuration of SELinux in general) is tricky. Regardless, the issue was fully on the crouton side in the setup script (i.e. the part running on the host as root before the chroot is fully bootstrapped and entered); remounting the selinuxfs mount point (after stringing together a shared -> slave set of mount points) effectively kills ARC on the host.

This isn't just "a way around this", honestly -- this is the fix. If someone can demonstrate how their current host<->chroot bridged SELinux setup in their chroot is broken by the fix, then it would need to be revisited. I suspect the number of people that effects is zero (I have a hard time imagining a practical use case for SELinux in a chroot).

DennisLfromGA:
@greymouser,

I saw your [other post](https://github.com/dnschneid/crouton/issues/2706#issuecomment-241061916) about the dbus errors and it looks like you've done a pretty thorough investigation of the problems. I'm just an interested novice and have no knowledge of the intricacies of SELinux & ARC++ on Chrome OS so please regard my posts as only curious observations.

----
As far as your PR, I'd say LGTM but that doesn't count for much I'm adraid. ;-)
Guess we'll just have to wait for the devs to take a look.

-**DennisL**

dnschneid:
This indeed seems quite reasonable.  Could you please sign the CLA so I can merge this and get it rolled out posthaste?

host-bin/enter-chroot
```diff
@@ -528,10 +528,10 @@ if ! mountpoint -q "$CHROOT/sys"; then
     mount --make-rshared /sys
     mount --rbind /sys "$CHROOT/sys"
     mount --make-rslave "$CHROOT/sys"
-    # Remount selinux as ro
-    if [ -d "$CHROOT/sys/fs/selinux" ]; then
```
dnschneid: Actually, you still probably want to check for the existence of selinux before unmounting it.

greymouser:
@dnschneid I've updated the umount selinux section to check existence of the dir first, as well as silence other warnings (e.g. if the directory exists, but isn't a mount point, we really don't care).

I am now looking for the CLA to sign now ...

greymouser:
@dnschneid CLA signed! If there's anything else I need to do, please let me know.

Do you think I should add myself to the CONTRIBUTORS file?

host-bin/enter-chroot
```diff
@@ -528,9 +528,11 @@ if ! mountpoint -q "$CHROOT/sys"; then
     mount --make-rshared /sys
     mount --rbind /sys "$CHROOT/sys"
     mount --make-rslave "$CHROOT/sys"
-    # Remount selinux as ro
+    # Mounting selinuxfs, remounted as RO or not, causes
+    # ARC / "Google Play for Chrome OS" to crash. The break
+    # is permanent until a clean reboot.
     if [ -d "$CHROOT/sys/fs/selinux" ]; then
-        mount -o remount,ro "$CHROOT/sys/fs/selinux"
+    	umount "$CHROOT/sys/fs/selinux" 2>/dev/null
```
dnschneid: Your 2>/dev/null brings up a good point; there's no guarantee this is a mountpoint.  Perhaps changing the -d check above into a `mountpoint` call would be safer?

(also, spaces, not tabs, please)

dnschneid:
No need to add yourself to CONTRIBUTORS; it gets autogenerated when I release a new version.

DennisLfromGA:
@greymouser & @dnschneid,

My non-Android Chromebook 'crouton' updates are failing now so I guess `/sys/fs/selinux` was added to the latest Chrome OS updates, or it was before and I didn't notice it or hadn't updated in a while.
```
Version 54.0.2837.0 dev (64-bit)
Platform 8737.1.0 (Official Build) dev-channel lulu
Firmware Google_Lulu.6301.136.57
```
I had to extract crouton and modify `enter-chroot` in order to get past the errors:
```
+ [ -d /run/crouton/var/crouton/chroots/trusty/sys/fs/selinux ]
+ mount -o remount,ro /run/crouton/var/crouton/chroots/trusty/sys/fs/selinux
mount: /run/crouton/var/crouton/chroots/trusty/sys/fs/selinux not mounted or bad option

       In some cases useful info is found in syslog - try
       dmesg | tail or so.
```

I made similar changes to @greymouser only I used `mountpoint` as @dnschneid suggested:
```
    # Unmount selinux if mounted
    if mountpoint -q "$CHROOT/sys/fs/selinux"; then
        umount "$CHROOT/sys/fs/selinux"
    fi
```

-**DennisL**

DennisLfromGA:
**FYI:** @dreveman's *'bind'* mount option in PR https://github.com/dnschneid/crouton/pull/2748 seems to work well without unmounting selinux.

greymouser:
I was away for a week; catching up now, and checking out the other possible fix branch, too.

dnschneid:
Thanks for all your testing, @greymouser!

DennisLfromGA:
@greymouser,

And I guess you were wise to ponder the 'bind' option early on too -
> "Knowing the full ramifications of not bind-mounting the selinux mount in the changeroot (why it was there to begin with) would put me at ease, for sure, but ... well, I'm happy enough that this bug seemingly is crushed now."

-**DennisL**

Herran:
@dnschneid did you put the newest version on the regular download link?? the https://goo.gl/fd3zc or is it automatically updated. I want to have the newest version.

Herran:
@dnschneid omg!! It actually ran past the "Installing crash_reporter_wrapper into the host..." crossing my fingers...

dnschneid:
@herran I guess you discovered that the normal download link was the way to go :)

Herran:
@dnschneid  yeah. thanks for fixing this. I was so scared because I had a python class starting tomorrow. And I only had a chromebook. thanks!!

Herran:
@dnschneid nevermind. failed again : 
Installing crash_reporter_wrapper into the host...
dpkg: dependency problems prevent configuration of libc6-dev:i386:
 libc6-dev:i386 depends on linux-libc-dev.
dpkg: error processing libc6-dev:i386 (--configure):
 dependency problems - leaving unconfigured
Errors were encountered while processing:
 libc6-dev:i386
Unmounting /mnt/stateful_partition/crouton/chroots/precise...
chronos@localhost / $

dnschneid:
That's a totally different issue; you'll need to skip the update, go into your chroot and fix packages.

Herran:
ok.

Herran:
@dnschneid is there an error page for that error too? `sudo help`

>

Herran:
got it fixed and installed. Thanks!!

Issue #2655, opened 2016-06-26
Creating hwaudio group with GID 18... Password:  su: Authentication failure


a1exus:
I'm using `ASUS Chromebook Flip` with `Google Chrome OS 53.0.2773.3.dev` and while running following command:
```
chronos@localhost / $ sudo edit-chroot -all
name: kali-rolling
encrypted: no
Entering /mnt/stateful_partition/crouton/chroots/kali-rolling...
crouton: version 1-20160620130246~master:c39102d9
release: kali-rolling
architecture: armhf
xmethod: xorg
targets: xfce
host: version 8481.2.0 (Official Build) dev-channel veyron_minnie 
kernel: Linux localhost 3.14.0 #1 SMP PREEMPT Fri Jun 24 00:01:48 PDT 2016 armv7l GNU/Linux
freon: yes
Unmounting /mnt/stateful_partition/crouton/chroots/kali-rolling...
name: precise
encrypted: no
Entering /mnt/stateful_partition/crouton/chroots/precise...
Creating hwaudio group with GID 18...
Password: 
su: Authentication failurePassword: 
su: Authentication failure
chronos@localhost / $ 
```
I'm being asked for _Password_
```
Creating hwaudio group with GID 18...
Password: 
su: Authentication failure
```
Please advise.

rudedog830:
I am having the same issue.  Not sure what hardware the OP is using, but I'm on an Asus Flip running ChromeOS dev v53.

drinkcat:
Please update your chroot, that should fix it.

a1exus:
@drinkcat, I'm getting same thing with latest version of crouton:
```
chronos@localhost ~ $ sh ~/Downloads/crouton -V
Downloading latest crouton installer...
######################################################################## 100.0%
crouton: version 1-20160627183059~master:e3c0a690
chronos@localhost ~ $ sh ~/Downloads/crouton -V
crouton: version 1-20160627183059~master:e3c0a690
chronos@localhost ~ $ sudo edit-chroot -all
Password: 
name: precise
encrypted: no
Entering /mnt/stateful_partition/crouton/chroots/precise...
Creating hwaudio group with GID 18...
Password: 
su: Authentication failure
chronos@localhost ~ $ 
```

DennisLfromGA:
[...]

PR #2748, opened 2016-08-29
Avoid disabling SELinux for rest of system

dreveman:
Bind mount the directory read-only in the chroot.

### DIFF
--- a/host-bin/enter-chroot
+++ b/host-bin/enter-chroot
@@ -530,7 +530,7 @@ if ! mountpoint -q "$CHROOT/sys"; then
     mount --make-rslave "$CHROOT/sys"
     # Remount selinux as ro
     if [ -d "$CHROOT/sys/fs/selinux" ]; then
-        mount -o remount,ro "$CHROOT/sys/fs/selinux"
+        mount -o remount,ro,bind "$CHROOT/sys/fs/selinux"
     fi
 fi
 


DennisLfromGA:
`enter-chroot` is now working again for me and without having to unmount selinix.

Note: I do not yet have Android apps on my Dell CB13 so I cannot attest to it working with Google Play - yet.

dnschneid:
Would this be preferred over #2735?

greymouser:
@dnschneid From some simple manual tests, this fix also seems to work. I think this fix is to be preferred over my suggested changes in #2735 -- it's less changes overall. We should definitely get one of these out ASAP. :-)

@dreveman You may want to also change the comment above the conditional you modified to the title of this issue ("Avoid disabling SELinux for rest of system"), instead of what it is now.

dnschneid:
Sounds like this is the way to go then.  Thanks all for testing!

rickyz:
Mounting selinuxfs read only in the chroot causes issues with certain distros, see https://github.com/dnschneid/crouton/issues/2655. The correct solution is in fact to unmount selinuxfs inside the chroot, as done in https://github.com/dnschneid/crouton/pull/2735.

ohan2:
I'm terribly sorry for making this entirely unrelated inquiry, but I've done some research and asked the question elsewhere and I've yet to find a conclusive answer. So I figured that someone here at least would have the relevant expertise to finally address this concern of mine.

Has there been and could there be a firmware-level malware/rootkit on a Chromebook which would survive power-washing and compromise the system even while running stock? 

I know very little about how the firmware operates, but this page (https://www.chromium.org/chromium-os/chromiumos-design-docs/verified-boot) explicitly states that the Chromebook's verified boot solution relies on the assumption that the firmware hasn't been tampered with, and there have been reports of UEFI rootkits on PCs, so something of a similar nature isn't out of the question for Chromebooks. My Chromebook is my primary computer device, and due to this and that I've become pretty wary over the years.

Sorry...

divx118:
@ohan2 AFAIK there has never been any firmware-level malware/rootkit on a chromebook. However this is not a 100% insurance that there never will be. There will always security holes to be fixed on a system. Also if you have put your chromebook in developer mode to also use crouton, This is a good read https://github.com/dnschneid/crouton/wiki/Security
That said I feel pretty safe in using my chromebook in developer mode and never had any security issues.
IMO chromebooks are very safe to use.
I think you never find any conclusive answer, because there isn't one. There is on every system always a opportunity of braking through the security. How hard it will be, depends also on how much time/resources people want to spend breaking through the security and if they have physical access  or not.

ohan2:
@divx118 Thanks for the answer. That's a big relief since I bought my Chromebook second-handedly, and I conduct all of my personal and financial matters through it.

dnschneid:
For the ultimate paranoid: you can guarantee that your Chromebook is only running Google-supplied code if you:

1. Never disassemble your system (nor let anybody else disassemble your system), ensuring the write protect screw stays in place
2. Shut down the system and then hit Power+Refresh to restart the chain of trust after using crouton
3. At each "scary boot screen", hit TAB to confirm that verified boot is enabled
4. Do your banking/whatever prior to entering a chroot

Of course, this assumes someone hasn't managed to steal the Google signing keys, which would be a Bad Thing for everyone.  It also assumes someone hasn't tampered with the Chromebook at the factory, installing their own keys and managing a shadow update server that re-signs all the official updates unchanged until that one day when they decide they want to steal @ohan2's checking account.

ohan2:
@dnschneid I'm not too worried about that myself, since I'm only concerned is about the hypothetical small-time identity thieves or hackers, since I wasn't sure if some hacked firmware for Chromebooks had been released into the wild.

But if someone actually manages to steal the signing keys... yeah, I'll be reassessing my relationship with Google.

Issue #1567, opened 2015-03-31
'screencast' using host-x11 no longer works


DennisLfromGA:
My HP Chromebook 14 just updated to using Freon and my 'screencast' script is now broken - and it won't get up. I get the error:

  **No Chromium OS X server is available.**

The pertinent command is:
```
sudo enter-chroot -c /usr/local/chroots -n screencast host-x11 avconv -f alsa -ac 2 -ab 128k -i pulse -f x11grab -r 30 -s 1366x768 -i :0.0 -vcodec libx264 -preset ultrafast -crf 0 Downloads/movie1.mkv
```

It's using the `chroot-bin/host-x11` script and the error stems from this logic, I believe:
```
if [ ! -f "/sys/class/tty/tty0/active" ]; then
    err="No Chromium OS X server is available."
```

I tried adding the 'xorg' xmethod to no avail.

Shall I try to add xmethod 'xiwi' to see if that fixes it?
Is there another, better way to run this or am I now out of luck?

[croutonversion]
```
crouton: version 1-20150331105333~master:589ee11d
release: precise
architecture: amd64
xmethod: xorg
targets: xorg,audio
host: version 6812.52.0 (Official Build) beta-channel falco 
kernel: Linux localhost 3.8.11 #1 SMP Tue Mar 24 20:56:57 PDT 2015 x86_64 x86_64 x86_64 GNU/Linux
freon: yes
```

dnschneid:
There's no X server in Chromium OS with freon, so abusing host-x11 won't work anymore.  You'll either need to make avconv directly access the display data (not sure if possible), or use some Chromium OS native mechanism for screencasting.

DennisLfromGA:
Well those *options* are out of my wheelhouse so I guess 'screencast' is dead :(
Maybe Francois Beaufort will have some ideas, I may try to run it by him.

theblacklion:
@DennisLfromGA Anything new, yet?

DennisLfromGA:
@theblacklion - Nope, I'm afraid not - as yet...
I tried a few things with the 'xiwi' app but haven't yet found anything that works.
The combination of options possible with '[avconv](http://manpages.ubuntu.com/manpages/precise/man1/avconv.1.html)' are extensive, to say the least, so there are many more to try. I'm just not sure I'll have the patience or knowledge to fix it. ;(

VenomousSteam81:
[...]

PR #2749, opened 2016-08-29
Bind mount /var/run/chrome

dreveman:


dnschneid:
Will this interfere at all with Chrome running inside the chroot?

dreveman:
A chromeos build of chrome inside the chroot would interfere with this if XDG_RUNTIME_DIR is set to /var/run/chrome. That's to be expected though. XDG_RUNTIME_DIR is not set by default in the chroot. It can be set to /var/run/chrome if clients should be using the host instance of chrome as wayland server or it can be set to something else if that's no desirable. This change doesn't force any specific behavior. It just makes it possible to set it in a way that allows the host instance to used as display server for wayland clients.

dnschneid:
Oh, of course, duh :)
Thanks!

dnschneid:
Now we just need an equivalent to the [old host-x11 script](https://github.com/dnschneid/crouton/blob/58b76a38ce4125373da8dc4023a9ab9570269482/chroot-bin/host-x11)...

DennisLfromGA:
Yup, so I can again [abuse the `host-x11` script](https://github.com/dnschneid/crouton/issues/1567#issuecomment-88235073 ) or it's equivalent with 'screencast'. ;-)

PR #2757, opened 2016-09-01
Add host-wayland

dreveman:


DennisLfromGA:
@dreveman,

**NOTE:** I updated the chroot to version to f7b5bad which included the `enter-chroot` changes and I added your `chroot-bin/host-wayland` and the changes in `targets/core` prior to my tests.

----
I'm not sure `/var/run/chrome/wayland-0` is accessible in the chroot -
> (trusty)denny@localhost:~$ ls -l /var/run/chrome/wayland-0
ls: cannot access /var/run/chrome/wayland-0: No such file or directory

And I'm not sure `/var/run/chrome` is getting bind-mounted as `/var/host/chrome` either.

I get the error displayed -
> (trusty)denny@localhost:~$ ls -l /var/host/chrome
ls: cannot access /var/host/chrome: No such file or directory

I tried changing the `host-wayland` script to check for `/var/host/chrome/wayland-0` too but that didn't work.

> chronos@localhost ~/Downloads/crouton.unbundled $ enter-chroot -n trusty host-wayland 
Entering /var/crouton/chroots/trusty...
echo 'No Chromium OS Wayland server is available.' 1>&2
Unmounting /var/crouton/chroots/trusty...
Sending SIGTERM to processes under /var/crouton/chroots/trusty...

The 'err' message gets displayed with the entire line too...

I'm probably doing something wrong but I can't seem to get it to work.

-**DennisL**

dnschneid:
@DennisLfromGA does /var/run/chrome/wayland-0 exist outside the chroot?

DennisLfromGA:
Yes. Sorry, I meant to mention that also:
> chronos@localhost ~ $ ls -l /var/run/chrome/wayland-0
srw-rw---- 1 chronos chronos 0 Sep  1 20:49 /var/run/chrome/wayland-0

DennisLfromGA:
I'm currently on:
> Version 53.0.2785.81 beta (64-bit)
Platform 8530.69.0 (Official Build) beta-channel lulu
Firmware Google_Lulu.6301.136.57

Awaiting M53 stable - due out next week prolly...

dreveman:
You're likely missing the necessary permissions on the socket. It should have been:
srw-rw---- 1 chronos wayland 0 Sep 1 20:49 /var/run/chrome/wayland-0
and I'm not sure why the socket is not in the 'wayland' group in your case. Does it work if you use 'chronos' as username in the chroot? The idea is to add all users who should have access to this socket to the wayland group but I'm not sure how to best accomplish this in the crouton case. I guess it just happen to work for me as I've been using 'chronos' as chroot username..

DennisLfromGA:
@dreveman,

Here are the 'groups' for user 'chronos' on my system:
> chronos@localhost ~ $ groups
audio video pkcs11 input brltty lpadmin devbroker-access cras chronos chronos-access

Also, there is no 'wayland' group name in `/etc/group`.

I have not tried the username 'chronos' in any of my chroots but I can't see how it would matter in my case without a 'wayland' group name.

----
This does not explain why `/var/run/chrome` is not getting bind-mounted as `/var/host/chrome` either, which is a requirement for this to work I believe.

-**DennisL**

dreveman:
Are you using a Chromebook with Play Store support? If not, then that's likely what's causing your system to be different.

DennisLfromGA:
No I'm not. Sorry, thought I'd mentioned that.

dreveman:
Ok, that's a requirement for host-wayland to work today but I'll look into making some ChromiumOS changes that would allow this to work on all Chromebooks.

DennisLfromGA:
I thought Freon/Wayland was introduced around M50 and not dependent on Google Play???

dreveman:
You might have it confused with wayland client support. Wayland server support is new and it doesn't really depend on Play. It just happens that Play uses it so that's why it works only on devices with Play enabled. There's no good reason for this limitation so I'll try to change this asap.

Manually adding the 'wayland' group should work for now and give the socket the right permissions:
https://cs.chromium.org/chromium/src/components/exo/wayland/server.cc?rcl=1472733616&l=3096

DennisLfromGA:
Okay, sorry to confuse and complicate things.
I'll hold off testing / using host-wayland until my Dell CB13 gets Googe Play.

DennisLfromGA:
Well I guess my stumbling in the dark did produce some positive results after all.

I can delete all of my comments above if you want to keep this PR pristine - just let me know.

dnschneid:
@dreveman the chroot and the host are 1:1 for file gids, but do not necessarily have the same group associated with that gid.  If you're lucky and the /etc/group inside and outside has the same association between group name and group id, then it'll appear the same inside and out.  You can force the chroot to be lucky by adding to the [list of groups to synchronize](https://github.com/dnschneid/crouton/blob/master/host-bin/enter-chroot#L575).

Something to consider if we need to synchronize the wayland group.

dnschneid:
Oops, wrong reference :(

dnschneid:
Do we want croutoncycle to cycle through wayland windows like we did for host-x11?  I would say "no" because the Chromium OS window manager actually handles looping through the wayland windows properly.

Issue #214, opened 2013-06-11
Mouse moving doesn't work


cnicodeme:
Hi!

I just installed Unity via crouton and called `sudo startunity` which perfectly loaded me Ubuntu, but I can't move the mouse.

I can right click (the context menus shows), close it (via esc), but I can't move my mouse (I can't say if I can left clic).

I tried going back to ChromeOs and then regoing into Ubuntu, but the mouse still don't move.

For information, I use a HP Chromebook 14 with the original mouse (not a plugged USB one). With an external mouse (USB), it works greats!

Is there something I am missing ?

After some trying it appears the mouse is moving, but I have to press really hard on the touchpad to make it move, and the behavior is not great at all.

dereckson:
In December 2014, on an Asus C200, installing Unity on the top of a chroot with Ubuntu Trusty, all worked fine.

Recently, powerwashed, with the last Crouton version, I have a queer behavior with the touchpad, not affected by the synclient FingerLow=1 FingerHigh=5.

* No cursor move is achievable with one, two or three fingers
* Left and right click replies (but not tap)
* When I left or right click, and maintain this tap pressed, I can move the cursor with a correct sensitivity, but only if I use two fingers.
* Triple finger click seems to be recognized too.

davidthewatson:
@dnschneid I'm running on Acer Chromebook 14. I installed crouton from the link on the github repo today 5/6/2016 and installed xfce-desktop on trusty with every extension except keyboard. I had the same problem described here evidenced by the fact that the touchpad responded to clicks fine but would not move unless I I used my palm as if it was my finger. The touchpad worked fine using my palm as a finger, but that is obviously not an effective prosthetic! My problem was solved by:

    synclient FingerLow=1 FingerHigh=5

at which point the touchpad works perfectly. Wasn't sure if you wanted to re-open this bug and address it in crouton since the Acer Chromebook 2014 just started shipping from Amazon a few days ago.

CreepGin:
Just got the Acer CB14 and I can confirm the same problem described by @davidthewatson above.

JonathanNagy:
I have the same problem on a Acer CB3-431 (Aluminum 14"),though maybe less severe that some.  Touchpad Input responds intermittently, but touchpad hardware click functions work fine.  Fixed with "synclient FingerLow=1 FingerHigh=5".

haithcockce:
HP Chromebook 12 G1, 
The `synclient` command worked for me as well.

[...]

PR #2759, opened 2016-09-02
Acer Chromebook 14 (edgar) specific touchpad settings.

Happy-Ferret:
The following PR should fix the touchpad problems mentioned in issue #2657.

I've chosen to implement a custom path as I felt setting `FingerLow` and `FingerHigh` alone didn't offer a sufficient enough usability improvement.


dnschneid:
Wow, did you customize each one of those values?

Happy-Ferret:
I manually tested every one of them.
Why? Is that a problem?

drinkcat:
The question is, does it work reliably with just `FingerLow=5 FingerHigh=10`? Are all the other settings really necessary?

DennisLfromGA:
I was wondering if those settings (`FingerLow=5 FingerHigh=10`) could also be applied to some other models too, like I mentioned in this comment - https://github.com/dnschneid/crouton/issues/214#issuecomment-232519598

Some have already been applied but these may need the settings too -
- chell (HP Chromebook 13)
- terra (ASUS Chromebook C202SA)
- reks (Lenovo N22 Chromebook)

Issue #2771, opened 2016-09-14
Run Unity on Xenial


idvoretskyi:
I'm going to run Ubuntu Xenial with Unity, but due to existing bugs (there's no package `unity-2d` in Xenial),  the installation fails.

How can I run Ubuntu Xenial with Unity manually?

DennisLfromGA:
@idvoretskyi - Take a look at this similar issue for a possible solution: https://github.com/dnschneid/crouton/issues/2720#issuecomment-239538217

drinkcat:
Let's duplicate to #2720.

Issue #2720, opened 2016-08-12
/org Missing in Xenial (Unity)


msolters:
Hi, I've managed to get Unity working in Xenial.  I did not use the `unity` target, but rather, entered the chroot and manually `sudo apt-get install ubuntu-destop`.

Then, I can start a working Unity session using `sudo xinit unity`.  This requires the `x11` target.
## Problem

It works almost perfectly!  There's some weird glitches that I'm trying to iron out.  It seems the errors that occur mostly all mention the keyword `org`.

Researching, I found similar issues where users had reset using e.g.

```
sudo dconf reset -f /org/compiz/
```

But, it seems the issue boils down to the fact that the `/org` directory does not exist!

Now, I am not an expert on the window managers.  **What package would I be looking for that is responsible for generating this `/org` directory?**
## Things I Have Tried
-  I have tried adding the `xorg` target but it does not help.
-  I have tried reinstalling basically anything to do with Unity (`ubuntu-desktop`, `unity`, `compiz`, etc.)  Does not help.

DennisLfromGA:
@msolters,

I'm not sure why you added the unity desktop yourself but that may be what's causing the problem.
When crouton adds desktops targets, it does things specifically for the chroot environment running under Chrome OS.

I would recommend that you update your chroot via crouton with the unity-desktop target, something similar to the following should work:

`sudo sh ~/Downloads/crouton -t unity-desktop -u [-n xenial]`

This way, packages that are already installed and updated will be skipped but other packages that are needed will be added.

Hope this helps,
-**DennisL**

msolters:
Hey @DennisLfromGA, the issue is that part of the `unity` target is the `unity-2d` package.  This is unresolvable in 16.04.

The two solutions are

*  To remove `unity-2d` from `prepare.sh` mid-installation, or to remove it from the `targets` file (https://github.com/dnschneid/crouton/blob/58b76a38ce4125373da8dc4023a9ab9570269482/targets/unity#L20) and recompile crouton.  Both have the same result.
*  Do not use the `unity` target and manually install.

You get different outcomes for each.  If I use the `unity` target, with the workaround described above, Unity will only render a background and a weird X-shaped cursor.

If I install manually, I get the full UI (dash, menu, windows, etc.).  This is why I used the manual approach.

I will verify whether or not that `/org` folder is generated by using the target method.  I doubt it will be but we simply don't know yet.

DennisLfromGA:
@msolters,

[...]

PR #2814, opened 2016-10-10
Fixed unity target install errors on ubuntu xenial

willy-vvu:
unity-2d doesn't exist in xenial's repository, which had caused chroot installation errors.

Even though the install works now, unity does not automatically start with gnome-session, and several applications are broken upon starting unity manually.

At least it's a start.

Issues addressed but not fully fixed:
https://github.com/dnschneid/crouton/issues/2771
https://github.com/dnschneid/crouton/issues/2720

Personally, I would like to have support for 16.04, as it's an LTS release. Since trusty, much has changed, though crouton has yet to reflect those changes.

Unfortunately, I don't think my knowledge of Ubuntu's workings is enough for me to singlehandedly add support for an OS. If anyone is interested, please help!

### DIFF
--- a/targets/unity
+++ b/targets/unity
@@ -17,9 +17,15 @@ CHROOTETC='unity-autostart.desktop unity-profiled'
 . "${TARGETSDIR:="$PWD"}/common"
 
 ### Append to prepare.sh:
-install unity unity-2d ubuntu-artwork gnome-session nautilus \
-        ubuntu-settings,ubuntu~precise= \
-        ubuntu-session,ubuntu~precise+ubuntu~quantal+ubuntu~raring+ubuntu~saucy= \
+# Note that whitespace is important for syntax...
+# See distropkgs method in installer/prepare.sh
+install unity ubuntu-artwork gnome-session nautilus \
+        `# unity-2d only exists on ubuntu trusty and precise.`\
+        ubuntu~trusty+ubuntu~precise=unity-2d, \
+        `# ubuntu-settings does not exist on ubuntu precise.`\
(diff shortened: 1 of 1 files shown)


targets/unity
```diff
@@ -17,9 +17,15 @@ CHROOTETC='unity-autostart.desktop unity-profiled'
...
-        ubuntu-session,ubuntu~precise+ubuntu~quantal+ubuntu~raring+ubuntu~saucy= \
+# Note that whitespace is important for syntax...
+# See distropkgs method in installer/prepare.sh
+install unity ubuntu-artwork gnome-session nautilus \
+        `# unity-2d only exists on ubuntu trusty and precise.`\
+        ubuntu~trusty+ubuntu~precise=unity-2d, \
+        `# ubuntu-settings does not exist on ubuntu precise.`\
+        ubuntu~precise=,ubuntu-settings \
+        `# ubuntu-session does not exist on the following releases.`\
+      ubuntu~precise+ubuntu~quantal+ubuntu~raring+ubuntu~saucy=,ubuntu-session \
```
drinkcat: Please do not interleave code and comments.

Also, avoid comments such as `ubuntu-settings does not exist on ubuntu precise.` which basically repeat what the code does just below.

willy-vvu: To me (and future collaborators), the behavior of the code is not obvious. It's in a format exclusive to crouton, its behavior only apparent in a method which took me much digging to find. For now, I've removed these interleaved comments, but I hope the two comments above are enough to help anyone who reads the file in the future, by pointing them in the right direction.

dnschneid:
Thanks!

PR #2817, opened 2016-10-11
xenial: fix kde target

smibarber:
kde-plasma-desktop has gone away, so just install kde-baseapps and
plasma-desktop directly.

Also fixes sid.


telebuss:
Hi, I just installed xenial with kde, and it does finish installation and start, but when I open any program, I cannot drag it and I can't see the upper bar with the close/minimize buttons. Also, programs start on the upper left and cannot be resized or dragged (as stated before)
Any ideas on how to fix it?
Thanks a lot

PR #2818, opened 2016-10-11
Fix GNOME fallback/flashback on trusty+,wheezy+,kali+

smibarber:
Addresses feedback in #2429.

### DIFF
--- a/targets/xorg
+++ b/targets/xorg
@@ -205,8 +205,7 @@ else
 fi
 
 # Fix launching X11 from inside crosh (user doesn't own a TTY)
-sed -i 's/allowed_users=.*/allowed_users=anybody/' \
-    '/etc/X11/Xwrapper.config' || true
+echo 'allowed_users=anybody' > '/etc/X11/Xwrapper.config'
 
 inteldrv='/usr/lib/xorg/modules/drivers/intel_drv.so'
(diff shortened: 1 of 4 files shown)


targets/xorg
```diff
@@ -205,8 +205,7 @@ else
 fi
 
 # Fix launching X11 from inside crosh (user doesn't own a TTY)
-sed -i 's/allowed_users=.*/allowed_users=anybody/' \
-    '/etc/X11/Xwrapper.config' || true
+echo 'allowed_users=anybody' > '/etc/X11/Xwrapper.config'
```
drinkcat: Are we sure that `/etc/X11/Xwrapper.config` never contains anything else of interest?

If not, we might want to write a small awk script instead to modify/add the line as required.

dnschneid: The only other configuration items, it seems, are needs_root_rights and nice_value, which aren't even documented.  So no, there's nothing else.

dnschneid:
Seems to be getting `[e] Error executing command as another user: Not authorized` on x86 sid

dnschneid:
Tests seem to be passing for the most part...good job!

onesneakymofo:
Just want to thank you guys for releasing this update.  I bricked my installation about two weeks ago, and this fixed it.

PR #2820, opened 2016-10-12
Add a crouton group that can access lockfiles

### LINKED ISSUE #908: trusty X server display malfunctions when uid != 1000
I have been using Crouton with `trusty` and Openbox (via the `lxde` target) on an Asus Chromebox for a little over a month.  Yesterday I rebooted, which caused Chrome OS to upgrade itself.  Following upgrade, inside `trusty` the X server display started malfunctioning.

Specifically:  Initially, most of the screen is black, the exception being the LXPanel at the bottom.  The screen only refreshes due to user activity, and even then, only partially refreshes.  Consequently, the screen is usually displaying stale and/or corrupted windows.

Interestingly, the problem only manifests when running as a user other than `1000`.

Steps to reproduce the problem:

    # Create a trusty chroot named vanilla.
    $ sudo sh -e ~/Downloads/crouton -t lxde -n vanilla -r trusty
    $ sudo startlxde -n vanilla
    # X display functions properly.

    # Then create another user named u1100.
    $ sudo enter-chroot -n vanilla
    $ sudo groupadd -g 1100 g1100
    $ sudo useradd -g g1100 -m -s /bin/bash -u 1100 u1100
    $ exit

    # Enter LXDE as user u1100.
    $ sudo startlxde -n vanilla -u u1100
    # The X display malfunctions, as described above.

Ancillary issues:

1) In LXDE, the Logout option does nothing.  In `precise` it works, but not in `trusty`.  Logging out requires switching to Chrome OS, and pressing ^C in the terminal to kill X.  (Aside: this does not really affect me, as I typically use a custom `~/.xsession` file.)

2) During the creation of the `vanilla` chroot, my keyboard layout in Chrome OS was reset from Dvorak to Qwerty.  The only way to correct was to log out of Chrome OS.

3) Simultaneous with the keyboard layout change, the responsive of my trackball dropped to almost nothing.  As in: spin and spin and spin and spin to slowly, slowly move the mouse cursor across the screen.  At this point, the Chrome OS settings page said that no mouse or touchpad was detected.  Again, logging out of Chrome OS was required to fix the problem.

Chrome OS version info:
Version 35.0.1916.155
Platform 5712.88.0 (Official Build) stable-channel panther
Firmware Google_Panther.4920.24.22

crouton version info:
VERSION='1-20140421214646~master:b766b319'

An obvious workaround should be to only run Crouton as user 1000.

dnschneid:
Fixes #908 if you add the user to the right groups.

### DIFF
--- a/chroot-bin/croutonclip
+++ b/chroot-bin/croutonclip
@@ -118,6 +118,8 @@ current='cros'
 
 mkdir -m 775 -p "$CROUTONLOCKDIR"
 exec 3>>"$CROUTONLOCKDIR/clip"
+chmod -Rf g+rwX "$CROUTONLOCKDIR" || true
+chgrp -Rf crouton "$CROUTONLOCKDIR" || true
 if ! flock -n 3; then
     echo "Another instance of croutonclip running, waiting..."
     flock 3
@@ -136,7 +138,7 @@ addtrap "echo -n > '$CROUTONLOCKDIR/clip' 2>/dev/null"
 
     waitwebsocket
 
-    # Update on VT change (if user types Ctrl-Alt instead of Ctrl-Alt-Shift)	
+    # Update on VT change (if user types Ctrl-Alt instead of Ctrl-Alt-Shift)
     if hash croutonvtmonitor 2>/dev/null; then
        croutonvtmonitor &
        addtrap "kill $! 2>/dev/null"
--- a/chroot-bin/croutontriggerd
+++ b/chroot-bin/croutontriggerd
@@ -39,8 +39,8 @@ EVENT_DEV_POLL=15
 # Only one at a time. xbindkeys lockfile is for legacy compatibility
 mkdir -m 775 -p "$CROUTONLOCKDIR"
 exec 3>>"$CROUTONLOCKDIR/xbindkeys"
-chmod -Rf g+rwX '/tmp/crouton-lock' || true
-chgrp -Rf crouton '/tmp/crouton-lock' || true
+chmod -Rf g+rwX "$CROUTONLOCKDIR" || true
+chgrp -Rf crouton "$CROUTONLOCKDIR" || true
 if ! flock -n 3; then
     echo "Another instance of ${0##*/} running, waiting..."
     flock 3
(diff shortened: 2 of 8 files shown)


chroot-bin/croutontriggerd
```diff
@@ -39,6 +39,8 @@ EVENT_DEV_POLL=15
 # Only one at a time. xbindkeys lockfile is for legacy compatibility
 mkdir -m 775 -p "$CROUTONLOCKDIR"
 exec 3>>"$CROUTONLOCKDIR/xbindkeys"
+chmod -Rf g+rwX '/tmp/crouton-lock' || true
```
smibarber: `"$CROUTONLOCKDIR"`?

dnschneid: "$CROUTONLOCKDIR".

chroot-bin/croutonclip
```diff
@@ -118,6 +118,8 @@ current='cros'
 
 mkdir -m 775 -p "$CROUTONLOCKDIR"
 exec 3>>"$CROUTONLOCKDIR/clip"
+chmod -Rf g+rwX "$CROUTONLOCKDIR" || true
```
drinkcat: Huh, I did not know about `X`!

drinkcat: (And it's POSIX!)

host-bin/enter-chroot
```diff
@@ -575,8 +575,9 @@ fi
 # so that users have access to shared hardware, such as video and audio.
 gfile="$CHROOT/etc/group"
 if [ -f "$gfile" ]; then
-    for group in audio:hwaudio cras:audio cdrom devbroker-access disk floppy \
-                 i2c input lp serial tape tty usb:plugdev uucp video wayland; do
+    for group in audio:hwaudio cras:audio cdrom chronos-access:crouton \
```
drinkcat: Wait, why chronos-access?

chroot-bin/croutonclip
dnschneid: Yeah, that's my favorite chmod parameter.  It makes -R behave sanely.

drinkcat:
+1

PR #2829, opened 2016-10-14
x11-common: only do upstart-xsessions hack on trusty

smibarber:
Upstart was taking over X sessions on saucy+, but xenial
no longer needs this since it's joined the realm of systemd.


smibarber:
Actually... we need this on xenial as well. The question then is why does this only hit a dpkg conflict on xenial.

smibarber:
Will merge this in with the unity branch since this is a prereq for unity on xenial.

Issue #2113, opened 2015-09-16
croutonurlhandler should work without X server


dnschneid:
croutonurlhandler does not work without an X server, since croutonclip is not started, and does not fail with a clear message: it times out.

dnschneid:
From #2161: we should fix the pipe permissions and autorun the daemon as an unprivileged system user as part of the glorious enter-chroot script.

radix:
fwiw, it seems that the timeout bug is fixed? I can run croutonwebsocket from a CLI. but I suppose this should still be automated in enter-chroot so I don't need to start it myself.

PR #2852, opened 2016-10-26
install "www-browser" symlink as well

jessaustin:
In the terminal, as with e.g. targets *xiwi* and *cli-extra*, tools such as `xdg-open` can't find a `$DISPLAY`, so they attempt to execute `www-browser` rather than `x-www-browser`. So, that symlink should also exist.

### DIFF
--- a/targets/extension
+++ b/targets/extension
@@ -36,7 +36,7 @@ fi
 
 # Use croutonurlhandler as a fallback URL handler
 handler='/usr/local/bin/croutonurlhandler'
-for link in x-www-browser gnome-www-browser; do
+for link in x-www-browser gnome-www-browser www-browser; do
     if ! update-alternatives --query "$link" | grep -q "$handler"; then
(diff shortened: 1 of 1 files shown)


dnschneid:
Good catch.  Does croutonclip get properly launched, though, in these situations?  Still worth merging since if you install a console-based browser it will override this.

jessaustin:
I've just been running `croutonwebsocket` at the beginning of my session, which seems to be enough to get this working, although please do advise if it's not the recommended solution. I read #2113 with interest, because it seems plausible that something similar could be done behind the scenes... I haven't yet understood the referenced pipe perm issues. If one wanted to fix that issue, how would one go about it?

dnschneid:
Right, that should be sufficient.  Pipe permissions have already been fixed thanks to the merging of the multiuser branch, so if you'd like to take a stab at it, just make sure to copy how croutonwebsocket is launched elsewhere.

Issue #2850, opened 2016-10-25
Getting error installing crouton on a ChromiumOS VirtualBox


aghasemi:
```
name: precise
encrypted: no
Entering /mnt/stateful_partition/crouton/chroots/precise...
cp: cannot stat ‘/run/crouton/mnt/stateful_partition/crouton/chroots/precise/dev/dri’: No such file or directory

```

I get the following error when trying to install crouton on the un-official VirtualBox image of the Neverware Cloudaready ChromiumOS:

```
cp: cannot stat ‘/run/crouton/mnt/stateful_partition/crouton/chroots/precise/dev/dri’: No such file or directory
```

I tried trusty and precise for release, and combinations of xiwi,xfce, and core for target. All the same.

dnschneid:
Huh, neat.  Should be easy enough to fix--just need to add a directory check to https://github.com/dnschneid/crouton/blob/master/host-bin/enter-chroot#L554

aghasemi:
How?

DennisLfromGA:
@aghasemi,

Just a guess but maybe something like:

from: 
> `if [ ! -d "$varrundrm" -a -d "$sysclassdrm" ]; then`

to: 
> `if [ ! -d "$varrundrm" -a -d "$sysclassdrm" -a -d "$devdri" ]; then`

*Hope this helps,*
-**DennisL**

aghasemi:
It worked. Thanks. I proposed the required change in enter-chroot.

DennisLfromGA:
Cool, I'll close this now...

PR #2855, opened 2016-10-26
The proposed fix to bug #2850

aghasemi:
Hi,

As Dennis suggested, I made small change to line 554 of enter-chroot.sh that solved bug #2850. By the way, it was line 554 in the file that was downloaded to my computer.
Thanks
Best


aghasemi:
I signed it!

dnschneid:
Thanks!

Issue #1638, opened 2015-04-19
cannot create regular file ‘/usr/local/chroots/trusty/.crouton-targets’: No such file or directory


PickleJuice:
ChromeOS trying to install Crouton and it stops with the following:

cp: cannot create regular file ‘/usr/local/chroots/trusty/.crouton-targets’: No such file or directory

Any ideas how to proceed?

dnschneid:
What command are you running?  Could you please post the full logs?

DennisLfromGA:
Closing. Please re-open if this issue is still a concern.

aghasemi:
Can you re-open this please? I got the same error on CloudReady ChromiumOS 52:
Installation command `sudo sh crouton -r trusty -t xiwi,core`

It ended with

```
Installing delete-chroot into the host...
Installing edit-chroot into the host...
Installing mount-chroot into the host...
Installing unmount-chroot into the host...
Installing crash_reporter_wrapper into the host...
cp: cannot create regular file ‘/usr/local/chroots/trusty/.crouton-targets’: No such file or directory
```

Any idea?

DennisLfromGA:
Re-opened, please provide some more details if you can -
- Is this the first time you've used Cloudready and you ran into this problem?
- Have you run into this problem on any other versions or using Chrome OS?

Or anything else that you can think of that might be relevant.

aghasemi:
Thanks.

- Yes. A standalone installation of CloudReady, with crouton tried right after installation
- No idea about ChromeOS. I've tried the unofficial VirtualBox image as well, where I received another error, issue #2850 .

DennisLfromGA:
**_@crouton-devs_**

It's evident that I don't understand how all this works but it looks to me like the condition statement on [main.sh#L688](https://github.com/dnschneid/crouton/blob/master/installer/main.sh#L688) only executes if we are *restoring* the host-bin files. 

And [main.sh#L726](https://github.com/dnschneid/crouton/blob/master/installer/main.sh#L726) doesn't check to see if `.crouton-targets` exists first.

*Curious,*
-**DennisL**

aghasemi:
I re-downloaded the script now and it successfully installed xenial. Unless you've fixed it yourself, it's not a good sign :-/

DennisLfromGA:
The crouton installer script should get a fresh copy of crouton when it's run so I don't think re-downloading it was the solution, unless yours was very old.

[...]

PR #2856, opened 2016-10-27
A simple non-intrusive solution to bug #1638

aghasemi:
In line 727, we touch "$CHROOTSRC/.crouton-targets" to make sure it exists before using it in the next line.

### DIFF
--- a/installer/main.sh
+++ b/installer/main.sh
@@ -723,6 +723,8 @@ done
 
 if [ -f "$PREPARE" ]; then
     # Update .crouton-targets in the unencrypted part of the chroot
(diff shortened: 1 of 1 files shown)


dnschneid:
Sorry, I don't see what this does.  Touching the file is unnecessary since the cp immediately overwrites it, whether the file exists or not.

aghasemi:
Aha Ok. So we should find another cause for #1638, after reproducing it...

dnschneid:
Yeah, we should definitely be able to consistently repro the issue before deciding whether a change fixes it or not :)

PR #2859, opened 2016-10-28
Updated README with download instruction from shell

idvoretskyi:
Sometimes it's easier to install crouton directly from the command line. These instructions will allow you to do that easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,17 @@ If you're just here to use crouton, you can grab the latest release from
 (Ctrl+Alt+T, type `shell` and hit enter), and run `sh ~/Downloads/crouton` to
 see the help text. See the "examples" section for some usage examples.
 
+Also, you can download it directly using shell (Ctrl+Alt+T, type `shell` and hit enter) with the following commands:
+
+```
+# verify if /usr/local/bin/ directory exists; it's created atomatically after running `dev_install` command, otherwise you may create it manually:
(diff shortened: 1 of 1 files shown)


DennisLfromGA:
@idvoretskyi,

I *definitely* think it's best to download 'crouton', either by clicking the link or manually via wget, etc., in the _**~/Downloads**_ folder and then launching it using: `sudo sh ~/Downloads/crouton ...`

For one, it's just simpler and doesn't add unnecessary steps. 
And two, the `dev_install` command is *dangerous* since it will wipe out anything under `/usr/local/`, which will remove the `chroots` and `bin` folders if they are present. 

However, that's just my opinion.

*Hope this helps,*
-**DennisL**

dnschneid:
Yeah, I don't think this really adds anything to the README, and poking at dev_install is indeed dangerous.

idvoretskyi:
@DennisLfromGA @dnschneid well, I would suggest having the "CLI-friendly" way to download Crouton in README. Anyway, I can adopt it and skip the "dev_install" section.

dnschneid:
Thanks, but it's not really clear to me why you'd need a CLI-friendly way to download crouton in the README, since you basically have to be in a web browser in order to read it.  Wouldn't it be more work to copy-paste whatever commands are there?

Seems like something you could put in the wiki, but not that useful in the README.

PR #2868, opened 2016-11-04
Small fix for edgar touchpad support

elefthei:
Edgar is a new chromebook model, this fixes low responsiveness on the synaptic touchpad.
I think all newer models have higher touchpad resolution so maybe a better solution would be to send the "default" case to FingerLow=1 FingerHigh=5.


DennisLfromGA:
And perhaps the devices mentioned [here](https://github.com/dnschneid/crouton/pull/2759#issuecomment-253617360) could be added also:
- chell (HP Chromebook 13)
- terra (ASUS Chromebook C202SA)
- reks (Lenovo N22 Chromebook)

elefthei:
Done at 1ca0397, not tested though. If somebody has these machines available, it would be great if they could try this PR.

elefthei:
Tested on chell and terra and I'm very confident this will work on reks. As well as newer models after that. Let's merge this soon since a lot of people are having touchpad problems?

xxcriticxx:
when you going to merge this?

PR #2874, opened 2016-11-08
Allow aarch64 chroots

ericvh:
With introduction of acer R13 chromebook, we now have 64-bit capable hardware for ARM.  Let's use it.

Signed-off-by: Eric Van Hensbergen <ericvh@gmail.com>

### DIFF
--- a/installer/ubuntu/defaults
+++ b/installer/ubuntu/defaults
@@ -13,7 +13,8 @@ fi
 case "$ARCH" in
 x86 | i?86) ARCH="i386";;
 x86_64 | amd64) ARCH="amd64";;
-arm* | aarch64) ARCH="armhf";;
+aarch64) ARCH="aarch64";;
+arm*) ARCH="armhf";;
 *) error 2 "Invalid architecture '$ARCH'.";;
 esac
 


ericvh:
I signed it.

dnschneid:
I was originally going to do this, but discovered that graphics seemed to be quite broken with aarch64.  It's possible my system was just the broken one.  Have you had better success?

If you find it's not quite right, then I'd suggest making it so the default (line 10) selects armhf on aarch64 systems, but otherwise keeps your patch as-is so that you can specify it explicitly.

ericvh:
Tested with core chroot only at the moment, going for graphics now.  I
guess if graphics are working, I need to fix up to recast aarch64 as arm64
anyways.  Will update appropriately based on results.  Thanks.

        -eric

ericvh:
FWIW - lxde seems to work fine with xenial - touchpad is a bit flakey (but
touchscreen works).  I'm gonna try xorg and xiwi next.

livejamie:
I have an R13, I'd love to help troubleshoot/test.

DennisLfromGA:
@livejamie,

Often the way I test some of these PRs is to unbundle the crouton master in my `~/Downloads` folder using crouton's '-x' option -

`cd ~/Downloads`
`sudo sh ./crouton -x`
`cd crouton.unbundled`

Here you'll find the hierarchy of crouton files & folders to work with.

You can modify `./installer/ubuntu/defaults` with the changes show here: 
https://github.com/dnschneid/crouton/pull/2874/files

Then install and/or update your chroot using `main.sh` instead of `crouton` similar to:  

`sudo sh ./installer/main.sh -r xenial -t xfce`

That's just an example so feel free to use any chroot and targets you choose.

*Hope this helps,*
-**DennisL**

dnschneid:
I'd like to first merge something that allows but doesn't default to aarch64 so that people can play with it.

alk:
I recently got Samsung Chromebook Plus and I was massively confused trying to install arm64 debian sid and crouton giving me armhf. Google search brought me here.

I tried suggestion above (unpacking installer and hacking arm64 "support" into it). I've got seemingly working debian sid arm64 (at least I compiled gperftools with no apparent issues). So it seems to work.

Further, I think it is ok to default to armhf if we're not confident about arm64 quality. But if is user explicitly asking for arm64 and gets armhf, that is arguably a bug.

mbgg:
I was able to get trusty xfce working on the Acer Chromebook R 13, but it uses armhf and not aarch64.

Apart from that I'm working on enabling openSUSE aarch64 port on crouton (not yet finished). So I propose to coordinate on any changes to enable arm64 distributions on crouton.

bitbank2:
Any chance of allowing trusty xfce to run aarch64 userland on the R13? I honestly bought the R13 so I could run aarch64 Linux (I also use crouton on my Asus Flip for 32-bit armhf).

dnschneid:
@bitbank2 until we get something merged, you can always extract the crouton installer (`sh crouton -x`), cd into crouton.unbundled, edit the installer/ubuntu/defaults file to force arm64 as the release, then run installer/main.sh as if it were the crouton installer.

bitbank2:
Thanks for the info. Tried to install trusty/xfce as arm64 and got the following errors:
The following packages have unmet dependencies:
xfce4-goodies : Depends: xfce4-battery-plugin but it is not installable
      recommends: xfce4-power-manager but it is not going to be installed
E: unable to correct problems, you have held broken packages.

PR #2896, opened 2016-11-20
Add 17.04 as a test release

StrawnPoint04:
Even though it's still pre-alpha, there are some who might already want to use


dnschneid:
Thanks!

Issue #2878, opened 2016-11-11
cursor issue after switch


giliammc:
i have an acer c910 core i3 4gb of ram and a 128gb ssd the output for sudo edit-chroot -all is
```
name: xenial
encrypted: no
Entering /mnt/stateful_partition/crouton/chroots/xenial...
crouton: version 1-20161028111244~master:948ee6dc
release: xenial
architecture: amd64
xmethod: xiwi
targets: xiwi,xorg,xfce,extension,audio
host: version 8743.76.0 (Official Build) stable-channel auron_yuna 
kernel: Linux localhost 3.14.0 #1 SMP PREEMPT Tue Oct 25 23:34:16 PDT 2016 x86_64 x86_64 x86_64 GNU/Linux
freon: yes
Not unmounting /mnt/stateful_partition/crouton/chroots/xenial as another instance is using it.
```

after switching back into my chroot my cursor turns into this 
![image](https://cloud.githubusercontent.com/assets/10850574/20224073/4ae703c4-a802-11e6-924f-704430a4530e.png) (sorry for the picture quality if i took a screenshot the cursor shows up as normal) if i change the resolution from 1920x1080 to something else and then back again this resets the cursor but if i switch to chrome and back again it reapers. thanks in advance for any help crouton is really great and the support community is one of the best.

sunnyps:
I was able to reproduce this on my Dell Chromebook 13 (i3/8G) on trusty with unity and sid with gnome. I use a scaled resolution for ChromeOS (1536 x 864) and the chroot uses the default resolution (1920 x 1080). I thought it was related to GL accelerated desktops but @giliammc was able to reproduce this on XFCE which is not hardware accelerated by default.

drinkcat:
Just to clarify, are you running crouton in a window (method `xiwi`?), or natively using method `xorg`?

drinkcat:
@sunnyps : Can I have your output of `edit-chroot -all` as well?

giliammc:
i was using xorg

dnschneid:
This is crbug.com/482801 ...I thought there was a corresponding bug on the crouton tracker as well, but I can't find it at the moment.

dnschneid:
Okay, sounds like this is going to need a fix on the crouton side.  Having chatted with the kernel graphics people, it sounds like the issue is that Chromium OS uses planes to render the cursor, while X11 uses the old cursor interface.

[...]

PR #2905, opened 2016-11-24
croutonfreon.so: Disable cursor on switch away from Chromium OS

drinkcat:
For reasons that I do not fully comprehend, cursor sometimes get corrupted in crouton (see #2878, http://crbug.com/482801).

It does appears the resetting cursors on all CRTCs fixes the issue, so let's do that.

### DIFF
--- a/src/freon.c
+++ b/src/freon.c
@@ -20,6 +20,8 @@
 #include <unistd.h>
 #include <linux/input.h>
 #include <linux/vt.h>
+#include <xf86drm.h>
+#include <xf86drmMode.h>
 
 #define LOCK_FILE_DIR "/tmp/crouton-lock"
 #define DISPLAY_LOCK_FILE LOCK_FILE_DIR "/display"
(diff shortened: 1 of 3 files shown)


drinkcat:
I'm not going to pretend I understand what's going on (and I don't really have the cycles to figure it out ,-( ). But, this seems to fix the issue.

dnschneid:
Thanks for the black magic!  Does libdrm-dev need to be installed for compilation?

drinkcat:
2016-11-24_07-53-46_drinkcat_chroagh_freon_cursor/status.html ... Turns out libdrm header files appear to be broken on most releases... Looking.

src/freon.c
```diff
@@ -95,6 +97,38 @@ static int set_display_lock(unsigned int pid) {
...
+    drmModeRes* resources;
+
+    fd = open("/dev/dri/card0", O_RDWR, 0);
+
+    TRACE("%s %d\n", __func__, fd);
+
+    if (fd < 0)
+        return;
+
+    resources = drmModeGetResources(fd);
```
drinkcat: Done

drinkcat:
Manually tested on precise/trusty/xenial and stretch (all on samus). I'll launch a test as well.

drinkcat:
Mostly pass `2016-11-25_02-51-07_drinkcat_chroagh_freon_cursor_35/status.html`. `sid` is broken, but I don't think that's related. Also, x86-alex is weird, not sure what's going on but does not seem related either.

PR #2963, opened 2016-12-20
Create Android Studio target

kennystrawnmusic:
Since it's rather difficult to manually get Android Studio ― the only reason why I in particular still use Crouton ― onto a Chromebook using Crouton without also installing a desktop environment and risking radically reducing the amount of available storage on a Chromebook, I decided it would be a great idea to actually create a target that installs Android Studio along with as little additional overhead as possible. First I got it installed; now it's time to share how I did it with the world.

### DIFF
--- a/host-bin/startandroidstudio
+++ b/host-bin/startandroidstudio
@@ -1,0 +1,17 @@
+#!/bin/sh -e
+# Copyright (c) 2016 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+set -e
+
+APPLICATION="${0##*/}"
+
+
+USAGE="$APPLICATION [options]
+Wraps enter-chroot to launch Android Studio in an xiwi window.
+By default, it will log into the primary user on the first chroot found.
+Options are directly passed to enter-chroot; run enter-chroot to list them."
+
+exec sh -e "`dirname "\`readlink -f "$0"\`"`/enter-chroot" -t android-studio "$@" "" \
+  exec xiwi -b /opt/google/android/studio/bin/studio.sh
--- a/host-bin/startandroidstudio
+++ b/host-bin/startandroidstudio
@@ -13,5 +13,4 @@ Wraps enter-chroot to launch Android Studio in an xiwi window.
 By default, it will log into the primary user on the first chroot found.
 Options are directly passed to enter-chroot; run enter-chroot to list them."
 
-exec sh -e "`dirname "\`readlink -f "$0"\`"`/enter-chroot" -t android-studio "$@" "" \
-  exec xiwi -b /opt/google/android/studio/bin/studio.sh
+exec startxiwi -b /opt/google/android/studio/bin/studio.sh
--- a/host-bin/startandroidstudio
+++ b/host-bin/startandroidstudio
@@ -13,5 +13,5 @@ Wraps enter-chroot to launch Android Studio in an xiwi window.
 By default, it will log into the primary user on the first chroot found.
 Options are directly passed to enter-chroot; run enter-chroot to list them."
 
-eval "exec sh -e "`dirname "\`readlink -f "$0"\`"`/enter-chroot" -t android-studio "$@" "" \
-  exec xiwi -b /opt/google/android/studio/bin/studio.sh"
+exec sh -e "`dirname "\`readlink -f "$0"\`"`/enter-chroot" -t android-studio "$@" "" \
+  exec xiwi -b /opt/google/android/studio/bin/studio.sh
--- a/host-bin/startandroidstudio
+++ b/host-bin/startandroidstudio
@@ -13,5 +13,5 @@ Wraps enter-chroot to launch Android Studio in an xiwi window.
 By default, it will log into the primary user on the first chroot found.
 Options are directly passed to enter-chroot; run enter-chroot to list them."
 
-exec sh -e "`dirname "\`readlink -f "$0"\`"`/enter-chroot" -t android-studio "$@" "" \
-  exec xiwi -b /opt/google/android/studio/bin/studio.sh
+eval "exec sh -e "`dirname "\`readlink -f "$0"\`"`/enter-chroot" -t android-studio "$@" "" \
+  exec xiwi -b /opt/google/android/studio/bin/studio.sh"
--- a/targets/android-studio
+++ b/targets/android-studio
@@ -24,7 +24,7 @@ if [ $DISTRO != "ubuntu" ]; then
      deb-src http://ppa.launchpad.net/ubuntu-desktop/ubuntu-make/ubuntu yakkety main
   EOF
 fi
-install ubuntu-make
+install openjdk-8-jre openjdk-8-jre-headless openjdk-8-jdk ubuntu-make
 yes 'a' | umake android /opt/google/android/studio
 
 TIPS="$TIPS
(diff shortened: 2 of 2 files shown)


idvoretskyi:
Is it compatible with Debian-based releases (not only Ubuntu)?

StrawnPoint04:
In order to make it compatible must add code to manually download/install umake if running a Debian-based, as opposed to Ubuntu-based, distribution. But yes, it's definitely possible for project maintainers to do.

StrawnPoint04:
Decided to add some conditions to allow more flexibility in the Debian regard

host-bin/startandroidstudio
```diff
@@ -0,0 +1,17 @@
...
+
+APPLICATION="${0##*/}"
+
+
+USAGE="$APPLICATION [options]
+Wraps enter-chroot to launch Android Studio in an xiwi window.
+By default, it will log into the primary user on the first chroot found.
+Options are directly passed to enter-chroot; run enter-chroot to list them."
+
+eval "exec sh -e "`dirname "\`readlink -f "$0"\`"`/enter-chroot" -t android-studio "$@" "" \
```
dnschneid: Not sure why you added the eval commit; you don't need it (and it screws up your quoting)

targets/android-studio
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh -e
+# Copyright (c) 2016 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+REQUIRES='xiwi gtk-extra'
```
dnschneid: I suppose this can be installed even if you prefer to use xorg; you'll just have the extra xiwi/extension install.

I do agree that startandroidstudio should target xiwi (since there's no sensible way to directly launch an app in native Xorg without using a DE or effectively recreating xiwi).

Yeah, I suppose this is the right way to do it...it just seems strange.

targets/android-studio
```diff
@@ -0,0 +1,23 @@
...
+DESCRIPTION='Installs Android Studio as an easily launchable xiwi application'
```
dnschneid: easily -> directly

targets/android-studio
```diff
@@ -0,0 +1,23 @@
...
+HOSTBIN='startandroidstudio'
+
+### Append to prepare.sh:
+mkdir /etc/bash-completion.d #Must do this in order to avoid package management error while installing umake
```
dnschneid: Put the comment on the previous line, please

targets/android-studio
```diff
@@ -0,0 +1,23 @@
...
+if [ $DISTRO != "ubuntu" ]; then
```
dnschneid: If you're going the PPA approach for non-ubuntu, you should do something more like [this](https://github.com/dnschneid/crouton/blob/90ad47ff22e148db41ba627992ee5cc5d0c14175/targets/xfce#L13).

targets/android-studio
```diff
@@ -0,0 +1,23 @@
...
+  echo "deb http://ppa.launchpad.net/ubuntu-desktop/ubuntu-make/ubuntu yakkety main\n\
+     deb-src http://ppa.launchpad.net/ubuntu-desktop/ubuntu-make/ubuntu yakkety main" >> /etc/apt/sources.list
+  install ubuntu-make
```
dnschneid: you will need to apt-get update before running install

targets/android-studio
```diff
@@ -0,0 +1,23 @@
...
+else
+  install ubuntu-make
```
dnschneid: both branches end in `install ubuntu-make`, so this can be outside the if statement

targets/android-studio
```diff
@@ -0,0 +1,23 @@
...
+fi
+yes 'a' | umake android /opt/google/android/studio
```
dnschneid: Is `yes` really the only way to go about this?

targets/android-studio
```diff
@@ -0,0 +1,23 @@
...
+
+TIPS="$TIPS
+You can run Android studio in a Chrome OS window via the startandroidstudio host command: sudo startandroidstudio
```
dnschneid: Wrap lines within 80 characters, please.

host-bin/startandroidstudio
```diff
@@ -13,5 +13,5 @@ Wraps enter-chroot to launch Android Studio in an xiwi window.
 By default, it will log into the primary user on the first chroot found.
 Options are directly passed to enter-chroot; run enter-chroot to list them."
 
-eval "exec sh -e "`dirname "\`readlink -f "$0"\`"`/enter-chroot" -t android-studio "$@" "" \
-  exec xiwi -b /opt/google/android/studio/bin/studio.sh"
+exec sh -e "`dirname "\`readlink -f "$0"\`"`/enter-chroot" -t android-studio "$@" "" \
+  exec xiwi -b /opt/google/android/studio/bin/studio.sh
```
dnschneid: actually, I don't think this is quite correct.  `xiwi` doesn't take -b, for example.  Look at `startxiwi` for reference.

targets/android-studio
StrawnPoint04: It's the only way to skip umake's license prompt.

host-bin/startandroidstudio
StrawnPoint04: Made it a wrapper for startxiwi instead

targets/android-studio
```diff
@@ -0,0 +1,33 @@
...
+# found in the LICENSE file.
+
+REQUIRES='xiwi gtk-extra'
+DESCRIPTION='Installs Android Studio as a directly launchable xiwi application'
+HOSTBIN='startandroidstudio'
+
+### Append to prepare.sh:
+# Must create the /etc/bash_completion.d directory
+# in order for umake to properly install
+mkdir /etc/bash_completion.d
```
DennisLfromGA: I get an error when the directory is pre-existing:
`mkdir: cannot create directory ‘/etc/bash_completion.d’: File exists`

Could/Should a '`|| true`' be appended to the command?

dnschneid: `mkdir -p` is the more elegant way

StrawnPoint04: Must be specific to yakkety/zesty then — get a "Directory non-existent" error when trying to configure in those cases.

targets/android-studio
```diff
@@ -0,0 +1,33 @@
...
+  #Pull in the signing key
+  apt-key adv --keyserver 'keyserver.ubuntu.com' \
+                --recv-keys '07D44A5424C7C12A4BAB1E80EB563F93142986CE'
+  #Add the software source
+  cat > '/etc/apt/sources.list.d/umake.list' << EOF
+     # Bring in Umake PPA on non-Ubuntu distros
+     deb http://ppa.launchpad.net/ubuntu-desktop/ubuntu-make/ubuntu yakkety main
+     deb-src http://ppa.launchpad.net/ubuntu-desktop/ubuntu-make/ubuntu yakkety main
+  EOF
+  install ubuntu-make
```
DennisLfromGA: Since `install ubuntu-make` is always needed, can't the conditional end on line 24 with a `fi` and the `install ubuntu-make` follow?

StrawnPoint04: Good point. Else statement not needed.

targets/android-studio
StrawnPoint04: OpenJDK8 and other dependencies of Android Studio in turn depend on GTK+, hence the gtk-extra target.

StrawnPoint04:
Running into another problem when testing ― "Target 'ubuntu-make' is not a directory" when running install function. Why is this?

PR #3130, opened 2017-03-09
lxde: Pull in additional packages on >= xenial

drinkcat:
Makes lxde at least somewhat functional on xenial

### DIFF
--- a/targets/lxde
+++ b/targets/lxde
@@ -11,6 +11,11 @@ CHROOTBIN='crouton-noroot startlxde'
 ### Append to prepare.sh:
 install lxde -- dictionaries-common chromium-browser network-manager wicd
 
+# xenial needs a few more packages.
+if release -ge xenial; then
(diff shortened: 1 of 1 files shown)


targets/lxde
```diff
@@ -11,6 +11,11 @@ CHROOTBIN='crouton-noroot startlxde'
 ### Append to prepare.sh:
 install lxde -- dictionaries-common chromium-browser network-manager wicd
 
+# xenial needs a few more packages.
+if release -ge xenial; then
+    install lxde-common lxsession-logout
```
dnschneid: these are available on trusty and some debian releases (at least sid) as well.  Do we want to install them explicitly there as well?

drinkcat: I think older Ubuntu releases pull in a reasonable set of packages. Not sure about Debian but I think it should be fine too.

PR #3143, opened 2017-03-10
EOL precise

dnschneid:
Marks precise as EOL'd!

precise!

### DIFF
--- a/installer/ubuntu/releases
+++ b/installer/ubuntu/releases
@@ -13,7 +13,7 @@ lucid!
 maverick!
 natty!
 oneiric!
-precise
+precise!
 quantal!
 raring!
 saucy!


DennisLfromGA:
Great timing 😜
image

drinkcat:
LGTM, actually ,-)

dnschneid:
I'm actually concerned there's no properly documented migration path from precise, though.  Especially since it looks like upgrading doesn't work; there needs to be documentation somewhere obvious so we don't get (quite as large) a torrent of bug reports.

dnschneid:
Any thoughts towards migration?  @DennisLfromGA did you ever figure out a decent upgrade path?

dnschneid:
OK at this point we should probably just merge this.

DennisLfromGA:
Sorry, I couldn't find a decent upgrade path, I tried and failed.
Going to almost any release from precise was problematic for me.

Issue #2922, opened 2016-12-02
Faulty regex used for release detection


keidax:
The `sed` command on line 23 of [`getrelease.sh`](https://github.com/dnschneid/crouton/blob/master/installer/ubuntu/getrelease.sh) assumes that `main` will always be first in the list of distribution components. However, when using Ubuntu (particularly Synaptic) the order of components in `/etc/apt/sources.list` sometimes change, and `main` is no longer the first component. This may very well be an Ubuntu bug, but I think it would make sense for crouton to adopt more lenient behavior here.

This handling of the release detection has contributed to other users' issues in the past (e.g. #1293, #2521, #2272).

w-barath:
Since the regex is incapable of parsing the sources.list [specification](http://manpages.ubuntu.com/manpages/yakkety/man5/sources.list.5.html), the script should either be rewritten to be able to parse the spec or it should simply not make the attempt to read sources.list in the first place.

Nowhere in the spec does it begin to suggest that "main" should even exist so using it as an anchor for a field is bogus.

This works better:
```
sed -n 's|^deb .*://[^ ]* \([^ -]*\).*$|\1|p' /etc/apt/sources.list
xenial
xenial
xenial
xenial
xenial
xenial
xenial
xenial
xenial
xenial

```

There's also other, easier, ways to get the release:

```
$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=16.04
DISTRIB_CODENAME=xenial
DISTRIB_DESCRIPTION="Ubuntu 16.04.2 LTS"

```
```
$ cat /etc/os-release 
NAME="Ubuntu"
VERSION="16.04.2 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04.2 LTS"
VERSION_ID="16.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
VERSION_CODENAME=xenial
UBUNTU_CODENAME=xenial

```

Issue #2521, opened 2016-04-07
Newly and suddenly invalid chroot, previously working for over a year


rusvnc:
```
sudo edit-chroot -all
name: trusty
encrypted: yes, locked
Unmounting /mnt/stateful_partition/crouton/chroots/trusty...
```

As of two days ago, my crouton install (trusty, command-line only) on this machine was working fine and had been for a long time. Today, when I try to start or update the chroot I get the following errors:

```
chronos@localhost ~ $ sudo enter-chroot
Password: 
Enter encryption passphrase for trusty: 
Entering /mnt/stateful_partition/crouton/chroots/trusty...
/mnt/stateful_partition/crouton/chroots/trusty doesn't appear to be a valid chroot.
Unmounting /mnt/stateful_partition/crouton/chroots/trusty...
```

```
chronos@localhost ~ $ sudo sh ~/Downloads/crouton -u -n trusty
Downloading latest crouton installer...
######################################################################## 100.0%
/usr/local/chroots/trusty already exists; updating it...
Enter encryption passphrase for trusty: 
Unable to determine the release in /usr/local/chroots/trusty. Please specify it with -r.
```

```
chronos@localhost ~ $ sudo sh ~/Downloads/crouton -u -n trusty -r trusty
/usr/local/chroots/trusty already exists; updating it...
Enter encryption passphrase for trusty: 
Release doesn't match! Please correct the -r option, or specify a second -u to
change the release, upgrading the chroot (dangerous).
chronos@localhost ~ $ sudo sh ~/Downloads/crouton -u -u -n trusty -r trusty
/usr/local/chroots/trusty already exists; updating it...
Enter encryption passphrase for trusty: 
WARNING: Changing the chroot release to trusty.
Press Control-C to abort; upgrade will continue in 5 seconds.
chronos@localhost ~ $
```

The upgrade does not complete, it aborts.

On ChromeOS Version 50.0.2661.57 beta. This problem may have started after the update, but I can't remember precisely.

cyrildtm:
Same issue here. Tried methods in #1293 but no help. 

```
@localhost:/etc/apt$ cat sources.list
deb http://archive.ubuntu.com/ubuntu/ trusty multiverse universe restricted main 
deb-src http://archive.ubuntu.com/ubuntu/ trusty multiverse universe restricted main 
deb http://archive.ubuntu.com/ubuntu/ trusty-updates multiverse universe restricted main 
[...]

Issue #2272, opened 2015-12-03
crouton update fails with trusty


seanmahrt:
I am trying to update trusty from crouton first before I go into the chroot to update it (seems like a good idea).  I deleted crouton and redownloaded a fresh version and it still won't update.

command:

sudo sh ./crouton -u -n trusty
 response: /usr/local/chroots/trusty already exists; updating it...
Unable to determine the release in /usr/local/chroots/trusty. Please specify it with -r.

name: trusty
encrypted: no
Entering /mnt/stateful_partition/crouton/chroots/trusty...
crouton: version 1-20150827155505~master:edbce112
release: trusty
architecture: amd64
xmethod: xiwi
targets: xiwi,xorg,lxde
host: version 7390.68.0 (Official Build) stable-channel peppy 
kernel: Linux localhost 3.8.11 #1 SMP Thu Nov 5 10:53:38 PST 2015 x86_64 x86_64 x86_64 GNU/Linux
freon: yes
Unmounting /mnt/stateful_partition/crouton/chroots/trusty...

Where to next?

DennisLfromGA:
Your chroot may not be fully setup, or something has changed.
Maybe just try to get into it first via `sudo enter-chroot -n trusty`
and see if that succeeds. If it tells you the setup didn't finish and ask if you want to proceed, go for it.
Then, either update your chroot via `sudo apt-get update; sudo apt-get upgrade`
or exit the chroot and run your crouton update command again.

seanmahrt:
I'll try that. I know the enter-Chroot works. Just was surprised the update crouton wasn't happy. I'm pretty sure the apt update will work too

Sean

DennisLfromGA:
Yeah, the message:
`Unable to determine the release in /usr/local/chroots/trusty. Please specify it with -r.`
seems to indicate something is definitely wrong with the chroot...

seanmahrt:
Somehow my update list was really hosed up (us.release.ubuntu.com) instead
of us.archive.ubuntu.com.  Wonder if that was from the original crouton
trusty install...

DennisLfromGA:
Yeah, I don't know how '*us.release.ubuntu.com*' got in there.
Is the crouton update working for you now?

seanmahrt:
Yes, after I corrected it, the crouton update worked great. Might be good to cough an error code or something for totally screwy repositories. 

Sean

w-barath:
[...]

PR #3170, opened 2017-03-20
fix regex to be more robust

### LINKED ISSUE #1293: crouton -u gives "Release doesn't match" errors for sid
Trying to add a target, I get "Release doesn't match" errors:

chronos@localhost ~ $ sudo sh ~/Downloads/crouton -u -t xiwi -n sid
/usr/local/chroots/sid already exists; updating it...
Enter encryption passphrase for sid: 
Unable to determine the release in /usr/local/chroots/sid. Please specify it with -r.
chronos@localhost ~ $ sudo sh ~/Downloads/crouton -u -t xiwi -n sid -r sid
/usr/local/chroots/sid already exists; updating it...
Enter encryption passphrase for sid: 
Release doesn't match! Please correct the -r option, or specify a second -u to
change the release, upgrading the chroot (dangerous).
chronos@localhost ~ $ sudo sh ~/Downloads/crouton -u -t xiwi -n sid -r sid -u
/usr/local/chroots/sid already exists; updating it...
Enter encryption passphrase for sid: 
WARNING: Changing the chroot release to sid.
Press Control-C to abort; upgrade will continue in 5 seconds.
chronos@localhost ~ $ sudo sh ~/Downloads/crouton -u -t xiwi -n sid -r sid
/usr/local/chroots/sid already exists; updating it...
Enter encryption passphrase for sid: 
Release doesn't match! Please correct the -r option, or specify a second -u to
change the release, upgrading the chroot (dangerous).

The chroot is encrypted and
$ croutonversion 
crouton: version 1-20141231011135~master:992afe3e
release: sid
architecture: amd64
targets: gnome
host: version 6310.68.0 (Official Build) stable-channel quawks 

I did a few tweaks inside the chroot after installing it, but I can't imagine how any of them would have broken then release detection...

Thanks,
JM

w-barath:
The original regex doesn't follow the deb88 specification for deb URLs, it assumes that "main" will immediately follow the release name, when the specification doesn't even mention "main".

This fixes #1293, #2521, #2272 and fulfills enhancement #2922

### DIFF
--- a/installer/ubuntu/getrelease.sh
+++ b/installer/ubuntu/getrelease.sh
@@ -20,7 +20,7 @@ fi
 
 # Lookup the release name from the field after the URI
 # We identify URI by '://'
-rel="`sed -n 's|^deb .*://[^ ]* \([^ ]*\) main\( .*\)\?$|\1|p' \
+rel="`sed -n 's|^deb .*://[^ ]* \([^ -]*\).*$|\1|p'' \
     "$sources" "$sources.d"/*.list 2>/dev/null | head -n 1`"
 if [ -z "$rel" ] || \
         ! grep -q "^$rel\([^a-z].*\)*$" "`dirname "$0"`/releases"; then


w-barath:
|I signed it!|

installer/ubuntu/getrelease.sh
```diff
@@ -20,7 +20,7 @@ fi
 
 # Lookup the release name from the field after the URI
 # We identify URI by '://'
-rel="`sed -n 's|^deb .*://[^ ]* \([^ ]*\) main\( .*\)\?$|\1|p' \
+rel="`sed -n 's|^deb .*://[^ ]* \([^ -]*\).*$|\1|p'' \
```
dnschneid: There's an extra ' at the end

dnschneid:
I dunno how much you've tested this, but even with the extra ' removed, this doesn't seem to actually work.

dnschneid:
@w-barath what would you like to do with this patch?

w-barath:
Sorry for the long delay responding.

Long story short, there are official repos for Ubuntu and Debian which contain hyphenated release names, ie trusty-backports.  If the user places one of these above the vanilla repo entry, or if the user has temporarily disabled the vanilla repo, then the script will reply with 'trusty-backports'.

If you look at the changes I made to the regex, I amended it to drop latter portion if it picks up a hyphenated release name.  I also removed the requirement to find 'main' since that can easily be disabled when the user only wants security updates and disables main etc to prevent installing new packages.

I've not very familiar with github's feature set... I'd like to close/remove this pull request and re-submit it with the extra ' removed, but can't see any way to do that... going to see if the 'close and comment' on this form gets me a step in that direction. :-)

PR #3185, opened 2017-04-02
Add dialout group to sync, for use with programming microcontrollers

WRansohoff:
This lets the crouton environment work with IDEs like Arduino, NodeMCU, etc. I've only tested on an Asus C100 Flip (ARM.)


dnschneid:
Are tape and tty also needed?  Or just dialout?

WRansohoff:
They were already in the file; I just moved them down a line to keep the line length about the same.

dnschneid:
haha I can't read

Yes, this looks good.  @smibarber can you think of any downsides here?

smibarber:
Nope, seems fine to me.

PR #3190, opened 2017-04-05
Add target for spectrwm.

jcvernaleo:
spectrwm is a lightweight tiling window manager.

### DIFF
--- a/chroot-bin/startspectrwm
+++ b/chroot-bin/startspectrwm
@@ -1,0 +1,10 @@
+#!/bin/sh -e
+# Copyright (c) 2016 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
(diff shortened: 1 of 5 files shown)


jcvernaleo:
I've only tested this on xenial.  Very old debian might have it with a different name.  I would test on everything, but my only chromebook is an arm device with fairly limited storage so testing lots of targets is pretty painful but of course if that is the process I can go and try to test this out on more install targets.

jcvernaleo:
I signed it.

dnschneid:
Thanks for the patches (and sorry you had to go through the CLA process), but in general, crouton tries to avoid having targets for apps and WMs (and tries to minimize the number of DEs as well).  You can see #1262 with a similar explanation.

Adding targets creates a pretty large testing and support burden on crouton (especially for people bringing in entirely new distros), so we try to stick to only the major (inferior? :) ) ones.

PR #3268, opened 2017-06-12
edit-chroot: use realpath to canonicalize $dest

### LINKED ISSUE #3199: When specifying a backup location, edit-chroot says it's literally backing up to what you put for -f
#### Please describe your issue:
Minor. `edit-chroot`, when backing up and specifying `-f`, it will just take what you put for the flag instead of the actual path.
pic
#### If known, describe the steps to reproduce the issue:
* Make a chroot.
* Go to your `bin` directory for the chroot.
* Back it up through `edit-chroot`, and specify a path for `-f` (maybe even one that includes `../`).
* See something like the picture above happen.

jbaum98:
Fixes #3199

### DIFF
--- a/host-bin/edit-chroot
+++ b/host-bin/edit-chroot
@@ -403,7 +403,7 @@ for NAME in "$@"; do
             tardest="$dest"
             splitpid=''
         fi
-        echo -n "  Backing up $CHROOT to $dest..." 1>&2
+        echo -n "  Backing up $CHROOT to $(readlink -m -- "$dest")..." 1>&2
         addtrap "echo 'Deleting partial archive.' 1>&2; \
                  kill '$splitpid' 2>/dev/null; rm -f '$dest' '$dest.part-'*"
         ret=0


dnschneid:
I think `readlink -m` would be a better choice; we don't use realpath anywhere else.

host-bin/edit-chroot
```diff
@@ -403,7 +403,7 @@ for NAME in "$@"; do
             tardest="$dest"
             splitpid=''
         fi
-        echo -n "  Backing up $CHROOT to $dest..." 1>&2
+        echo -n "  Backing up $CHROOT to `realpath $dest`..." 1>&2
```
DennisLfromGA: I believe the crouton devs have converted most if not all syntax to replace grave accents (backticks) to the newer convention using $(...).

jbaum98:
I've changed the commit to use `readlink -m` and `$(...)` instead of backticks.

host-bin/edit-chroot
```diff
@@ -403,7 +403,7 @@ for NAME in "$@"; do
             tardest="$dest"
             splitpid=''
         fi
-        echo -n "  Backing up $CHROOT to $dest..." 1>&2
+        echo -n "  Backing up $CHROOT to $(readlink -m $dest)..." 1>&2
```
dnschneid: Readlink should look like `$(readlink -m -- "$dest")`
...that is, with double-quotes and the -- parameter.

PR #3269, opened 2017-06-12
src/freon.c: Changes long casts to uintptr_t casts

### LINKED ISSUE #3141: use `uintptr_t` instead of `long` with pointers->integers
src/freon.c does:
```c
        if ((request == VT_RELDISP && (long)data == 1) ||
            (request == VT_ACTIVATE && (long)data == 0)) {
```

that should be `(uintptr_t)` instead of `(long)`.

jbaum98:
Fixes #3141


drinkcat:
Thanks!

PR #3273, opened 2017-06-14
src/freon.c: Check before extracting mode in open

### LINKED ISSUE #3140: src/freon.c mishandles 3rd arg to open functions (mode)
the `open` signature is:
`int open(const char *pathname, int flags, ...);`

and `mode` is only passed in for certain flags (`O_CREAT` and `O_TMPFILE`), so trying to extract it otherwise is undefined behavior.  you might get garbage, or you might crash.

freon.c assumes it's always passed in:
```c
int open(const char *pathname, int flags, ...) {
    if (!orig_open) preload_init();

    va_list argp;
    va_start(argp, flags);
    mode_t mode = va_arg(argp, mode_t);
    va_end(argp);
```

same for `open64`.

you can do something instead like:
```c
int open(const char *pathname, int flags, ...) {
    if (!orig_open) preload_init();

    mode_t mode = 0;
    va_list argp;
    va_start(argp, flags);
    if (flags & (O_CREAT | O_TMPFILE))
        mode = va_arg(argp, mode_t);
    va_end(argp);
```

jbaum98:
Fixes #3140

### DIFF
--- a/src/freon.c
+++ b/src/freon.c
@@ -218,9 +218,11 @@ static int _open(int (*origfunc)(const char *pathname, int flags, mode_t mode),
 int open(const char *pathname, int flags, ...) {
     if (!orig_open) preload_init();
 
+    mode_t mode = 0;
     va_list argp;
     va_start(argp, flags);
-    mode_t mode = va_arg(argp, mode_t);
+    if (flags & (O_CREAT | O_TMPFILE))
+        mode = va_arg(argp, mode_t);
     va_end(argp);
 
     return _open(orig_open, "open", pathname, flags, mode);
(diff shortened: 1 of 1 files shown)


src/freon.c
```diff
@@ -218,9 +218,11 @@ static int _open(int (*origfunc)(const char *pathname, int flags, mode_t mode),
 int open(const char *pathname, int flags, ...) {
     if (!orig_open) preload_init();
 
+    mode_t mode = 0;
     va_list argp;
     va_start(argp, flags);
-    mode_t mode = va_arg(argp, mode_t);
+    if (flags & (O_CREAT | O_TMPFILE))
```
dnschneid: O_TMPFILE apparently isn't defined in precise.  I think you need some sort of compatibility define.

dnschneid:
Fixing this PR should be pretty straightforward...

jbaum98:
Sorry I didn't see this. So should I define `O_TMPFILE` to be a number with only ones so that bitwise or-ing it has no effect? Ie `~0`?

dnschneid:
No, I suspect you're just missing an include.

jbaum98:
Okay I've added an `#include <fcntl.h>` based on the man page, but I don't have a linux machine handy right now to test it.

Issue #2875, opened 2016-11-10
Installing xserver-xorg-legacy (part of the kali-linux meta package) breaks startxfce4 


Normola:
```
name: kali-rolling
encrypted: yes, locked
Unmounting /media/removable/SD Card/chroots/kali-rolling...```

#### Please describe your issue:
Installing xserver-xorg-legacy (part of the kali-linux meta package) breaks startxfce4 

log file

[  7453.336] 
X.Org X Server 1.18.4
Release Date: 2016-07-19
[  7453.336] X Protocol Version 11, Revision 0
[  7453.336] Build Operating System: Linux 3.16.0-4-amd64 x86_64 Debian
[  7453.336] Current Operating System: Linux localhost 3.14.0 #1 SMP PREEMPT Tue Oct 25 23:34:16 PDT 2016 x86_64
[  7453.337] Kernel command line: cros_secure console= loglevel=7 init=/sbin/init cros_secure oops=panic panic=-1 root=PARTUUID=59508fa1-1d55-ce47-815a-f08a570e08be/PARTNROFF=1 rootwait rw dm_verity.error_behavior=3 dm_verity.max_bios=-1 dm_verity.dev_wait=0 dm="1 vroot none ro 1,0 3584000 verity payload=ROOT_DEV hashtree=HASH_DEV hashstart=3584000 alg=sha1 root_hexdigest=60f50309ebc53ad91473ded14618cfe55a2e9532 salt=21acf051d6c7c0608925c13793ab33b7afd97ec40ac4336251b254d192e4e1f0" noinitrd vt.global_cursor_default=0 kern_guid=59508fa1-1d55-ce47-815a-f08a570e08be add_efi_memmap boot=local noresume noswap i915.modeset=1 tpm_tis.force=1 tpm_tis.interrupts=0 nmi_watchdog=panic,lapic  
[  7453.337] Build Date: 06 September 2016  01:32:44PM
[  7453.337] xorg-server 2:1.18.4-2 (https://www.debian.org/support) 
[  7453.338] Current version of pixman: 0.34.0
[  7453.338] 	Before reporting problems, check http://wiki.x.org
	to make sure that you have the latest version.
[  7453.338] Markers: (--) probed, (**) from config file, (==) default setting,
	(++) from command line, (!!) notice, (II) informational,
	(WW) warning, (EE) error, (NI) not implemented, (??) unknown.
[...]

PR #3277, opened 2017-06-15
Kodi

zguithues:
Kodi is the new name for XBMC.

These changes will rename the target and supporting files to kodi.  I also set it up to add the repository in order to get the latest version.

Ubuntu - Official TeamKodi PPA
Debian - Jessie Backports (recommended for debian in kodi wiki)
Kali - not supported.  will bail out of install with error.

I've tested with several releases, and it seems to work best with modern ubuntu (xenial+).

Features/Notes

- closes chroot when kodi is closed
- can play h.264 & h.265 out of the box (tested with Acer C720 peppy & Toshiba Chromebook 2 swanky)
- smb/samba works out of the box.  cannot scan, must enter port with hostname. 
- volume is separate from ChromeOS. (make sure it's turned up in chromeos)
- DOES NOT work with kiwi/extension
- I recommend making separate from other targets. (sudo sh ~/Downloads/crouton -t kodi -n kodi)

p.s. this is my first pull request ever, let me know if i did something wrong ;)


targets/kodi
```diff
@@ -0,0 +1,50 @@
...
+# found in the LICENSE file.
+REQUIRES='x11'
+DESCRIPTION='Installs the KODI media player. (Approx. 140MB)'
+HOSTBIN='startkodi'
+CHROOTETC='kodi-keyboard.xml kodi-cycle.py'
+. "${TARGETSDIR:="$PWD"}/common"
+
+### Append to prepare.sh:
+if [ "$DISTRO" = 'debian' ]; then
+    echo "deb http://http.debian.net/debian jessie-backports main" | sudo tee /etc/apt/sources.list.d/kodi.list
```
dnschneid: prepare already runs as root, so you can just use `>` to write the file.

targets/kodi
```diff
@@ -0,0 +1,50 @@
...
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+REQUIRES='x11'
+DESCRIPTION='Installs the KODI media player. (Approx. 140MB)'
+HOSTBIN='startkodi'
+CHROOTETC='kodi-keyboard.xml kodi-cycle.py'
+. "${TARGETSDIR:="$PWD"}/common"
+
+### Append to prepare.sh:
+if [ "$DISTRO" = 'debian' ]; then
```
dnschneid: We shouldn't be using jessie-backports in sid, should we?  The right condition would then be `if release -le jessie; then`, which will apply only to debian releases jessie or earlier.

targets/kodi
```diff
@@ -0,0 +1,50 @@
...
+    echo "deb http://http.debian.net/debian jessie-backports main" | sudo tee /etc/apt/sources.list.d/kodi.list
+fi
+
+if [ "$DISTRO" = 'ubuntu' ]; then
+    sudo apt-get install -y software-properties-common
```
dnschneid: you never need sudo in a target; you already have it

dnschneid: Refer to this [old xfce target](https://github.com/dnschneid/crouton/blob/90ad47ff22e148db41ba627992ee5cc5d0c14175/targets/xfce#L13) for how to pull in PPAs.  We don't want to unnecessarily install things.

targets/kodi
```diff
@@ -0,0 +1,50 @@
...
+    sudo add-apt-repository -y ppa:team-xbmc/ppa
+fi
+
+if [ "$DISTRO" = 'kali' ]; then
+    error 99 "Kodi not supported on Kali Linux"
```
dnschneid: This check should be done in the pre-prepare area, before you've gone through most of the crouton install.  See the ever-shrinking [unity target](https://github.com/dnschneid/crouton/blob/90ad47ff22e148db41ba627992ee5cc5d0c14175/targets/unity).

targets/kodi
```diff
@@ -0,0 +1,50 @@
...
+fi
+
+sudo apt-get update
```
dnschneid: This should only be run on distros that add the repository (e.g., put it in the if statements above).  Also, it seems apt-get update should be `|| true`'d in case the user has broken repos.

targets/kodi
```diff
@@ -0,0 +1,50 @@
...
+# add a script in /etc/profile.d to link ~/.kodi/userdata/keymaps/keyboard.xml
+# to /etc/crouton/kodi-keyboard.xml
+
+profiledsh='/etc/profile.d/crouton-kodi-keymaps.sh'
+# Make sure symbolic link is setup on login
+echo '#!/bin/sh
+
+keyboardxmldir="$HOME/.kodi/userdata/keymaps"
+# Do not install if user is root, or $HOME does not exist
+if [ "`id -u`" -ne 0 -a -d "$HOME" -a ! -e "$keyboardxmldir/keyboard.xml" ]; then
```
dnschneid: should be `"$(id -u)"`

targets/xbmc
```diff
@@ -1,35 +0,0 @@
-#!/bin/sh -e
```
dnschneid: I don't think we can outright delete this target; we need to provide a migration path.

One quick-and-dirty solution would be to make xbmc a minimal target like [x11](https://github.com/dnschneid/crouton/blob/90ad47ff22e148db41ba627992ee5cc5d0c14175/targets/x11), with an unconditional `REQUIRES=kodi`, and add `PROVIDES='xbmc'` to the kodi target (like in [xorg](https://github.com/dnschneid/crouton/blob/90ad47ff22e148db41ba627992ee5cc5d0c14175/targets/xorg)).  Perhaps even leave DESCRIPTION blank and update installer/main.sh to avoid printing out any targets that have a blank DESCRIPTION for `-t list`.

The best solution would be to make xbmc a symlink to kodi, and then add some smarts to installer/main.sh to interpret a symlinked target as if it were the above.  This would be a little tricky (but not too hard) because the script would have to automatically populate PROVIDES by detecting any target symlinks that point to the selected target.

zguithues:
Thanks for the suggestions!

I'll work through the list and resubmit when i've got them resolved.

targets/kodi
zguithues: I'm not certain what " if [ "${TARGETNOINSTALL:-c}" = 'c' ]; then " means..

Here is what i put at the top:

```
if [ "${TARGETNOINSTALL:-c}" = 'c' ]; then
    if [ "$DISTRO" = 'kali' ]; then
        error 99 "Kodi not supported on Kali Linux."
    fi
fi
```

Is this correct?

targets/xbmc
zguithues: I've created the minimal target, but i'm unsure what exactly to add to installer/main.sh.

I assume it would be around line 306, but i'm not familiar with the syntax.

zguithues:
Ok @dnschneid, I've made the requested changes.  I think everything is solid except for my questions regarding the kali linux and the xbmc migration stuff.

I haven't tested yet, as my chromebook is at home.  I will test it tonight and report back.

Thanks for your help getting this setup!  kodi is one of the main reasons i use crouton, and i'm glad to see it returning! :)

targets/xbmc
```diff
@@ -2,34 +2,8 @@
 # Copyright (c) 2016 The crouton Authors. All rights reserved.
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
-REQUIRES='x11'
-DESCRIPTION='Installs the XBMC media player. (Approx. 140MB)'
-HOSTBIN='startxbmc'
-CHROOTETC='xbmc-keyboard.xml xbmc-cycle.py'
-. "${TARGETSDIR:="$PWD"}/common"
+REQUIRES='kodi'
+PROVIDES='xbmc'
```
dnschneid: PROVIDES would go in the kodi target, not the xbmc target.

targets/kodi
dnschneid: Yep!  It's not well-documented.

targets/xbmc
dnschneid: It looks like in the help case, we set `TARGETS='help'`, on line 309 we set `TARGETNOINSTALL='y'` and then source the target file.  So it looks like the right tweak would be actually in [the target common file](https://github.com/dnschneid/crouton/blob/master/targets/common#L29), line 29, where it prints out the help text.  If DESCRIPTION is blank, we don't want to do that.

zguithues: i'm sorry, this went right over my head...  please dumb it down a bit ;)

I believe this is the only remaining item to resolve before this is good to merge.  Thanks a ton for your help! :)
Once merged, i'll reach out to the kodi guys and see if they'll feature it in their blog ;)

I also plan to make a wiki page that talks about it a bit.

dnschneid: Wrap [the common target](https://github.com/dnschneid/crouton/blob/master/targets/common#L30) lines 30 through 34 in a `if [ -n "$DESCRIPTION" ]` block :)

zguithues:
ok @dnschneid I think i've resolved all of the issues.  I've tested installing to target xbmc and it installs kodi.  also, xbmc does not display in the target list due to it's blank description.

I tested on sid, and kodi was unable to start.  kodi in debian repositories is very old and i decided it would be better to just make it ubuntu only, as that is the only official repository.  I edited the description to: "Installs the KODI media player. Ubuntu Only. (Approx. 140MB)".

Let me know if you think i should handle anything differently.

zguithues:
one thing i noticed... when i was testing, it would bail out for kali & debian before creating the chroot & downloading the bootstrap. but now it's downloading the whole thing before stopping. I'm not sure what i did to change it, and i can't duplicate the quick bailout.

Any thoughts? am i just crazy?

targets/kodi
```diff
@@ -5,27 +5,19 @@
 if [ "${TARGETNOINSTALL:-c}" = 'c' ]; then
     if [ "$DISTRO" = 'kali' ]; then
         error 99 "Kodi not supported on Kali Linux."
+    elif [ "$DISTRO" = 'debian' ]; then
```
dnschneid: Seems to be current in [sid](https://packages.debian.org/sid/video/kodi) (17), and jessie-backports seems to be version 16, which doesn't seem that old.  Can you get it to work so we don't artificially limit targets?

Generally we avoid creating targets that are only available in certain distros (unity being the obvious exception to that rule).

zguithues: I'll play around with it...  When i tried to install it on sid, it wasn't able to connect to X and couldn't start kodi.  I'll see if i can figure out why tonight.

zguithues:
@dnschneid 
I figured out why i was having trouble with kodi on kali & debian.  Its the xserver-xorg-legacy package!
detailed in #2875 

xfce is also broken in debian (and i assume kali) releases.

I'm sure there is a better way to solve this on the release script side, and i defer to your wisdom on this.  but the quickfix hammer i found was to add "remove xserver-xorg-legacy" to the target file.  Since it is such a detrimental package, and running that command doesn't have negative effects if the package is not installed; i just had it run for all releases

I'm running a few tests on various releases to see if i need to set anything specific.  I'll probably add back the jessie backports for jessie and older.  I'll update when i think it's ready to merge.

FYI, looking through the log, it looks like the package xserver-xorg-legacy was installed along with a bunch of xserver-xorg packages.  right after "Compiling /usr/local/lib/croutonfreon.so"
I'll add this to the open issue as well ;)

zguithues:
turns out kodi doesn't work on Wheezy or Jessie.  I'm testing on trusty and precise right now.  I'll have it bail out on any releases that don't work.

should have final changes tomorrow

zguithues:
I think i have everything resolved now @dnschneid 
Let me know if there is anything i missed :)

Thanks for all your help!

targets/xbmc
zguithues: done

zguithues:
I can't figure out how to clear the remaining requested changes...  I'm pretty sure I have addresses each requested change...

Should i clean the changes up into fewer commits and resubmit?

targets/kodi
```diff
@@ -0,0 +1,77 @@
+#!/bin/sh -e
+# Copyright (c) 2016 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+if [ "${TARGETNOINSTALL:-c}" = 'c' ]; then
+    if release -le wheezy; then
+        error 99 "Kodi not supported on Debian Wheezy or older."
+    fi
+    if release = precise; then
```
dnschneid: better to use -eq here...this probably would work, though.

dnschneid: this could be elif

targets/kodi
```diff
@@ -0,0 +1,77 @@
...
+        error 99 "Kodi not supported on Precise Ubuntu."
+    fi
+fi
+REQUIRES='x11'
+PROVIDES='xbmc'
+DESCRIPTION='Installs the KODI media player. Ubuntu Only. (Approx. 140MB)'
```
dnschneid: "ubuntu only" isn't true anymore; you can just delete that

targets/kodi
dnschneid: (also down below)

targets/kodi
```diff
@@ -0,0 +1,77 @@
...
+    cat > '/etc/apt/sources.list.d/kodi.list' <<EOF
+# Bring in Jessie Backports for newer Kodi
+deb http://http.debian.net/debian jessie-backports main
+EOF
+    # Update database
+    apt-get update || true
+fi
+
+# Get rid of disruptive xserver-xorg-legacy package
+if release = stretch || release = sid || release = kali-rolling; then
```
dnschneid: can be combined: `if release -ge stretch -eq kali-rolling`

targets/kodi
```diff
@@ -0,0 +1,77 @@
...
+    remove xserver-xorg-legacy
```
dnschneid: wait, why is this workaround here, anyway?  It even happens before installing kodi.  Probably needs to be a separate pull request, implementing the workaround I mention [here](https://github.com/dnschneid/crouton/issues/2875#issuecomment-261073012).

zguithues: I put this workaround in so it would work when testing.  I think a dummy package is probably the way to go.  I'll remove this workaround from the kodi pull request now.  I will try to take a crack at the dummy package later.

zguithues:
OK, I've made the requested changes.

are we good to go now? :)

targets/kodi
```diff
@@ -0,0 +1,71 @@
...
+DESCRIPTION='Installs the KODI media player. (Approx. 140MB)'
+HOSTBIN='startkodi'
+CHROOTETC='kodi-keyboard.xml kodi-cycle.py'
+. "${TARGETSDIR:="$PWD"}/common"
+
+### Append to prepare.sh:
+
+## Debian Overrides
+# Kodi wiki recommends using the Jessie Backports repository
+if release = jessie; then
```
dnschneid: missed one :)

dnschneid:
Also, I think the xbmc test needs to be changed into kodi: https://github.com/dnschneid/crouton/blob/master/test/tests/w6-xbmc

zguithues:
ok, i changed the text to say kodi...  I'm not 100% sure what this is or does, so let me know if more is needed. :)

targets/kodi
zguithues: actually, there were 2 instances of "=", but now i'm not sure if i am supposed to switch both of them...

dnschneid: yeah, always use -eq instead of = for the release command

targets/kodi
```diff
@@ -0,0 +1,71 @@
...
+# Bring in Jessie Backports for newer Kodi
+deb http://http.debian.net/debian jessie-backports main
+EOF
+    # Update database
+    apt-get update || true
+fi
+
+## Ubuntu Overrides
+# Kodi has an official ppa for the latest stable version
+if [ "$DISTRO" -eq 'ubuntu' ]; then
```
dnschneid: ah, sorry, but string comparisons should stick to ='s.

`release` is a special command that converts the release name into a number and does numerical comparisons.

zguithues: i figured it was something like that...

zguithues:
ok, that's all the pending changes requested...  anything else?

Thanks a ton for being so helpful! :)

dnschneid:
Looks good to me!  Now I run some tests and merge if things pass!

Thanks for making and cleaning up the patches!

dnschneid:
It looks like the PPA doesn't have any armhf/arm64 packages, so ARM devices will have to go without the PPA.  Currently, adding the PPA breaks install on ARM devices, since the metapackages are for any architecture and reference packages that cannot be fulfilled:
```
152 [i] The following packages have unmet dependencies:
153 [i]  kodi : Depends: kodi-bin (>= 2:17.3+git20170525.0741-final-0xenial) but 15.2+dfsg1-3ubuntu1.1 is to be installed
153 [i]         Depends: libshairplay0 but it is not installable
153 [i]         Depends: libcec4 but it is not installable
153 [i]         Recommends: libvdpau1 but it is not going to be installed
153 [i]         Recommends: i965-va-driver but it is not installable or
153 [i]                     libva-intel-vaapi-driver but it is not installable
```

zguithues:
I'll set it to not add the PPA for armhf/arm64 devices.
It looks like old 15.2 kodi is in the default repositories, should i let it install that, or should i just bail out if arm/arm64?  I don't have a way to test how it works on armhf/arm64.

zguithues:
ok, i set it to only add the ppa on x86,  can you test to see if it installs, or if i should just block arm? @dnschneid

dnschneid:
I'll give it a shot.  Thanks!

zguithues:
@dnschneid i've been using kodi and the one annoyance that i still have is the screen timeout...

reading the page on the wiki here: https://github.com/dnschneid/crouton/wiki/Power-manager-overrides
it seems like i should setup the wrapper script, but i feel like this could/should be implemented through the startkodi executable.  Either way, I'm not really sure how i should proceed.  Is there a precedent that I should follow, or do you have some insight?

Regarding which settings to adjust.  I believe we want to disable the dim_ms & off_ms for both plugged and unplugged.  probably need to adjust the suspend_ms as well...

Alternatively, is there a way for kodi to report that it's being "active" when playing video, so that it doesn't trigger any timeouts?

zguithues:
I created the Kodi wiki page here: https://github.com/dnschneid/crouton/wiki/Kodi-Media-Center

I'll fluff it out in the next few weeks.

host-bin/startkodi
```diff
@@ -9,10 +9,10 @@ APPLICATION="${0##*/}"
...
-Wraps enter-chroot to start an XBMC session.
+Wraps enter-chroot to start an KODI session.
 By default, it will log into the primary user on the first chroot found.
 
 Options are directly passed to enter-chroot; run enter-chroot to list them."
 
-exec sh -e "`dirname "\`readlink -f -- "$0"\`"`/enter-chroot" -t xbmc "$@" "" \
-    exec xinit /usr/bin/xbmc
+exec sh -e "`dirname "\`readlink -f -- "$0"\`"`/enter-chroot" -t kodi "$@" "" \
+    exec xinit /usr/bin/kodi --standalone
```
dnschneid: `exec croutonpowerd -i xinit /usr/bin/kodi --standalone` might do what you want with respect to power management.  It will prevent the system from idly going to sleep as long as the command (xinit /usr/bin/kodi in this case) is running.

zguithues: Thanks a ton!  I'll check it out!

zguithues:
I have made the change that disables power management while kodi is running.  I believe this was my last remaining issue to resolve.

How has your testing gone @dnschneid ?  Do you think it's ready to merge with master?

dnschneid:
Looks good from a testing point of view, although we need to fix the xserver-xorg-legacy thing.  Thanks again for this patch!

zguithues:
Thanks so much for your help @dnschneid !

I'll keep it updated if i find anything that needs fixed.
Next step is to tell the kodi people they just got a new platform ;)

DennisLfromGA:
RE: https://github.com/dnschneid/crouton/issues/3395

Not sure exactly what device this is since `edit-chroot -all` wasn't run but it's an ARM according to the `chrome` target output. They specified the `xbmc` target and got "E: Unable to locate package kodi" :(

zguithues:
Thanks for pointing it out to me, i responded to the issue.  I think the problem is that he was using trusty, and kodi isn't being maintained on arm trusty.

PR #3298, opened 2017-06-26
Add support for opensuse

mbgg:
Know issues:

1. adhd compilation with gcc7 is broken. Patch to fix this is submitted. As a workaround we would need to pass --Wno-error=int-in-bool-context (this holds for tumbleweed, leap 42.2 should work without a problem).
2. no audio working, need investigation

normal instalation should work without problems:
sudo sh crouton -r tumbleweed -n test -t xiwi,xfce,keyboard,cli-extra
sudo sh crouton -r leap-42.2 -n test -t xiwi,xfce,keyboard,cli-extra

### DIFF
--- a/installer/opensuse/bootstrap
+++ b/installer/opensuse/bootstrap
@@ -15,6 +15,7 @@ url_tumbleweed_aarch64="http://download.opensuse.org/ports/aarch64/tumbleweed/im
 # Leap
 url_leap_42_2_armv7hl="http://download.opensuse.org/ports/armv7hl/distribution/leap/42.2/appliances/openSUSE-Leap42.2-ARM-JeOS.armv7-rootfs.armv7l.tbz"
 url_leap_42_2_aarch64="http://download.opensuse.org/ports/aarch64/distribution/leap/42.2/appliances/openSUSE-Leap42.2-ARM-JeOS.aarch64-rootfs.aarch64.tbz"
+url_leap_42_3_aarch64="http://download.opensuse.org/ports/aarch64/distribution/leap/42.3/appliances/openSUSE-Leap42.3-ARM-JeOS.aarch64-rootfs.aarch64.tbz"
 
 BOOTSTRAP_ARCH="$ARCH"
 BOOTSTRAP_RELEASE="$RELEASE"
--- a/installer/opensuse/bootstrap
+++ b/installer/opensuse/bootstrap
@@ -1,0 +1,45 @@
+#!/bin/sh -e
+# Copyright (c) 2016 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# This is a distro-specific bootstrap script, sourced from main.sh, and as such
+# has access to all of the variables set by main.sh, namely $tmp (the temporary
+# directory), $INSTALLERDIR/$DISTRO, $RELEASE, $BOOTSTRAP_RELEASE (if different
+# from $RELEASE), $ARCH, and $MIRROR.
+
+# Tumbleweed
+url_tumbleweed_armv7hl="http://download.opensuse.org/ports/armv7hl/tumbleweed/images/openSUSE-Tumbleweed-ARM-JeOS.armv7hl-rootfs.armv7l-Current.tbz"
+url_tumbleweed_aarch64="http://download.opensuse.org/ports/aarch64/tumbleweed/images/openSUSE-Tumbleweed-ARM-JeOS.aarch64-rootfs.aarch64-Current.tbz"
+
+# Leap
+url_leap_42_2_armv7hl="http://download.opensuse.org/ports/armv7hl/distribution/leap/42.2/appliances/openSUSE-Leap42.2-ARM-JeOS.armv7-rootfs.armv7l.tbz"
+url_leap_42_2_aarch64="http://download.opensuse.org/ports/aarch64/distribution/leap/42.2/appliances/openSUSE-Leap42.2-ARM-JeOS.aarch64-rootfs.aarch64.tbz"
+
+BOOTSTRAP_ARCH="$ARCH"
+BOOTSTRAP_RELEASE="$RELEASE"
+
+# Rename factory to tumbleweed
+if [ "$RELEASE" = "factory" ]; then
+    BOOTSTRAP_RELEASE=tumbleweed
+elif [ "${RELEASE#leap}" != "$RELEASE" ]; then
+     BOOTSTRAP_RELEASE=$(echo "$RELEASE" | tr - _ | tr . _)
+     echo $BOOTSTRAP_RELEASE
+fi
+
+# fix up the good old toolchain/kernel naming conflict
+if [ "$ARCH" = "arm64" ]; then
+     BOOTSTRAP_ARCH=aarch64
+fi
+
+eval "URL=\"\${url_${BOOTSTRAP_RELEASE}_${BOOTSTRAP_ARCH}-}\""
+
+if [ -z "$URL" ]; then
+    echo "Unknown Distribution / Architecture: $BOOTSTRAP_RELEASE $BOOTSTRAP_ARCH" >&2
+    exit 1
+fi
+
+mkdir -p "$tmp/$subdir"
+curl -# -L --connect-timeout 60 --retry 2 "$URL" | tar -jx -C "$tmp/$subdir"
+
+rm -rf "$tmp/$subdir/dev/"*
--- a/installer/opensuse/defaults
+++ b/installer/opensuse/defaults
@@ -1,0 +1,18 @@
+#!/bin/sh -e
+# Copyright (c) 2016 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# This file is sourced from main.sh to update distro-specific defaults.
+# It must set at least ARCH and MIRROR if not already specified.
+
+if [ -z "$ARCH" ]; then
+    ARCH="$(uname -m)"
+fi
+
+case "$ARCH" in
+arm64 | aarch64) ARCH="arm64";;
+arm*) ARCH="armv7hl";;
+*) error 2 "Invalid architecture '$ARCH'.";;
+esac
+
--- a/installer/opensuse/getrelease.sh
+++ b/installer/opensuse/getrelease.sh
@@ -1,0 +1,50 @@
+#!/bin/sh -e
+# Copyright (c) 2017 The crouton authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+USAGE="${0##*/} -a|-r /path/to/chroot
+
+Detects the release (-r) or arch (-a) of the chroot and prints it on stdout.
+Fails with an error code of 1 if the chroot does not belong to this distro."
+
+if [ "$#" != 2 ] || [ "$1" != '-a' -a "$1" != '-r' ]; then
+    echo "$USAGE" 1>&2
+    exit 2
+fi
+
+sources="${2%/}/etc/os-release"
+if [ ! -s "$sources" ]; then
+    exit 1
+fi
+
+# Create release name from /etc/os-release file
+# sed specialist can make that more elegant...
+rel="$(grep CPE_NAME "$sources" | cut -d ":" -f 4)"
+if [ "$rel" = "opensuse" -o "$rel" = "tumbleweed" ]; then
+    rel=tumbleweed
+elif [ "$rel" = "leap" ]; then
+    ver="$(grep CPE_NAME "$sources" | cut -d ":" -f 5)"
+    rel="$rel-${ver%\"}"
+else
+    exit 1
+fi
+
+# Print the architecture if requested
+# Why use sed if you know cut...
+sources="${2%/}/etc/products.d/openSUSE.prod"
+if [ ! -s "$sources" ]; then
+    exit 1
+fi
+
+if [ "$1" = '-a' ]; then
+    arch="$(grep "<arch>" "$sources" | cut -d '>' -f 2 | cut -d '<' -f 1)"
+    if [ "$arch" = "aarch64" ]; then
+        arch="arm64"
+    fi
+    echo "$arch"
+else
+    echo "$rel"
+fi
+
+exit 0
--- a/installer/opensuse/prepare
+++ b/installer/opensuse/prepare
@@ -1,0 +1,130 @@
+#!/bin/sh -e
+# Copyright (c) 2016 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# This is a distro-specific continuation of the prepare.sh script.
+
+PKGEXT='rpm'
+DISTROAKA='opensuse'
+
+
+# install_dist: see install() in prepare.sh for details.
+install_dist() {
+    local pkgs='' params='' asdeps=''
+    while [ "$#" != 0 ]; do
+        if [ "$1" = "--minimal" ]; then
+            params='--no-recommends'
+        elif [ "$1" = "--asdeps" ]; then
+            shift
+            continue
+        else
+            break
+        fi
+        shift
+    done
+    while [ "$#" != 0 ]; do
+        if [ "$1" = '--' ]; then
+            shift
+            break
+        fi
+        pkgs="$pkgs $1"
+        shift
+    done
+    zypper -q install -y $params $pkgs $(list_uninstalled_dist - "$@") || return $?
+}
+
+
+# install_pkg_dist: see install_pkg() in prepare.sh for details.
+install_pkg_dist() {
+    local params=''
+    if [ "$1" = '--minimal' ]; then
+        params='--no-recommends'
+        shift
+    fi
+    if [ "$#" != 0 ]; then
+        zypper -q install -y $params "$@" || return $?
+    fi
+}
+
+
+# install_dummy_dist: see install_dummy() in prepare.sh for details.
+install_dummy_dist() {
+    if [ "$#" = 0 ]; then
+        return
+    fi
+    local pkgname="crouton-$1" pkgprovides="$1" requires=""
+    shift
+    while [ "$#" != 0 ]; do
+        if [ "$1" = '--' ]; then
+            shift
+            break
+        fi
+        pkgprovides="$pkgprovides, $1"
+        shift
+    done
+    local pkgdepends="$1"
+    if [ "$#" != 0 ]; then
+        shift
+        while [ "$#" != 0 ]; do
+            pkgdepends="$pkgdepends, $1"
+            shift
+        done
+    fi
+
+    if [ -n "$pkgdepends" ]; then
+        requires="Requires: ${pkgdepends}"
+    fi
+
+    local tmp="$(mktemp -d crouton.XXXXXX --tmpdir=/tmp)"
+    addtrap "rm -rf '$tmp'"
+    cat > "$tmp/dummy.spec" <<EOF
+Name: $pkgname
+Version: 1.0.0
+Release: 1.0.0
+Group: dummy
+Vendor: dummy
+License: %{vendor}
+Provides: $pkgprovides
+$requires
+Summary: Provides a dummy ${pkgname#*-} for crouton
+BuildRoot: %{_tmppath}/%{name}-%{version}-build
+
+%description
+%{summary}
+
+%files
+EOF
+    rpmbuild --define='_rpmdir "$tmp"' -bb "$tmp/dummy.spec" > "$tmp/build.log"
+    if [ $? != 0 ]; then
+        echo "ERROR: Could not build dummy rpm!"
+	exit 1
+    fi
+    pkg="$(awk '/^wrote:/ { print $2 }' < "$tmp/build.log")"
+    rm "$tmp/build.log"
+    zypper -q install -y --force-resolution "$pkg"
+}
+
+
+# remove_dist: see remove() in prepare.sh for details.
+remove_dist() {
+    zypper -q remove --clean-deps -y "$@"
+}
+
+
+# list_uninstalled_dist: see list_uninstalled() in prepare.sh for details.
+list_uninstalled_dist() {
+    local suffix="$1" pkg
+    shift
+    for pkg in "$@"; do
+        if ! zypper se -i "$pkg" 2>/dev/null | grep -q '^i'; then
+            echo -n "$pkg$suffix "
+        fi
+    done
+}
+
+# enable network ping for all users
+chmod u+s /usr/bin/ping
+
+# add rpmbuild for dummy packages
+install rpmbuild
--- a/installer/opensuse/releases
+++ b/installer/opensuse/releases
@@ -1,2 +1,3 @@
 tumbleweed
 leap-42.2
+leap-42.3
--- a/installer/opensuse/releases
+++ b/installer/opensuse/releases
@@ -1,0 +1,2 @@
+tumbleweed
+leap-42.2
--- a/targets/audio
+++ b/targets/audio
@@ -85,8 +85,8 @@ build_cras() {
     install --minimal alsa-utils \
         libasound2$pkgsuffix libspeexdsp1$pkgsuffix
 
-    install --minimal --asdeps gcc $pkgdepextra libc6-dev$pkgsuffix \
-        pkg-config libspeexdsp-dev$pkgsuffix
+    install --minimal --asdeps gcc $pkgdepextra opensuse=glibc-devel,libc6-dev$pkgsuffix \
+        pkg-config opensuse=speexdsp-devel,libspeexdsp-dev$pkgsuffix
 
     # precise does not allow libasound2-dev and libasound2-dev:i386 to be
     # installed simultaneously
@@ -99,6 +99,13 @@ build_cras() {
             ln -sfT libasound.so.2 "$libasoundso"
         fi
         ALSALIBDIR="/usr/lib$archextrapath/alsa-lib"
+    elif [ "$DISTROAKA" = "opensuse" ]; then
+        install --minimal --asdeps alsa-devel
+        if [ "$ARCH" = "arm64" -o "$ARCH" = "x86_64"]; then
+            ALSALIBDIR="/usr/lib64/alsa-lib"
+        else
+            ALSALIBDIR="/usr/lib/alsa-lib"
+        fi
     else
         install --minimal --asdeps libasound2-dev$pkgsuffix
         ALSALIBDIR="`PKG_CONFIG_PATH="$pkgconfigpath" \
--- a/targets/chromium
+++ b/targets/chromium
@@ -5,7 +5,7 @@
 if [ "${TARGETNOINSTALL:-c}" = 'c' ] &&
         [ "$DISTRO" = 'debian' -o "$DISTRO" = 'kali' ] &&
         [ "${ARCH#arm}" != "$ARCH" ]; then
-    error 99 "chromium target is not supported on Debian/ARM."
+    error 99 "chromium target is not supported on $DISTRO on ARM."
 fi
 REQUIRES='x11'
 DESCRIPTION="Chromium browser. Uses the distro's version, which may be old."
--- a/targets/chromium
+++ b/targets/chromium
@@ -3,7 +3,7 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 if [ "${TARGETNOINSTALL:-c}" = 'c' ] &&
-        [ "$DISTRO" = 'debian' -o "$DISTRO" = 'kali' ] &&
+        [ "$DISTRO" = 'debian' -o "$DISTRO" = 'kali' -o "$DISTROAKA" = 'opensuse' ] &&
         [ "${ARCH#arm}" != "$ARCH" ]; then
     error 99 "chromium target is not supported on $DISTRO on ARM."
 fi
--- a/targets/gtk-extra
+++ b/targets/gtk-extra
@@ -8,17 +8,19 @@ DESCRIPTION='GTK-based tools including gdebi, gksu, and a simple browser.'
 
 ### Append to prepare.sh:
 install_dummy network-manager network-manager-gnome
-install gdebi gksu
+install opensuse=,gdebi opensuse=libgnomesu,gksu
 
-for BROWSER in netsurf-gtk dillo hv3 ""; do
-    test -n "$BROWSER"
-    if install "$BROWSER"; then
-        bin="/usr/bin/$BROWSER"
-        for link in x-www-browser gnome-www-browser; do
-            if ! update-alternatives --query "$link" | grep -q "$bin"; then
-                update-alternatives --install "/usr/bin/$link" "$link" "$bin" 10
-            fi
-        done
-        break
-    fi
-done
+if [ "$DISTROAKA" != "opensuse" ]; then
+    for BROWSER in netsurf-gtk dillo hv3 ""; do
+        test -n "$BROWSER"
+        if install "$BROWSER"; then
+            bin="/usr/bin/$BROWSER"
+            for link in x-www-browser gnome-www-browser; do
+                if ! update-alternatives --query "$link" | grep -q "$bin"; then
+                    update-alternatives --install "/usr/bin/$link" "$link" "$bin" 10
+                fi
+            done
+            break
+        fi
+    done
+fi
--- a/targets/post-common
+++ b/targets/post-common
@@ -34,8 +34,15 @@ if [ "${DISTROAKA:-"$DISTRO"}" = 'debian' ]; then
     apt-get clean
 fi
 
+groups='audio,input,video,wayland,plugdev,crouton'
+
 # Add the primary user
-groups='audio,input,video,wayland,sudo,plugdev,crouton'
+if [ "${DISTROAKA:-"$DISTRO"}" = 'opensuse' ]; then
+    groups=$groups',root'
+else
+    groups=$groups',sudo'
+fi
+
 if ! grep -q ':1000:' /etc/passwd; then
     while ! echo "$username" | grep -q '^[a-z][-a-z0-9_]*$'; do
         if [ -n "$username" ]; then
--- a/targets/x11-common
+++ b/targets/x11-common
@@ -23,16 +23,16 @@ if [ -n "$offset" ]; then
 fi
 
 # Install utilities and links for powerd-poking daemon
-compile xi2event '-lX11 -lXi' libx11-dev libxi-dev
-install --minimal dbus xdg-utils
+compile xi2event '-lX11 -lXi' opensuse=libX11-devel,libx11-dev opensuse=libXi-devel,libxi-dev
+install --minimal opensuse=dbus-1,dbus xdg-utils
 ln -sf croutonpowerd /usr/local/bin/gnome-screensaver-command
 ln -sf croutonpowerd /usr/local/bin/xscreensaver-command
 
 # Install nicer cursors
-install --minimal dmz-cursor-theme
+install --minimal opensuse=dmz-icon-theme-cursors,dmz-cursor-theme
 
 # Install bsdmainutils, xbindkeys and xautomation for shortcuts
-install --minimal bsdmainutils xbindkeys xautomation
+install --minimal opensuse=,bsdmainutils xbindkeys xautomation
 
 # We need chvt on freon systems
 if [ -f "/sys/class/tty/tty0/active" ]; then
--- a/targets/xiwi
+++ b/targets/xiwi
@@ -28,7 +28,8 @@ if [ "${DISTROAKA:-"$DISTRO"}" = 'debian' ]; then
     apt-mark unhold xserver-xorg-video-dummy$ltspackages || true 2>/dev/null
 fi
 
-install xorg xserver-xorg-video-dummy$ltspackages
+install opensuse=xinit, opensuse=xorg-x11-server,xorg \
+    opensuse=xf86-video-dummy,xserver-xorg-video-dummy$ltspackages
 install --minimal i3
 
 # Remove some unsupported options on old versions of i3
@@ -38,7 +39,8 @@ fi
 
 # Compile croutonfbserver
 compile fbserver '-lX11 -lXfixes -lXdamage -lXext -lXtst' \
-    libx11-dev libxfixes-dev libxdamage-dev libxext-dev libxtst-dev
+    opensuse=libX11-devel,libx11-dev opensuse=libXfixes-devel,libxfixes-dev opensuse=libXdamage-devel,libxdamage-dev \
+    opensuse=libXext-devel,libxext-dev opensuse=libXtst-devel,libxtst-dev
 compile findnacld ''
 
 ln -sf /etc/crouton/xorg-dummy.conf /etc/X11/
@@ -60,8 +62,8 @@ fi
 wget $superinsecure -O "$DUMMYBUILDTMP/dummy.tar.gz" \
      "$urlbase/xf86-video-dummy-0.3.8.tar.gz"
 
-install --minimal --asdeps patch gcc libc-dev pkg-config \
-    xserver-xorg-dev$ltspackages x11proto-xf86dga-dev
+install --minimal --asdeps patch gcc opensuse=glibc-devel,libc-dev pkg-config \
+    opensuse=xorg-x11-server-sdk,xserver-xorg-dev$ltspackages opensuse=xf86dgaproto-devel,x11proto-xf86dga-dev
 
 (
     cd "$DUMMYBUILDTMP"
@@ -398,9 +400,15 @@ EOF
     echo "Installing Xorg dummy driver..." 1>&2
 
     DRIVERDIR="/usr/lib/xorg/modules/drivers"
+    if [ "${DISTROAKA:-"$DISTRO"}" = 'opensuse' ]; then
+        if [ "${ARCH}" = "arm64" -o "${ARCH}" = "x86_64" ]; then
+            DRIVERDIR="/usr/lib64/xorg/modules/drivers"
+        fi
+    fi
     mkdir -p "$DRIVERDIR/"
     /usr/bin/install -s dummy_drv.so "$DRIVERDIR/"
 ) # End compilation subshell
+
 
 if [ "${DISTROAKA:-"$DISTRO"}" = 'debian' ]; then
     # Hold xserver-xorg-video-dummy to make sure the driver does not get erased
--- a/targets/xorg
+++ b/targets/xorg
@@ -68,7 +68,7 @@ if [ "${ARCH#arm}" = "$ARCH" ]; then
 fi
 
 # Catalog relevant and irrelevant video drivers
-fbdev="xserver-xorg-video-fbdev$backport"
+fbdev="opensuse=xf86-video-fbdev,xserver-xorg-video-fbdev$backport"
 
 if release -lt sid -lt kali-rolling -lt vivid; then
     modesetting="xserver-xorg-video-modesetting$backport"
@@ -97,7 +97,7 @@ fi
 
 # On Freon, we need crazy xorg hacks
 if [ -n "$freon" ]; then
-    compile freon '-ldl -ldrm -I/usr/include/libdrm' so libdrm-dev
+    compile freon '-ldl -ldrm -I/usr/include/libdrm' so opensuse=libdrm-devel,libdrm-dev
 fi
 
 # Pin precise's version of mesa if necessary
@@ -183,6 +183,8 @@ fi
 
 if release -lt stretch -lt kali-rolling -lt yakkety; then
     vmmouse='xserver-xorg-input-vmmouse'
+elif [ "$DISTROAKA" = "opensuse" ]; then
+    vmmouse='xorg-x11-driver-input'
 else
     # xserver-xorg-input-vmmouse is obsolete
     vmmouse=''
@@ -194,11 +196,12 @@ if [ -z "$inteldriver" -a -n "$backport" ] && release -eq precise; then
     # so replace it with something not broken
     install_dummy xorg -- xserver-xorg$backport libgl1-mesa-glx$backport \
         libgl1-mesa-dri$backport libglu1-mesa xfonts-base x11-apps \
-        x11-session-utils x11-utils x11-xfs-utils x11-xkb-utils \
+        x11-session-utils opensuse=xprop,x11-utils x11-xfs-utils x11-xkb-utils \
         x11-xserver-utils xauth xinit xfonts-utils xkb-data xorg-docs-core \
         xterm x11-common xinput
 else
-    install xorg $installvideodrivers -- xserver-xorg-video-all$backport \
+    install opensuse=xorg-x11-server,xorg $installvideodrivers \
+        -- opensuse=xorg-x11-driver-video,xserver-xorg-video-all$backport \
         $vmmouse
 fi
 
--- a/host-bin/enter-chroot
+++ b/host-bin/enter-chroot
@@ -676,11 +676,18 @@ if [ "$NOLOGIN" != 1 ] && grep -q '^root:' "$passwd" 2>/dev/null; then
     fi
 
     # Launch systemd-logind if available and not already running
-    # Whitelisted for saucy and trusty
+    # Whitelisted for saucy, trusty and tumbleweed
     systemd_dir="`fixabslinks '/run/systemd'`"
-    if [ -x "$CHROOT/lib/systemd/systemd-logind" ] && \
+    systemd_logind="/lib/systemd/systemd-logind"
+    # openSUSE has a different systemd folder
+    if [ "$CHROOTRELEASE" = 'tumbleweed' ]; then
+        systemd_logind="/usr/lib/systemd/systemd-logind"
+    fi
+
+    if [ -x "$CHROOT$systemd_logind" ] && \
             [ ! -d "$systemd_dir" ] && \
-            [ "$CHROOTRELEASE" = 'saucy' -o "$CHROOTRELEASE" = 'trusty' ]; then
+            [ "$CHROOTRELEASE" = 'saucy' -o "$CHROOTRELEASE" = 'trusty' \
+            -o "$CHROOTRELEASE" = 'tumbleweed' ]; then
         # Every piece of systemd code ever assumes that this directory exists
         mkdir -p "$systemd_dir"
 
@@ -691,7 +698,7 @@ if [ "$NOLOGIN" != 1 ] && grep -q '^root:' "$passwd" 2>/dev/null; then
                   "$CHROOT/sys/fs/cgroup/systemd"
         fi
         # systemd-logind doesn't fork
-        chrootcmd "/lib/systemd/systemd-logind >/dev/null 2>&1 </dev/null &"
+        chrootcmd "${systemd_logind} >/dev/null 2>&1 </dev/null &"
     fi
 
     chrootcmd "/usr/local/bin/croutonfindnacld >/dev/null 2>&1 </dev/null &"
--- a/installer/prepare.sh
+++ b/installer/prepare.sh
@@ -199,7 +199,7 @@ compile() {
     fi
     shift 2
     echo "Installing dependencies for $out..." 1>&2
-    local pkgs="gcc libc6-dev $*"
+    local pkgs="gcc opensuse=glibc-devel,libc6-dev $*"
     install --minimal --asdeps $pkgs </dev/null
     echo "Compiling $out..." 1>&2
     local tmp="`mktemp crouton.XXXXXX --tmpdir=/tmp`"
--- a/targets/cli-extra
+++ b/targets/cli-extra
@@ -9,7 +9,7 @@ CHROOTBIN='croutonpowerd'
 . "${TARGETSDIR:="$PWD"}/common"
 
 ### Append to prepare.sh:
-install dbus openssh-client kbd
+install opensuse=dbus-1,dbus opensuse=openssh,openssh-client kbd
 
 TIPS="$TIPS
 You can start a shell in a new VT via the startcli host command: sudo startcli
--- a/targets/core
+++ b/targets/core
@@ -110,7 +110,7 @@ fi
 fixkeyboardmode
 
 # Install critical packages
-install --minimal sudo wget ca-certificates apt-transport-https
+install --minimal sudo wget ca-certificates opensuse=,apt-transport-https
 
 # Generate and set default locale
 if [ ! -f '/etc/default/locale' ] && hash locale-gen 2>/dev/null; then
--- a/targets/extension
+++ b/targets/extension
@@ -19,7 +19,7 @@ fi
 . "${TARGETSDIR:="$PWD"}/common"
 
 ### Append to prepare.sh:
-install x11-utils xsel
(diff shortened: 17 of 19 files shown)


installer/openSUSE/defaults
```diff
@@ -0,0 +1,23 @@
...
+
+case "$ARCH" in
+x86_64 | amd64) ARCH="x86_64";;
+arm*) ARCH="armv7l";;
+aarch64) ARCH="aarch64";;
+*) error 2 "Invalid architecture '$ARCH'.";;
+esac
+
+if [ -z "$MIRROR" ]; then
+    MIRROR="${CROUTON_MIRROR_opensuse_arm:-http://download.opensuse.org/ports/}"
```
dnschneid: Is this the right mirror for x86?

installer/openSUSE/defaults
```diff
@@ -0,0 +1,23 @@
...
+# This file is sourced from main.sh to update distro-specific defaults.
+# It must set at least ARCH and MIRROR if not already specified.
+
+if [ -z "$ARCH" ]; then
+    ARCH="`uname -m`"
+fi
+
+case "$ARCH" in
+x86_64 | amd64) ARCH="x86_64";;
+arm*) ARCH="armv7l";;
```
dnschneid: Should this be armv7hl?

installer/openSUSE/getrelease.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/sh -e
+# Copyright (c) 2017 Matthias Brugger. All rights reserved.
```
dnschneid: Please release all code to be merged as copyright the crouton authors.

installer/openSUSE/getrelease.sh
```diff
@@ -0,0 +1,47 @@
...
+
+sources="${2%/}/usr/lib/os-release"
+if [ ! -s "$sources" ]; then
+    exit 1
+fi
+
+# Create release name from /etc/os-release file
+# sed specialist can make that more elegant...
+rel="`grep CPE_NAME "$sources" | cut -d ":" -f 4`"
+if [ $rel = "opensuse" ]
```
dnschneid: `if [ "$rel" = 'opensuse' ]; then`

installer/openSUSE/getrelease.sh
```diff
@@ -0,0 +1,47 @@
...
+then
+    rel=tumbleweed
+elif [ $rel = "leap" ]
+then
+    ver=`grep CPE_NAME "$sources" | cut -d ":" -f 5 | sed 's/"$//'`
```
dnschneid: quote and `$( )`

installer/openSUSE/getrelease.sh
```diff
@@ -0,0 +1,47 @@
...
+    exit 2
+fi
+
+sources="${2%/}/usr/lib/os-release"
+if [ ! -s "$sources" ]; then
+    exit 1
+fi
+
+# Create release name from /etc/os-release file
+# sed specialist can make that more elegant...
```
dnschneid: @drinkcat :)

installer/openSUSE/getrelease.sh
```diff
@@ -0,0 +1,47 @@
...
+if [ $rel = "opensuse" ]
+then
+    rel=tumbleweed
+elif [ $rel = "leap" ]
+then
+    ver=`grep CPE_NAME "$sources" | cut -d ":" -f 5 | sed 's/"$//'`
+    rel=${rel}-${ver}
+elif [ $rel = "tumbleweed" ]
+then
+    continue
```
dnschneid: `continue` doesn't work to exit an if statement.  If you want a no-op, use `:`, or make the exit 1 conditional on `rel != tumbleweed`.

installer/openSUSE/getrelease.sh
```diff
@@ -0,0 +1,47 @@
...
+else
+    exit 1
+fi
+
+# Print the architecture if requested
+# Why use sed if you know cut...
+sources="${2%/}/etc/products.d/openSUSE.prod"
+if [ "$1" = '-a' ]; then
+    grep "<arch>" "$sources"| cut -d '>' -f 2 | cut -d '<' -f 1
```
dnschneid: nit: space around `|`

installer/openSUSE/getrelease.sh
```diff
@@ -0,0 +1,47 @@
...
+    rel=${rel}-${ver}
+elif [ $rel = "tumbleweed" ]
+then
+    continue
+else
+    exit 1
+fi
+
+# Print the architecture if requested
+# Why use sed if you know cut...
```
dnschneid: what, you don't find it satisfying to spend an hour forcing a three-stage pipeline into a single command? ;)

installer/openSUSE/prepare
```diff
@@ -0,0 +1,147 @@
+#!/bin/sh -e
+# Copyright (c) 2016 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# This is a distro-specific continuation of the prepare.sh script.
+
+PKGEXT='rpm'
+DISTROAKA='openSUSE'
```
dnschneid: let's keep non-user-visible references to openSUSE lowercased to avoid case-sensitivity issues.  That also includes the directory structure.

installer/openSUSE/prepare
```diff
@@ -0,0 +1,147 @@
...
+
+
+# install_dist: see install() in prepare.sh for details.
+install_dist() {
+    local pkgs='' pkgsasdeps='' params='' asdeps=''
+    while [ ! "$#" = 0 ]; do
```
dnschneid: `!=`...this is another thing we need to fix in the rest of the codebase :)

(here and elsewhere)

installer/openSUSE/prepare
```diff
@@ -0,0 +1,147 @@
...
+        fi
+        pkgs="$pkgs $1"
+       shift
+    done
+    if [ -n "$asdeps" ]; then
+        pkgsasdeps="`list_uninstalled_dist '' $pkgs`"
+    fi
+    zypper -q install -y $params $pkgs `list_uninstalled_dist - "$@"` || return $?
+# TODO see how we can implement that reasonable with zypper
+# for now we just bloat the chroot
```
dnschneid: but at least you're honest about it!

installer/openSUSE/prepare
```diff
@@ -0,0 +1,147 @@
...
+
+# install_pkg_dist: see install_pkg() in prepare.sh for details.
+install_pkg_dist() {
+    local params=''
+    if [ "$1" = '--minimal' ]; then
+        params='--no-recommends'
+        shift
+    fi
+    if [ ! "$#" = 0 ]; then
+#        rpm --nodeps -i "$@" || return $?
```
dnschneid: remove?

installer/openSUSE/prepare
```diff
@@ -0,0 +1,147 @@
...
+%{summary}
+
+%files
+EOF
+    rpmbuild --define='_rpmdir ${tmp}' -bb ${tmp}/dummy.spec > ${tmp}/build.log
+    if [ $? != 0 ]
+    then
+        echo "ERROR: Could not build dummy rpm!"
+    fi
+    pkg=$(awk '/^wrote:/ { print $2 }' < ${tmp}/build.log)
```
dnschneid: watch out for quotes

installer/openSUSE/prepare
```diff
@@ -0,0 +1,147 @@
...
+
+
+# remove_dist: see remove() in prepare.sh for details.
+remove_dist() {
+    zypper -q remove -y "$@"
+}
+
+
+# list_uninstalled_dist: see list_uninstalled() in prepare.sh for details.
+# If the package is virtual (e.g. libc-dev), we need to find the binary package
```
dnschneid: does this comment apply?

installer/openSUSE/prepare
```diff
@@ -0,0 +1,147 @@
...
+    done
+}
+
+if [ -r /debootstrap.sh ]; then
+    # Clean contents of /var/run (since we'll always be mounting over it).
+    rm -rf --one-file-system /var/run/*
+fi
+
+# enable network ping for all users
+sudo chmod u+s /usr/bin/ping
```
dnschneid: don't need sudo in chroot preparation; everything is already run as root

installer/openSUSE/releases
```diff
@@ -0,0 +1 @@
+tumbleweed
```
dnschneid: what about `leap`?

targets/audio
```diff
@@ -99,6 +99,13 @@ build_cras() {
             ln -sfT libasound.so.2 "$libasoundso"
         fi
         ALSALIBDIR="/usr/lib$archextrapath/alsa-lib"
+    elif release -le tumbleweed; then
+        install --minimal --asdeps alsa-devel
+        if [ $ARCH = "aarch64" ] || [ $ARCH = "x86_64"]; then
```
dnschneid: `] || [` -> `-o`

Also, $ARCH needs quotes.

targets/gtk-extra
```diff
@@ -8,17 +8,19 @@ DESCRIPTION='GTK-based tools including gdebi, gksu, and a simple browser.'
...
-        bin="/usr/bin/$BROWSER"
-        for link in x-www-browser gnome-www-browser; do
-            if ! update-alternatives --query "$link" | grep -q "$bin"; then
-                update-alternatives --install "/usr/bin/$link" "$link" "$bin" 10
-            fi
-        done
-        break
-    fi
-done
+if release -ne tumbleweed; then
```
dnschneid: that will actually match all openSUSE releases that aren't tumbleweed.

Instead, you want `if ! release -eq tumbleweed`, or `if [ "$DISTROAKA" != opensuse ]` or something.

dnschneid: Does opensuse not have an equivalent to alternatives?  How does it pick the most appropriate browser to use?

targets/post-common
```diff
@@ -35,7 +35,12 @@ if [ "${DISTROAKA:-"$DISTRO"}" = 'debian' ]; then
 fi
 
 # Add the primary user
-groups='audio,input,video,wayland,sudo,plugdev,crouton'
+if [ "${DISTROAKA:-"$DISTRO"}" = 'openSUSE' ]; then
+    groups='audio,input,video,wayland,root,plugdev,crouton'
```
dnschneid: instead of two mostly-similar lists, why not put the common list above and just add sudo/root to groups in the if statements

targets/x11-common
```diff
@@ -7,6 +7,11 @@
 # forms of X11.
 
 ### Append to prepare.sh:
+
+#we need to install xinit to be sure we overwrite the path
+if release -eq tumbleweed; then
```
dnschneid: anything that should happen on all opensuse releases should use the DISTROAKA check.

targets/xiwi
```diff
@@ -38,7 +38,8 @@ fi
 
 # Compile croutonfbserver
 compile fbserver '-lX11 -lXfixes -lXdamage -lXext -lXtst' \
-    libx11-dev libxfixes-dev libxdamage-dev libxext-dev libxtst-dev
+    openSUSE=libX11-devel,libx11-dev openSUSE=libXfixes-devel,libsXxfixes-dev openSUSE=libXdamage-devel,libxdamage-dev \
```
dnschneid: `libsXxfixes-dev` is mangled

targets/xiwi
```diff
@@ -402,7 +403,13 @@ EOF
     /usr/bin/install -s dummy_drv.so "$DRIVERDIR/"
 ) # End compilation subshell
 
-if [ "${DISTROAKA:-"$DISTRO"}" = 'debian' ]; then
+if [ "${DISTROAKA:-"$DISTRO"}" = 'openSUSE' ]; then
+    if [ $ARCH = "aarch64" ] || [ $ARCH = "x86_64"]; then
+        cp /usr/lib/xorg/modules/drivers/* /usr/lib64/xorg/modules/drivers/
```
dnschneid: Why is this necessary?

targets/xiwi
```diff
@@ -402,7 +403,13 @@ EOF
...
+    else
+        cp /usr/lib/xorg/modules/drivers/* /usr/lib/xorg/modules/drivers/
```
dnschneid: this line doesn't make sense

targets/xorg
```diff
@@ -43,7 +43,7 @@ inteldriver=''
 intelhasfbc=''
 pinmesa=''
 intelfbcsupport='y'
-if [ "${ARCH#arm}" = "$ARCH" ]; then
+if [ "${ARCH#arm}" = "$ARCH" ] && [ "${ARCH#aarch64}" = "$ARCH" ]; then
```
dnschneid: oh, this looks like a bug...could you make a separate PR for this?  I wonder if we've done similar mistakes elsewhere.

@drinkcat

dnschneid: ...or is it just that debian/ubuntu refers to it as arm64?

targets/xorg
```diff
@@ -68,7 +68,7 @@ if [ "${ARCH#arm}" = "$ARCH" ]; then
 fi
 
 # Catalog relevant and irrelevant video drivers
-fbdev="xserver-xorg-video-fbdev$backport"
+fbdev="openSUSE=xf86-video-fbdev,xserver-xorg-video-fbdev$backport"
```
dnschneid: I wonder if it would be cleaner to have each distro set the xorg package name prefix at the top and then just reference `$xorg-video-fbdev` (etc) without needing any distro modifiers

targets/xorg
```diff
@@ -97,7 +97,7 @@ fi
 
 # On Freon, we need crazy xorg hacks
 if [ -n "$freon" ]; then
-    compile freon '-ldl -ldrm -I/usr/include/libdrm' so libdrm-dev
+    compile freon '-ldl -ldrm -I/usr/include/libdrm' so openSUSE=libdrm-devel,libdrm-dev
```
dnschneid: same goes for devel vs. dev, I guess.  You don't have to make any changes, though.

targets/xorg
```diff
@@ -181,7 +181,7 @@ if [ -n "$backport" ]; then
         "xserver-xorg-input-synaptics$backport" $installvideodrivers
 fi
 
-if release -lt stretch -lt kali-rolling -lt yakkety; then
+if release -lt stretch -lt kali-rolling -lt yakkety -lt tumbleweed; then
```
dnschneid: `-lt tumbleweed` means any opensuse release that is earlier than tumbleweed, not including tumbleweed itself.  Is that what you meant?

dnschneid: note that if you don't want this for any opensuse release, you can just leave the line unmodified (without `-lt tumbleweed`) and it won't match opensuse.

targets/xorg
```diff
@@ -194,12 +194,12 @@ if [ -z "$inteldriver" -a -n "$backport" ] && release -eq precise; then
     # so replace it with something not broken
     install_dummy xorg -- xserver-xorg$backport libgl1-mesa-glx$backport \
         libgl1-mesa-dri$backport libglu1-mesa xfonts-base x11-apps \
-        x11-session-utils x11-utils x11-xfs-utils x11-xkb-utils \
+        x11-session-utils openSUSE=xprop,x11-utils x11-xfs-utils x11-xkb-utils \
```
dnschneid: is xprop the only thing we need from x11-utils?

dnschneid (changes requested):
This is a really great start; thanks!  Maybe this will beat gentoo and arch :)

The O of croagh!

installer/openSUSE/releases
mbgg: it's not supported at the moment, I'll evaluate if dropping the leap part for now.

targets/audio
mbgg: yes, I realized this tonight as well, most probably this is the reason why X is not working in targets/xiwi

targets/post-common
mbgg: sure

targets/xiwi
mbgg: uups

targets/xiwi
mbgg: ah, yes you are right :)

targets/xiwi
mbgg: ok, I see, I should change the DRIVEDIR variable accordingly. I just wonder if on ubuntu/debian etc the drivers are installed under /usr/lib/... although the architecture is amd64. Can you confirm? Otherwise we could get rid of the first if, checking for openSUSE

installer/openSUSE/getrelease.sh
DennisLfromGA: Note to @drinkcat: add 'sed specialist' to [LinkedIn page](https://www.linkedin.com/in/nicolas-boichat-721a2622/). :)

targets/xiwi
mbgg: It's for the Xorg dummy driver we install, all the other drivers should be fine.

targets/xorg
mbgg: Well in install/debian/defaults we set ARCH to arm64 while I did set it to aarch64. We should unify this. I tend to use aarch64, as arm64 is only a kernel thing, but I don't care too much honestly.

installer/openSUSE/debootstrap/debootstrap.sh
```diff
@@ -0,0 +1,122 @@
...
+# (at your option) any later version.
+#
+# It is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# A copy of the GNU General Public License can be downloaded from
+# http://www.gnu.org/licenses/gpl.html
```
mbgg: Ok, I'll see how to resolve this.

installer/openSUSE/defaults
mbgg: No, but actually MIRROR is not used anywhere. I'll delete it.

installer/openSUSE/prepare
mbgg: what do you mean exactly?

dnschneid: `pkg="$(awk '/^wrote:/ { print $2 }' < "$tmp/build.log")"`

targets/xiwi
dnschneid: @drinkcat what's the best way to make sure the driver ends up in the right place?

targets/xorg
dnschneid: let's be consistent with the other distros and use arm64

mbgg: ok, I'll fix installer/opensuse/getrelease.sh

targets/xorg
mbgg: I don't see any schema that would be easy to implement. Some have a prefix, some have a postfix and some have a different name.

targets/xorg
mbgg: yes, AFAIK it is the only tool needed

mbgg:
Just a friendly ping that I send a new pull request in here (not sure if I screwed up because we lost the comments on the old one). So any comments are welcome :)

installer/opensuse/debootstrap.sh
```diff
@@ -0,0 +1,76 @@
+#!/bin/sh -e
+# Copyright (c) 2017 The crouton Authors. All rights reserved.
```
dnschneid: You can't claim copyright just by deleting code, and most (all?) of this file appears to be copied from [here](https://github.com/openSUSE/debootstrap/blob/master/debootstrap.sh).

There's also no reason to fake the interface of debootstrap.  Please make an effort to implement opensuse/bootstrap without the use or reference of this code.  I suspect you can write what you need from scratch in a few lines inside opensuse/bootstrap.

mbgg:
I updated the pull request. I dropped debootstrap and implemented the necessary bits directly in bootstrap. Also I added two commits for chromium (un)support.

Happly awaiting any comments :)

installer/opensuse/bootstrap
```diff
@@ -0,0 +1,41 @@
...
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# This is a distro-specific bootstrap script, sourced from main.sh, and as such
+# has access to all of the variables set by main.sh, namely $tmp (the temporary
+# directory), $INSTALLERDIR/$DISTRO, $RELEASE, $BOOTSTRAP_RELEASE (if different
+# from $RELEASE), $ARCH, and $MIRROR.
+
+# Tumbleweed
+url_tumbleweed_armv7hl="http://download.opensuse.org/ports/armv7hl/tumbleweed/images/openSUSE-Tumbleweed-ARM-JeOS.armv7hl-rootfs.armv7l-Current.tbz"
```
dnschneid: how did you decide on these URLs?  The sizes are dramatically different (the X11 ones are 800 megs!).  There're also distribution/openSUSE-current/appliances/ and distribution/openSUSE-stable/appliances/ directories with similar images.

There are unversioned entries in the leap subdirectory.  Are those always updated?  Are we going to have to keep updating the tarball in this file if we specify a version?

And what happens on i386/x86_64?

installer/opensuse/bootstrap
```diff
@@ -0,0 +1,41 @@
...
+
+# Leap
+url_leap_42_2_armv7hl="http://download.opensuse.org/ports/armv7hl/distribution/leap/42.2/appliances/openSUSE-Leap42.2-ARM-X11.armv7-rootfs.armv7l-2017.02.02-Build1.1.tbz"
+url_leap_42_2_aarch64="http://download.opensuse.org/ports/aarch64/distribution/leap/42.2/appliances/openSUSE-Leap42.2-ARM-JeOS.aarch64-rootfs.aarch64-2017.02.02-Build1.1.tbz"
+
+BOOTSTRAP_ARCH="$ARCH"
+BOOTSTRAP_RELEASE="$RELEASE"
+
+# Rename factory to tumbleweed
+[ "$RELEASE" = "factory" ] && BOOTSTRAP_RELEASE=tumbleweed
```
dnschneid: please convert this to a full if statement

installer/opensuse/bootstrap
```diff
@@ -0,0 +1,41 @@
...
+# Fix leap release name
+[ "$RELEASE" = "leap-42.2" ] && BOOTSTRAP_RELEASE=leap_42_2
+# fix up the good old toolchain/kernel naming conflict
+[ "$ARCH" = "arm64" ] && BOOTSTRAP_ARCH=aarch64
+
+key=$(echo "\$url_${BOOTSTRAP_RELEASE}_${BOOTSTRAP_ARCH}")
+URL=$(echo $(eval echo "${key}"))
```
dnschneid: I see what you're trying to do here, but considering all the setup and checking necessary, is this really shorter than just using if statements to explicitly set the URL?

installer/opensuse/bootstrap
```diff
@@ -0,0 +1,41 @@
...
+
+if [ ! "$URL" ]; then
+    echo "Unknown Distribution / Architecture: $key"
+    exit 1
+fi
+
+mkdir -p $tmp/$subdir
```
dnschneid: quoting

installer/opensuse/bootstrap
```diff
@@ -0,0 +1,41 @@
...
+( cd $tmp/$subdir; wget -O - $URL | bzip2 -d | tar x ) 1>&2
```
dnschneid: quoting.  Also, `bzip2 -d | tar x` could just be `tar -jx`

installer/opensuse/bootstrap
```diff
@@ -0,0 +1,41 @@
...
+
+echo 'Remove /dev files as not needed'
```
dnschneid: I don't think the user needs to see this

installer/opensuse/defaults
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh -e
+# Copyright (c) 2016 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# This file is sourced from main.sh to update distro-specific defaults.
+# It must set at least ARCH and MIRROR if not already specified.
+
+if [ -z "$ARCH" ]; then
+    ARCH="`uname -m`"
```
dnschneid: use `$( )`

installer/opensuse/bootstrap
```diff
@@ -0,0 +1,41 @@
...
+# Fix leap release name
+[ "$RELEASE" = "leap-42.2" ] && BOOTSTRAP_RELEASE=leap_42_2
+# fix up the good old toolchain/kernel naming conflict
+[ "$ARCH" = "arm64" ] && BOOTSTRAP_ARCH=aarch64
+
+key=$(echo "\$url_${BOOTSTRAP_RELEASE}_${BOOTSTRAP_ARCH}")
+URL=$(echo $(eval echo "${key}"))
+
+if [ ! "$URL" ]; then
+    echo "Unknown Distribution / Architecture: $key"
```
dnschneid: this should really be failing in `defaults`, not in bootstrap

installer/opensuse/getrelease.sh
```diff
@@ -0,0 +1,46 @@
...
+fi
+
+sources="${2%/}/etc/os-release"
+if [ ! -s "$sources" ]; then
+    exit 1
+fi
+
+# Create release name from /etc/os-release file
+# sed specialist can make that more elegant...
+rel=$(grep CPE_NAME "$sources" | cut -d ":" -f 4)
```
dnschneid: quoting

installer/opensuse/getrelease.sh
```diff
@@ -0,0 +1,46 @@
...
+if [ $rel = "opensuse" ]; then
```
dnschneid: quoting...

dnschneid: going to stop pointing it out here...but you need to make sure all your variable expansions are properly quoted

installer/opensuse/prepare
```diff
@@ -0,0 +1,142 @@
...
+            break
+        fi
+        pkgs="$pkgs $1"
+       shift
+    done
+    if [ -n "$asdeps" ]; then
+        pkgsasdeps="`list_uninstalled_dist '' $pkgs`"
+    fi
+    zypper -q install -y $params $pkgs `list_uninstalled_dist - "$@"` || return $?
+# TODO see how we can implement that reasonable with zypper
```
dnschneid: what's the plan here?

installer/opensuse/prepare
```diff
@@ -0,0 +1,142 @@
...
+        if [ "$1" = "--minimal" ]; then
+            params='--no-recommends'
+        elif [ "$1" = "--asdeps" ]; then
+            asdeps='y'
+        else
+            break
+        fi
+        shift
+    done
+    while [ ! "$#" = 0 ]; do
```
dnschneid: !=

installer/opensuse/prepare
```diff
@@ -0,0 +1,142 @@
...
+
+
+# install_pkg_dist: see install_pkg() in prepare.sh for details.
+install_pkg_dist() {
+    local params=''
+    if [ "$1" = '--minimal' ]; then
+        params='--no-recommends'
+        shift
+    fi
+    if [ ! "$#" = 0 ]; then
```
dnschneid: !=

installer/opensuse/prepare
```diff
@@ -0,0 +1,142 @@
...
+$requires
+Summary: Provides a dummy ${pkgname#*-} for crouton
+BuildRoot:      %{_tmppath}/%{name}-%{version}-build
+
+%description
+%{summary}
+
+%files
+EOF
+    rpmbuild --define='_rpmdir ${tmp}' -bb ${tmp}/dummy.spec > ${tmp}/build.log
```
dnschneid: quoting.  also, is single quotes correct for --define?

installer/opensuse/prepare
```diff
@@ -0,0 +1,142 @@
...
+    if [ $? != 0 ]
```
dnschneid: if [ ]; then

installer/opensuse/prepare
```diff
@@ -0,0 +1,142 @@
...
+    local suffix="$1" pkg
+    shift
+    for pkg in "$@"; do
+        if ! zypper se -i "$pkg" 2>/dev/null | grep -q '^i'; then
+            echo -n "$pkg$suffix "
+        fi
+    done
+}
+
+if [ -r /debootstrap.sh ]; then
```
dnschneid: not needed?

installer/opensuse/prepare
```diff
@@ -0,0 +1,142 @@
...
+    # Clean contents of /var/run (since we'll always be mounting over it).
+    rm -rf --one-file-system /var/run/*
+fi
+
+# enable network ping for all users
+chmod u+s /usr/bin/ping
+
+# add rpmbuild for dummy packages
+sudo zypper -q install -y --no-recommends rpmbuild
```
dnschneid: no need for sudo.  also, shouldn't this be an `install rpmbuild` command in core?

targets/chromium
```diff
@@ -3,9 +3,9 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 if [ "${TARGETNOINSTALL:-c}" = 'c' ] &&
-        [ "$DISTRO" = 'debian' -o "$DISTRO" = 'kali' ] &&
+        [ "$DISTRO" = 'debian' -o "$DISTRO" = 'kali' -o "$DISTRO" = 'opensuse' ] &&
         [ "${ARCH#arm}" != "$ARCH" ]; then
-    error 99 "chromium target is not supported on Debian/ARM."
+    error 99 "chromium target is not supported on "$DISTRO"/"$ARCH"."
```
dnschneid: this quoting isn't quite what you want

dnschneid: is it really not available though?  seems like it exists, although I have no idea what it involves: https://software.opensuse.org/package/chromium

targets/x11-common
```diff
@@ -7,6 +7,12 @@
 # forms of X11.
 
 ### Append to prepare.sh:
+
+#we need to install xinit to be sure we overwrite the path
```
dnschneid: what does this mean

installer/opensuse/bootstrap
mbgg: JeOS images are minimal images (Just Enough Operating System). For armv7hl no plain JeOS image exists. I'm investigating this. Maybe we can just delete the armv7hl support for Leap42.2 and add it in a later patch.

Leap images don't get updated, they get released and that's it. The unversioned images are links to the versioned ones. I'll update the links, as the version is of no use.

We should try to add a automatic resolution of architecutre + os-version -> download link, but I would prefer to leave this for a latter patch, e.g. when adding Leap-42.3 support. TBH I'm not too confident that this will work out, as the folder structure of the download server might change for newer releases.

i386/x86_64 images don't exists at this point in time, only the installer ISO is created.

installer/opensuse/bootstrap
mbgg: The idea behind that is, that if we can't be sure the download links keep the same folder structure over time, we just add a new identifier for, let's say leap-42.3 and we are fine.

This means that we need to convert "leap-42.3" to "leap_42_3" which is done in a dump way up to now. For the next revision I will change this using sed. So in effect if a new Leap release is added, we just need to put url_leap_42_3_aarch64 URL in the file and we are settled. This is still worse then other distributions in crouton, but IMHO better then start adding if statements of every new release. Thoughts?

installer/opensuse/bootstrap
mbgg: so as we can't support i386 right now, I will delete the architecutre from installer/opensuse/defaults as well.

mbgg: Ok, armv7hl image got created now, I will fix the URL.

installer/opensuse/getrelease.sh
mbgg: ok, hopefully I will catch all of them.

installer/opensuse/prepare
mbgg: There is none such functionality in zypper, I think we should get rid of all the asdeps stuff in this file. Thoughts?

dnschneid: Sure; the resulting chroot will just be larger because of it.

installer/opensuse/bootstrap
dnschneid: A new if statement (or a case statement) would probably be easier to follow, but maybe there's some complexity in that approach I'm not seeing.

mbgg: I think with this approach, adding a new version/architecture is easier, as it is a one-line add.
When you put that in if statements, it's more complicated, as you have to figure out between different if's, the arch and distribution name.

installer/opensuse/prepare
mbgg: yes, man page shows single quotes as well.

installer/opensuse/prepare
mbgg: no, not needed.

targets/chromium
mbgg: It's only available for x86, not for ARM.

mbgg: would something like this be acceptable for you:
    error 99 "chromium target is not supported on "$DISTRO"/ARM."

targets/x11-common
mbgg: TBH I can't remember, but we need to install xinit, otherwise graphic does not work.
Do you prefer that I add this line instead of the if:
install opensuse=xinit,

Just let me know.

targets/chromium
dnschneid: you don't need the quotes around $DISTRO, since the entire string is quoted

dnschneid: and I'd prefer "chromium target is not supported on $DISTRO on ARM"

targets/x11-common
dnschneid: I'd prefer to not install things unnecessarily, so please check what xinit is pulling in when you try to install it here

installer/opensuse/bootstrap
dnschneid: How about building up the URL in pieces?

targets/x11-common
mbgg: It's not pulling in anything, no recommends and sugguests:

linux@localhost:~> zypper info --recommends --suggests xinit
Loading repository data...
Reading installed packages...


Information for package xinit:
------------------------------
Repository     : openSUSE-Ports-Leap-42.3-repo-oss
Name           : xinit                            
Version        : 1.3.4-3.1                        
Arch           : aarch64                          
Vendor         : openSUSE                         
Installed Size : 234.6 KiB                        
Installed      : Yes (automatically)              
Status         : up-to-date                       
Source package : xinit-1.3.4-3.1.src              
Summary        : X Window System initializer      
Description    :                                  
    The xinit program is used to start the X Window System server and a
    first client program on systems that are not using a display manager
    such as xdm or in environments that use multiple window systems.
    When this first client exits, xinit will kill the X server and then
    terminate.
Recommends     : ---                              
Suggests       : ---

installer/opensuse/bootstrap
mbgg: Ok, I hacked that up for you, let me know if you prefer this, I definitely do not. File path and file names are not homogenous on architectures and releases. I expect this to break with a new architecture. Apart I'm quite sure that if you don't have the actual links in front of you, it is really hard to understand how this links will look like:

```bash
BOOTSTRAP_ARCH="$ARCH"
BOOTSTRAP_RELEASE="$RELEASE"
CURRENT=""

if [ "$ARCH" = "arm64" ] || [ "$ARCH" = "aarch64" ]; then
    BOOTSTRAP_ARCH="aarch64"
    BOOTSTRAP_ARCH2="aarch64"
    BOOTSTRAP_ARCH3="aarch64"
elif [ "$ARCH" = "armv7hl" ]; then
    BOOTSTRAP_ARCH="armv7hl"
    BOOTSTRAP_ARCH2="armv7hl"
    BOOTSTRAP_ARCH3="armv7l"
fi

if [ "$RELEASE" = "factory" ] || [ "$RELEASE" = "tumbleweed" ]; then
    BOOTSTRAP_VERSION="."
    BOOTSTRAP_RELEASE="tumbleweed/images"
    BOOTSTRAP_RELEASE2="."
    BOOTSTRAP_RELEASE3="Tumbleweed"
    CURRENT="-Current"
fi

case "$RELEASE" in leap*)
    BOOTSTRAP_VERSION=${RELEASE#leap-}
    BOOTSTRAP_RELEASE="distribution/leap"
    BOOTSTRAP_RELEASE2="appliances"
    BOOTSTRAP_RELEASE3="Leap$BOOTSTRAP_VERSION"

    if [ "$ARCH" = "armv7hl" ]; then
        BOOTSTRAP_ARCH2="armv7"
    fi
esac

URL="http://download.opensuse.org/ports/$BOOTSTRAP_ARCH/$BOOTSTRAP_RELEASE/$BOOTSTRAP_VERSION/$BOOTSTRAP_RELEASE2/openSUSE-$BOOTSTRAP_RELEASE3-ARM-JeOS.$BOOTSTRAP_ARCH2-rootfs.$BOOTSTRAP_ARCH3$CURRENT.tbz"
```

So for the record, here are the links we are talking about up to now:
http://download.opensuse.org/ports/armv7hl/tumbleweed/images/openSUSE-Tumbleweed-ARM-JeOS.armv7hl-rootfs.armv7l-Current.tbz
http://download.opensuse.org/ports/aarch64/tumbleweed/images/openSUSE-Tumbleweed-ARM-JeOS.aarch64-rootfs.aarch64-Current.tbz

http://download.opensuse.org/ports/armv7hl/distribution/leap/42.2/appliances/openSUSE-Leap42.2-ARM-JeOS.armv7-rootfs.armv7l.tbz
http://download.opensuse.org/ports/aarch64/distribution/leap/42.2/appliances/openSUSE-Leap42.2-ARM-JeOS.aarch64-rootfs.aarch64.tbz
http://download.opensuse.org/ports/aarch64/distribution/leap/42.3/appliances/openSUSE-Leap42.3-ARM-JeOS.aarch64-rootfs.aarch64.tbz

targets/chromium
mbgg: sure, no problem

installer/opensuse/bootstrap
dnschneid: Haha alright, I'm convinced.  Apologies for making you go through that.

targets/x11-common
dnschneid: Ah, I see.  This is fine, then.  Just move it under the "remove old vgem hack" section and quote $DISTROAKA

dnschneid: Oh weird, we're already installing xinit in the xorg.  Is this an issue for xorg or only for xiwi?

dnschneid: ...in which case, xiwi line 31 is probably the right place for it.

installer/opensuse/bootstrap
mbgg: No problem, the important thing is, that we reached consens on this :)

targets/x11-common
mbgg: ok, I put it at line 31 in xiwi.

installer/opensuse/bootstrap
```diff
@@ -0,0 +1,41 @@
...
+BOOTSTRAP_RELEASE="$RELEASE"
+
+# Rename factory to tumbleweed
+[ "$RELEASE" = "factory" ] && BOOTSTRAP_RELEASE=tumbleweed
+# Fix leap release name
+[ "$RELEASE" = "leap-42.2" ] && BOOTSTRAP_RELEASE=leap_42_2
+# fix up the good old toolchain/kernel naming conflict
+[ "$ARCH" = "arm64" ] && BOOTSTRAP_ARCH=aarch64
+
+key=$(echo "\$url_${BOOTSTRAP_RELEASE}_${BOOTSTRAP_ARCH}")
```
dnschneid: I *think* these two lines can be replaced with:

`eval "URL=\"\$url_${BOOTSTRAP_RELEASE}_${BOOTSTRAP_ARCH}\""`

dnschneid: or to handle the nonexistent case better:

`eval "URL=\"\${url_${BOOTSTRAP_RELEASE}_${BOOTSTRAP_ARCH}-}\""`

which just sets URL to an empty string.

installer/opensuse/bootstrap
```diff
@@ -0,0 +1,41 @@
...
+URL=$(echo $(eval echo "${key}"))
+
+if [ ! "$URL" ]; then
```
dnschneid: -z

installer/opensuse/bootstrap
dnschneid: wget is being removed from the rootfs, so use curl:

`curl -# -L --connect-timeout 60 --retry 2 "$URL" | tar -jx -C "$tmp/$subdir"`

installer/opensuse/bootstrap
mbgg: yes, will do so.

installer/opensuse/bootstrap
mbgg: done

mbgg:
@dnschneid , I think there was an overlap between you adding more comments on the code and me updating the pull request. I integrated your last two comments. Please feel free to review :)

installer/opensuse/bootstrap
```diff
@@ -0,0 +1,48 @@
...
+BOOTSTRAP_ARCH="$ARCH"
+BOOTSTRAP_RELEASE="$RELEASE"
+
+# Rename factory to tumbleweed
+if [ "$RELEASE" = "factory" ]; then
+    BOOTSTRAP_RELEASE=tumbleweed
+fi
+
+# Fix leap release name
+case "$RELEASE" in leap*)
```
dnschneid: elif [ "${RELEASE#leap}" != "$RELEASE" ]; then

installer/opensuse/bootstrap
```diff
@@ -0,0 +1,48 @@
...
+     BOOTSTRAP_RELEASE=$(echo "$RELEASE" | sed -e 's/[-.]/_/g')
```
dnschneid: quoting, and tr would probably be a better choice

installer/opensuse/bootstrap
```diff
@@ -0,0 +1,48 @@
...
+
+# fix up the good old toolchain/kernel naming conflict
+if [ "$ARCH" = "arm64" ]; then
+     BOOTSTRAP_ARCH=aarch64
+fi
+
+eval "URL=\"\${url_${BOOTSTRAP_RELEASE}_${BOOTSTRAP_ARCH}-}\""
+
+if [ -z "$URL" ]; then
+    echo "Unknown Distribution / Architecture: $key"
```
dnschneid: if the `error` function is available, you probably want to use that.  Otherwise, you want to redirect to STDERR using `>&2`

Also, `$key` doesn't exist anymore

installer/opensuse/bootstrap
```diff
@@ -0,0 +1,48 @@
...
+    exit 1
+fi
+
+mkdir -p "$tmp"/"$subdir"
```
dnschneid: `"$tmp/$subdir"`

installer/opensuse/bootstrap
```diff
@@ -0,0 +1,48 @@
...
+( cd "$tmp"/"$subdir"; wget -O - $URL |  tar -jx ) 1>&2
```
dnschneid: wget is being removed from the rootfs, so use curl:

`curl -# -L --connect-timeout 60 --retry 2 "$URL" | tar -jx -C "$tmp/$subdir"`

installer/opensuse/bootstrap
```diff
@@ -0,0 +1,48 @@
...
+
+rm -rf "$tmp"/"$subdir"/dev/*
```
dnschneid: `"$tmp/$subdir/dev/"*`

installer/opensuse/defaults
```diff
@@ -0,0 +1,18 @@
...
+# This file is sourced from main.sh to update distro-specific defaults.
+# It must set at least ARCH and MIRROR if not already specified.
+
+if [ -z "$ARCH" ]; then
+    ARCH="$(uname -m)"
+fi
+
+case "$ARCH" in
+arm*) ARCH="armv7hl";;
+arm64 | aarch64) ARCH="arm64";;
```
dnschneid: I think this needs to be above the arm* line, otherwise arm64 won't be matched properly

installer/opensuse/defaults
```diff
@@ -0,0 +1,18 @@
...
+*) error 2 "Invalid architecture '$ARCH'.";;
```
dnschneid: wait what happened to the x86 architectures

installer/opensuse/getrelease.sh
```diff
@@ -0,0 +1,46 @@
...
+fi
+
+# Create release name from /etc/os-release file
+# sed specialist can make that more elegant...
+rel="$(grep CPE_NAME "$sources" | cut -d ":" -f 4)"
+if [ "$rel" = "opensuse" ]; then
+    rel=tumbleweed
+elif [ "$rel" = "leap" ]; then
+    ver="$(grep CPE_NAME "$sources" | cut -d ":" -f 5 | sed 's/"$//')"
+    rel="$rel"-"$ver"
```
dnschneid:
```
ver="$(grep CPE_NAME "$sources" | cut -d ":" -f 5)"
rel="$rel-${ver%\"}"
```

installer/opensuse/getrelease.sh
```diff
@@ -0,0 +1,46 @@
...
+elif [ "$rel" != "tumbleweed" ]; then
```
dnschneid: make this "else" and add `-o "$rel" = 'tumbleweed'` to the first block

installer/opensuse/getrelease.sh
```diff
@@ -0,0 +1,46 @@
...
+    exit 1
+fi
+
+# Print the architecture if requested
+# Why use sed if you know cut...
+sources="${2%/}/etc/products.d/openSUSE.prod"
```
dnschneid: fail if this doesn't exist?

installer/opensuse/prepare
```diff
@@ -0,0 +1,129 @@
...
+
+# install_dist: see install() in prepare.sh for details.
+install_dist() {
+    local pkgs='' params='' asdeps=''
+    while [ "$#" != 0 ]; do
+        if [ "$1" = "--minimal" ]; then
+            params='--no-recommends'
+        elif [ "$1" = "--asdeps" ]; then
+            shift
+	    continue
```
dnschneid: indentation

installer/opensuse/prepare
```diff
@@ -0,0 +1,129 @@
...
+        fi
+        shift
+    done
+    while [ "$#" != 0 ]; do
+        if [ "$1" = '--' ]; then
+            shift
+            break
+        fi
+        pkgs="$pkgs $1"
+       shift
```
dnschneid: indentation

installer/opensuse/prepare
```diff
@@ -0,0 +1,129 @@
...
+    done
+    zypper -q install -y $params $pkgs `list_uninstalled_dist - "$@"` || return $?
```
dnschneid: use `$( )` instead of `` ` ` ``

installer/opensuse/prepare
```diff
@@ -0,0 +1,129 @@
...
+    local pkgdepends="$1"
+    if [ "$#" != 0 ]; then
+        shift
+        while [ "$#" != 0 ]; do
+            pkgdepends="$pkgdepends, $1"
+            shift
+        done
+    fi
+
+    if [ $pkgdepends ]; then
```
dnschneid: `if [ -n "$pkgdepends" ]; then`

installer/opensuse/prepare
```diff
@@ -0,0 +1,129 @@
...
+        requires="Requires: ${pkgdepends}"
+    fi
+
+    local tmp="`mktemp -d crouton.XXXXXX --tmpdir=/tmp`"
```
dnschneid: use `$( )`

installer/opensuse/prepare
```diff
@@ -0,0 +1,129 @@
...
+Name: $pkgname
+Version: 1.0.0
+Release: 1.0.0
+Group: dummy
+Vendor: dummy
+License: %{vendor}
+Provides: $pkgprovides
+$requires
+Summary: Provides a dummy ${pkgname#*-} for crouton
+BuildRoot:      %{_tmppath}/%{name}-%{version}-build
```
dnschneid: extra spaces unnecessary?

installer/opensuse/prepare
```diff
@@ -0,0 +1,129 @@
...
+
+%description
+%{summary}
+
+%files
+EOF
+    rpmbuild --define='_rpmdir "$tmp"' -bb "$tmp"/dummy.spec > "$tmp"/build.log
```
dnschneid: quotes should surround the whole string (e.g., `"$tmp/dummy.spec"`) as long as wildcards aren't involved

installer/opensuse/prepare
```diff
@@ -0,0 +1,129 @@
...
+    if [ $? != 0 ]; then
+        echo "ERROR: Could not build dummy rpm!"
```
dnschneid: non-fatal?

installer/opensuse/prepare
```diff
@@ -0,0 +1,129 @@
...
+    fi
+    pkg="$(awk '/^wrote:/ { print $2 }' < "$tmp/build.log")"
+    rm "$tmp"/build.log
```
dnschneid: quote surround string

installer/opensuse/releases
```diff
@@ -0,0 +1,3 @@
+tumbleweed
+leap-42.2
+leap-42.3
```
dnschneid: and we don't want just a "leap" release that grabs the latest version, right?

targets/audio
```diff
@@ -99,6 +99,13 @@ build_cras() {
             ln -sfT libasound.so.2 "$libasoundso"
         fi
         ALSALIBDIR="/usr/lib$archextrapath/alsa-lib"
+    elif [ "$DISTROAKA" = "opensuse" ]; then
+        install --minimal --asdeps alsa-devel
+        if [ "$ARCH" = "arm64" ] || [ "$ARCH" = "x86_64"]; then
```
dnschneid: `] || [` should be `-o`

targets/chromium
```diff
@@ -3,9 +3,9 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 if [ "${TARGETNOINSTALL:-c}" = 'c' ] &&
-        [ "$DISTRO" = 'debian' -o "$DISTRO" = 'kali' ] &&
+        [ "$DISTRO" = 'debian' -o "$DISTRO" = 'kali' -o "$DISTRO" = 'opensuse' ] &&
```
dnschneid: not distroaka?

targets/gtk-extra
```diff
@@ -8,17 +8,19 @@ DESCRIPTION='GTK-based tools including gdebi, gksu, and a simple browser.'
 
 ### Append to prepare.sh:
 install_dummy network-manager network-manager-gnome
-install gdebi gksu
+    install opensuse=,gdebi opensuse=libgnomesu,gksu
```
dnschneid: indentation

targets/xiwi
```diff
@@ -29,6 +29,13 @@ if [ "${DISTROAKA:-"$DISTRO"}" = 'debian' ]; then
 fi
 
 install xorg xserver-xorg-video-dummy$ltspackages
+
+# We need to install xinit to make graphics work
+if [ $DISTROAKA = "opensuse" ]; then
+    install xinit
```
dnschneid: Just add it as `opensuse=xinit,` in the install line below

targets/xorg
```diff
@@ -194,11 +196,11 @@ if [ -z "$inteldriver" -a -n "$backport" ] && release -eq precise; then
     # so replace it with something not broken
     install_dummy xorg -- xserver-xorg$backport libgl1-mesa-glx$backport \
         libgl1-mesa-dri$backport libglu1-mesa xfonts-base x11-apps \
-        x11-session-utils x11-utils x11-xfs-utils x11-xkb-utils \
+        x11-session-utils opensuse=xprop,x11-utils x11-xfs-utils x11-xkb-utils \
         x11-xserver-utils xauth xinit xfonts-utils xkb-data xorg-docs-core \
         xterm x11-common xinput
 else
-    install xorg $installvideodrivers -- xserver-xorg-video-all$backport \
+    install opensuse=xorg-x11-server,xorg $installvideodrivers -- opensuse=xorg-x11-driver-video,xserver-xorg-video-all$backport \
```
dnschneid: probably need to break the line

dnschneid (changes requested):
Lots of nits and a few minor errors, but no major concerns!  Woo!

installer/opensuse/bootstrap
mbgg: it seems, that no error function is available.

installer/opensuse/defaults
mbgg: good catch!

installer/opensuse/defaults
mbgg: I dropped it for now, as there does no JeOS image exist, but just a installer image. That's something for the future. Creating an x86 image in the open build service and use it.

installer/opensuse/releases
mbgg: no, these are independent releases.

installer/opensuse/prepare
mbgg: no

mbgg:
Hi @dnschneid just now the newest version. Feel free to comment :)

mbgg:
Gentle ping, would be nice if you find a moment to review :-)

mbgg:
any further comments on that?

installer/opensuse/bootstrap
```diff
@@ -0,0 +1,46 @@
...
+url_leap_42_3_aarch64="http://download.opensuse.org/ports/aarch64/distribution/leap/42.3/appliances/openSUSE-Leap42.3-ARM-JeOS.aarch64-rootfs.aarch64.tbz"
+
+BOOTSTRAP_ARCH="$ARCH"
+BOOTSTRAP_RELEASE="$RELEASE"
+
+# Rename factory to tumbleweed
+if [ "$RELEASE" = "factory" ]; then
+    BOOTSTRAP_RELEASE=tumbleweed
+elif [ "${RELEASE#leap}" != "$RELEASE" ]; then
+     BOOTSTRAP_RELEASE=$(echo "$RELEASE" | tr - _ | tr . _)
```
dnschneid: quoting (and line below)

installer/opensuse/defaults
dnschneid: definitely need x86 support before we can merge; this would confuse most of our users

installer/opensuse/prepare
```diff
@@ -0,0 +1,130 @@
...
+
+%description
+%{summary}
+
+%files
+EOF
+    rpmbuild --define='_rpmdir "$tmp"' -bb "$tmp/dummy.spec" > "$tmp/build.log"
+    if [ $? != 0 ]; then
+        echo "ERROR: Could not build dummy rpm!"
+	exit 1
```
dnschneid: spaces, not tabs

dnschneid (changes requested):
Sorry I'm so slow at looking at this, especially since you've done such a great job cleaning it up.

It would be helpful if you avoid squashing fixes back into old commits (maybe until it's approved for merge and then you can rebase and make it pretty), since that causes the Github review UI to drop most of the history and comments and results in me needing to do a full review every time you update.

Issue #2674, opened 2016-07-10
Hard freeze on Chromebook Flip after switching to XFCE after sleep


floe:
```
$ sudo edit-chroot -all
name: xenial
encrypted: no
Entering /mnt/stateful_partition/crouton/chroots/xenial...
crouton: version 1-20160627183059~master:e3c0a690
release: xenial
architecture: armhf
xmethod: xorg
targets: keyboard,extension,xfce-desktop
host: version 8350.38.0 (Official Build) beta-channel veyron_minnie 
kernel: Linux localhost 3.14.0 #1 SMP PREEMPT Tue Jun 21 21:43:55 PDT 2016 armv7l armv7l armv7l GNU/Linux
freon: yes
Not unmounting /mnt/stateful_partition/crouton/chroots/xenial as another instance is using it.
```
#### Please describe your issue:

After starting XFCE with `sudo startxfce4`, I can switch back and forth between ChromeOS and XFCE just fine, unless I suspend the device in between. In this case, attempting to switch to XFCE triggers a hard freeze, just the cursor on black screen. Need to hold power for 10s to hard-reset.
#### If known, describe the steps to reproduce the issue:
- Launch XFCE4 with `sudo startxfce4`.
- Switch to XFCE and back to ChromeOS.
- Suspend the device by closing the lid.
- Open it again.
- Switch to XFCE -> hard freeze.

floe:
Maybe related: https://github.com/dnschneid/crouton/issues/2576

floe:
Problem persists after updating to:
`host: version 8350.55.0 (Official Build) beta-channel veyron_minnie `

drinkcat:
Can't reproduce on recent images, wondering if the recent render node fixes helped. Please reopen if it happens again.

floe:
Please reopen, still happens with the following configuration:

```
chronos@localhost / $ sudo edit-chroot -all
name: xenial
encrypted: no
Entering /mnt/stateful_partition/crouton/chroots/xenial...
crouton: version 1-20161014161205~master:16fd55fd
release: xenial
architecture: armhf
xmethod: xorg
targets: keyboard,extension,xfce-desktop
host: version 8743.65.0 (Official Build) beta-channel veyron_minnie 
kernel: Linux localhost 3.14.0 #1 SMP PREEMPT Wed Oct 12 19:54:05 PDT 2016 armv7l armv7l armv7l GNU/Linux
freon: yes
[...]

Issue #3372, opened 2017-08-26
Debian Stretch gives a mysterious message.


Asternitix:
```
Please paste the output of the following command here: sudo edit-chroot -all
```
name: beta
encrypted: no
Entering /mnt/stateful_partition/crouton/chroots/beta...
crouton: version 1-20170825164243~master:6af640de
release: stretch
architecture: amd64
xmethod: xorg
targets: gnome-desktop,gnome,extension
host: version 9592.71.0 (Official Build) stable-channel peppy 
kernel: Linux localhost 3.8.11 #1 SMP Thu Jul 27 00:04:59 PDT 2017 x86_64 GNU/Linux
freon: yes
#### Please describe your issue:
When startgnome is run this error message is given:
Entering /mnt/stateful_partition/crouton/chroots/beta...
Unknown username "geoclue" in message bus configuration file

_XSERVTransmkdir: Owner of /tmp/.X11-unix should be set to root

X.Org X Server 1.19.2
Release Date: 2017-03-02
X Protocol Version 11, Revision 0
Build Operating System: Linux 4.9.0-3-amd64 x86_64 Debian
Current Operating System: Linux localhost 3.8.11 #1 SMP Thu Jul 27 00:04:59 PDT 2017 x86_64
Kernel command line: cros_secure console= loglevel=7 init=/sbin/init cros_secure oops=panic panic=-1 root=/dev/dm-0 rootwait ro dm_verity.error_behavior=3 dm_verity.max_bios=-1 dm_verity.dev_wait=1 dm="1 vroot none ro 1,0 2539520 verity payload=PARTUUID=84de0146-4580-ca4f-823c-989c25105b57/PARTNROFF=1 hashtree=PARTUUID=84de0146-4580-ca4f-823c-989c25105b57/PARTNROFF=1 hashstart=2539520 alg=sha1 root_hexdigest=678204cbba46b8ad7ee4f033691fbe11a833d104 salt=475c61b6682e9441f2e7ffe37e2cb2e38283084af1e2993424a01692ff6e2d65" noinitrd vt.global_cursor_default=0 kern_guid=84de0146-4580-ca4f-823c-989c25105b57 add_efi_memmap boot=local noresume noswap i915.modeset=1 tpm_tis.force=1 tpm_tis.interrupts=0 nmi_watchdog=panic,lapic iTCO_vendor_support.vendorsupport=3  
Build Date: 07 July 2017  06:14:06AM
[...]

Issue #3369, opened 2017-08-25
PSA - Beta Channel ChromeOS v61 breaks crouton without xiwi


zguithues:
Sorry I didn't save an output of `sudo edit-chroot -all`

My chromebook (Toshiba Chromebook2 swanky) which is in the beta channel, updated to ChromeOS v61 yesterday.  Immediately after, when launching a crouton that doesn't use xiwi, i got a blank screen.  The key combination to switch back to ChromeOS did not work, i was forced to hold down the power to turn off.

I tested with a fresh official crouton, I power washed, i tried other targets, nothing worked.  Only when i had the xiwi target installed and used the crouton extension was I able to actually use the chroot's x.

I switched to the Stable Channel (v60), and crouton worked per normal.   I switched to Dev Channel(v62), crouton also worked per normal.  After testing both, i returned to Beta Channel (v61), and had the same blank screen issue.

I tried looking through logs, but nothing jumped out at me.

I'm going to switch to the Dev channel for now, and re-setting up my chroots.  I just wanted to report in-case others experience it.  I'm not sure if it's device specific, or an issue with ChromeOS v61 globally.

If requested, prior to setting it back up under Dev Channel, i can switch back to Beta Channel and run `sudo edit-chroot -all` and post the output.

DennisLfromGA:
@zguithues,

I have a lulu, Dell Chromebook 13, that upgraded to version 61 yesterday -

> Google Chrome	61.0.3163.51 (Official Build) beta (64-bit)
Platform	9765.31.0 (Official Build) beta-channel lulu
Firmware Version	Google_Lulu.6301.136.57
ARC	4279895

I haven't seen any other reports of this yet, I launched three different desktops in 'xorg' and had no problem.
Mine is using a much older 3.14 kernel compared to yours on 4.4 so it may be it's device specific.
Since the issue is on M61 only for you and seems to be fixed in M62 I guess your workaround may be the best course of action. 

Sounds like it'll work itself out but thanx for reporting it since others may run into this and now have a way around it.

*Hope this helps,*
-**DennisL**

Asternitix:
@zguithues #3310's solution should fix the problem. Thanks for PSA in the header.

DennisLfromGA:
@Asternitix,

Glad you got it fixed.
With no mention of what release or desktop was having issues, it's just a shot in the dark.

-**DennisL**

rsriniv:
I too have Toshiba CB2 on the beta channel.

[...]

Issue #3310, opened 2017-07-05
can't startxfce4 (sid&stretch)  


swors:
```
chronos@localhost / $ sudo edit-chroot -all
name: kali-rolling
encrypted: no
Entering /mnt/stateful_partition/crouton/chroots/kali-rolling...
crouton: version 1-20170619181610~master:980f090f
release: kali-rolling
architecture: amd64
targets: core
host: version 9592.42.0 (Official Build) beta-channel cave 
kernel: Linux localhost 3.18.0-14861-g70c266fc4519 #1 SMP PREEMPT Thu Jun 29 22:03:53 PDT 2017 x86_64 GNU/Linux
freon: yes
Unmounting /mnt/stateful_partition/crouton/chroots/kali-rolling...
name: sid
encrypted: no
Entering /mnt/stateful_partition/crouton/chroots/sid...
crouton: version 1-20170619181610~master:980f090f
release: sid
architecture: amd64
xmethod: xorg
targets: xfce,core
host: version 9592.42.0 (Official Build) beta-channel cave 
kernel: Linux localhost 3.18.0-14861-g70c266fc4519 #1 SMP PREEMPT Thu Jun 29 22:03:53 PDT 2017 x86_64 GNU/Linux
freon: yes
Unmounting /mnt/stateful_partition/crouton/chroots/sid...
name: stretch
encrypted: no
Entering /mnt/stateful_partition/crouton/chroots/stretch...
crouton: version 1-20170619181610~master:980f090f
release: stretch
architecture: amd64
xmethod: xorg
targets: xfce,core
host: version 9592.42.0 (Official Build) beta-channel cave 
kernel: Linux localhost 3.18.0-14861-g70c266fc4519 #1 SMP PREEMPT Thu Jun 29 22:03:53 PDT 2017 x86_64 GNU/Linux
freon: yes
Not unmounting /mnt/stateful_partition/crouton/chroots/stretch as another instance is using it.
```

#### Please describe your issue:
Can't startxfce4 , get these message
chronos@localhost / $ sudo enter-chroot -n sid startxfce4
Entering /mnt/stateful_partition/crouton/chroots/sid...
/usr/bin/startxfce4: Starting X server

_XSERVTransmkdir: Owner of /tmp/.X11-unix should be set to root

X.Org X Server 1.19.3
Release Date: 2017-03-15
X Protocol Version 11, Revision 0
Build Operating System: Linux 3.16.0-4-amd64 x86_64 Debian
[...]

PR #3380, opened 2017-08-31
Fix xserver-xorg-legacy issue with dummy package

zguithues:
not sure how or why all my synced changes are showing up in this pull request...  I'm a beginner with git...

This pull request makes the suggested change to the x11-common target, but it's not working properly for me.  It seems to create a dummy package named `crouton-xserver-xorg-legacy`, which doesn't block the installation of the real `xserver-xorg-legacy` package which breaks x.

@dnschneid do you have any ideas?  Have i just uncovered an unrelated issue?

### DIFF
--- a/installer/ubuntu/prepare
+++ b/installer/ubuntu/prepare
@@ -86,9 +86,11 @@ Architecture: all
 Maintainer: crouton
 Installed-Size: 0
 Provides: $pkgprovides
+Conflicts: $pkgprovides
+Replaces: $pkgprovides
 Depends: $pkgdepends
 Section: misc
-Priority: optional
+Priority: Required
 Description: Provides a dummy ${pkgname#*-} for crouton
 EOF
     touch "$tmp/md5sums"
--- a/targets/x11-common
+++ b/targets/x11-common
@@ -7,6 +7,10 @@
 # forms of X11.
 
 ### Append to prepare.sh:
+
+# Create dummy xserver-xorg-legacy
+install_dummy xserver-xorg-legacy
+
 # Store and apply the X11 method
 ln -sfT '/etc/crouton/xserverrc' '/etc/X11/xinit/xserverrc'
 echo "$XMETHOD" > '/etc/crouton/xmethod'


dnschneid:
Hmm, maybe we need to add a `Conflicts:` line to the dummy package control.

As for the git stuff, best thing to do is use `git rebase` to rebase onto dnschneid/crouton's master, then force-push your branch again.

zguithues:
@dnschneid I do not understand what you mean "maybe we need to add a Conflicts: line to the dummy package control"
I dug around a bit, but nothing is jumping out at me...

I would assume that making a dummy package would install a package with the same name as requested.  Is it supposed to make one with the prefix "crouton"?

dnschneid:
Sorry for being unclear; I'm referring to adding a ["conflicts"](https://www.debian.org/doc/debian-policy/ch-relationships.html#s-conflicts) relationship to the [package control file](https://github.com/dnschneid/crouton/blob/master/installer/ubuntu/prepare#L88).  Basically, the dummy package is called "crouton-<whatever>" to avoid having a version fight with the actual distro's packages, but is marked as Provides (and perhaps Conflicts) with the packages listed (in this case, xserver-xorg-legacy) in order to replace them.

eyqs:
This probably fixes #2674, #2875, #3228, #3242, #3289, #3310, #3369, #3372, by the way.

zguithues:
Thanks @eyqs for linking all of the Issues, i hadn't thought to do that...  When i re-create this pull request from a rebased fork, I'll make sure to include them. ;)

@dnschneid ok, i think i understand...  Are you suggesting that i edit the prepare.sh file to also indicate that it conflicts with `$pkgprovides` in all cases, or should we add some kind of option to only add the conflict when indicated?

zguithues:
i've rebased the branch.  apparently i didn't need to open a new pull request, should be able to keep using this one.

zguithues:
ok @dnschneid I think this fixes it...

i edited the install_dummy_dist to add `Replaces` and `Conflicts`, and set priority to Required.  This makes the dummy package force an uninstall of the offending package.

This might be too big of a hammer, but it should help eliminate this stupid package.

Let me know your thoughts...

dnschneid:
That certainly looks like the right kind of thing to try.  I'll run it through some tests and see if it breaks anything unexpected.

zguithues:
the way it's setup now, if `xserver-xorg-legacy` is installed, it will force an uninstall of `crouton-xserver-xorg-legacy`.  While that will break xorg, an update of the chroot targets will reinstall the `crouton-xserver-xorg-legacy`; which will remove `xserver-xorg-legacy`.

I couldn't figure out a way to completely block the installation of `xserver-xorg-legacy`, but i figured i'd wait till if/when it reinstalling itself becomes an issue to upgrade this hammer to a sledgehammer.

also, this doesn't check what release is being installed, it just arbitrarily adds the `crouton-xserver-xorg-legacy` dummy package.  Since it's such a disruptive package, i don't see a strong reason to restrict only to problem releases; but i defer to your judgement.

dnschneid:
I see.  Well, I haven't found any unintended side-effects of the hammer approach.

The question is, if you install something like kali-linux-full, does it still install xserver-xorg-legacy at the expense of crouton-xserver-xorg-legacy, or is the requirement fulfilled and the only concern is if the user explicitly tries to install xserver-xorg-legacy?

We could try using apt preferences to make a hold on the xserver-xorg-legacy package.  xiwi does something similar for the dummy driver.

zguithues:
I tried installing kali-linux-full in a kali-rolling chroot.  It filled up the drive and crashed.
I removed all the chroots, tried again (8.2gb free)...  filled up the drive and crashed later in the process...  

I think it got past the xorg stuff (furthest it got was mysql stuff), so i assume it would have collided with the dummy package, but i'm not certain...

Do you think i should add `apt-mark hold crouton-xserver-xorg-legacy` to the X11-Common target?  or should we just wait and see?
Perhaps add a comment saying it may be needed later?

dnschneid:
Wait and see...merging this is long-overdue.

Thanks again for digging into this and producing a patch!

Issue #3222, opened 2017-05-04
Ctrl-w tab protection for xiwi


spiffytech:
When I have a xiwi tab open and hit `Ctrl-w` the tab gets closed.

Please add basic tab protection (prompt the browser's unsaved input "Are you sure you want to close this tab?"), or more preferable, create an app version of xiwi windows (similar to the Crosh Window app) that can capture and use `Ctrl-w`.

jeremylinlin:
I have the same issue here. Any shortcuts that are available in chromeOS won't work inside the xiwi tab. Some really annoying ones include ctrl+w, ctrl+alt+t.

I am thinking it would be the best to assign all key combinations to crouton if the xiwi tab is active. Otherwise some useful hotkeys will become useless in xiwi mode.

spiffytech:
I found a partial solution to this on [the cheatsheet page](https://github.com/dnschneid/crouton/wiki/Crouton-Command-Cheat-Sheet): 

> Launch a single app in a Window in the background: `sudo startxiwi -b -n chrootname app_name`

This is a specific version of the generality `sudo enter-chroot -n chrootname xiwi app_name`, which opens any application in a new window that is protected from `Ctrl-w`, `Ctrl-tab`, etc.

However:
1. [The xiwi documentation](https://github.com/dnschneid/crouton/wiki/crouton-in-a-Chromium-OS-window-(xiwi)) doesn't mention this is possible
2. I can't find documentation on launching new windows from inside an existing chroot
    * The crouton extension seems to only support grabbing one window at a time. This should be a separate ticket.

So given (1), I'd say this ticket could be closed with a documentation update.

NerdsvilleCEO:
@spiffytech do you mean that if you run ``sudo startxiwi -b -n chrootname app_name``, that it is protected against ``Ctrl-w``? That does not seem to be the case on my end :/

DennisLfromGA:
Clicking on the ['Crouton Integration'](https://chrome.google.com/webstore/detail/crouton-integration/gcpneefbbnfalgjniomfjknbcgkbijom) extension allows you to select and open any chroots in the drop-down.

*Hope this helps,*
-**DennisL**

PR #3383, opened 2017-09-02
Add confirmation on xiwi tab close for #3222

eyqs:
This addresses #3222 by providing a "Do you want to leave this site?" dialog when the user closes a crouton extension tab. For some reason, this dialog does not show up if you immediately close the tab as soon as it opens, no matter where the onbeforeunload line is placed. This shouldn't be too much of an issue though, because there's no chance of losing any unsaved work if you immediately close the tab like that.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -19,7 +19,7 @@ David Schneider
 Dennis Lockhart
 dimonf
 dumpweed
-eyqs
+Eugene Y. Q. Shen
 George Shank
 icecream95
 Igor Bukanov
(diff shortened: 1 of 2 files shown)


CONTRIBUTORS
```diff
@@ -19,7 +19,7 @@ David Schneider
 Dennis Lockhart
 dimonf
 dumpweed
-eyqs
+Eugene Y. Q. Shen
```
dnschneid: CONTRIBUTORS is an auto-generated file.  It didn't seem to grab your name properly from the git history, so the actual fix is to add a line to build/CONTRIBUTORS.sed to make the adjustment you did here.

eyqs:
Ah, I didn't realize that. No, I don't think this is necessary anymore.

dnschneid:
Cool.  Thanks for proposing the patch, though.  I'll fix your CONTRIBUTORS entry.

eyqs:
Thank you!

PR #3421, opened 2017-09-21
Mark Ubuntu Yakkety as EOL

icecream95:
Too bad Yakkety never made it out of the test state.


dnschneid:
Non-LTS releases rarely do.

PR #3483, opened 2017-10-31
Add Bionic (Ubuntu 18.04) to release list

icecream95:


icecream95:
Turns out there are currently hardly any packages so this is probably only suitable for command-line development at the moment - it seems that currently there aren't even any DE's available.

EnderNightLord-ChromeBook:
very true This will take a while

icecream95:
I can't even install the xorg target yet...

tsimonq2:
It's now open for development and daily images have started being produced.

Might be a good thing to have, eventually if it might be used as default. :wink: 

What do you think, @dnschneid?

dnschneid:
Since it's an LTS, bionic is certainly an option for a default in the future.  The migration path for defaults is a little odd, but since we've implemented somewhat reasonable behavior for the precise->xenial default migration, this would probably be straightforward to do as well.

Previously, we skipped trusty and went from precise to xenial due to some of the massive compatibility changes that took place.  Hopefully there's nothing major changing from xenial to bionic that would give us pause?

Anyway, assuming most of the packages are there, we should make the release available to people who like playing with broken things.

EnderNightLord-ChromeBook:
DONE !!!

PR #3503, opened 2017-11-11
Kodi - Open ports to allow smb/samba browsing

zguithues:
browsing smb/samba shares requires having ports 1025 - 65535 open. 
Source: [https://github.com/dnschneid/crouton/wiki/How-to-mount-network-shares-on-Chromebook-(sshfs,-cifs,-nfs-etc)](url)

this patch grabs the ip address from wlan0 and opens the ports when the source ip is within the local subnet.  if wlan0 has no ip, it does not modify iptables.

I'm not sure how many ChromeOS devices have an eth0, or any other interfaces that should be accounted for. or perhaps there is a better way to accomplish the same thing.

iptables gets reset on reboots, so it's probably not a big deal that the rule doesn't get cleared.  however, i would love to clear the port forward when kodi closes, or the chroot exits.  do you have any ideas for this @dnschneid ?

### DIFF
--- a/host-bin/startkodi
+++ b/host-bin/startkodi
@@ -14,5 +14,10 @@ By default, it will log into the primary user on the first chroot found.
 
 Options are directly passed to enter-chroot; run enter-chroot to list them."
 
+# Forward ports needed to browse smb shares
+if test -n "$(ip -4 -o addr show dev wlan0| awk '{split($4,a,"/");print a[1]}')"; then
+	iptables -I INPUT 1 -p udp --source $(ip -4 -o addr show dev wlan0| awk '{split($4,a,"/");print a[1]}')/255.255.255.0 --dport 1025:65535 -j ACCEPT
+fi
+
 exec sh -e "`dirname "\`readlink -f -- "$0"\`"`/enter-chroot" -t kodi "$@" "" \
     exec croutonpowerd -i xinit /usr/bin/kodi --standalone


host-bin/startkodi
```diff
@@ -14,5 +14,10 @@ By default, it will log into the primary user on the first chroot found.
 
 Options are directly passed to enter-chroot; run enter-chroot to list them."
 
+# Forward ports needed to browse smb shares
+if test -n "$(ip -4 -o addr show dev wlan0| awk '{split($4,a,"/");print a[1]}')"; then
```
divx118: This would assume you are using wlan0, you also could be using a usb ethernet adapter or another wlan device.

host-bin/startkodi
```diff
@@ -14,5 +14,10 @@ By default, it will log into the primary user on the first chroot found.
...
+	iptables -I INPUT 1 -p udp --source $(ip -4 -o addr show dev wlan0| awk '{split($4,a,"/");print a[1]}')/255.255.255.0 --dport 1025:65535 -j ACCEPT
+fi
```
divx118: Max line length  should be 80

host-bin/startkodi
zguithues: running that command returns a "1"

i agree that we should take into consideration other devices, i just don't know how...

divx118: @zguithues hmm strange I just tried it on some different linux devices and my chromebook and that seems to work. What do you get when you run `ip route get 1` for output. Maybe the awk part isn't full proof. If not, there sure will be other ways to get the used ip-address, without looking at the device.

host-bin/startkodi
DennisLfromGA: @divx118 method works for me, here's the output of each:

@zguithues -
```
chronos@localhost ~ $ ip -4 -o addr show dev wlan0| awk '{split($4,a,"/");print a[1]}'
192.168.1.225
```
@divx118  -
``` 
chronos@localhost ~ $ ip route get 1
1.0.0.0 via 192.168.1.1 dev wlan0 src 192.168.1.225 uid 1000 
    cache 
 
chronos@localhost ~ $ ip route get 1 | awk -F 'src ' '{ split($2,a," ");print a[1];exit}'
192.168.1.225
```

zguithues: ok, i got it working @divx118.  I was missing the space between the 2 " in the split($2,a," ").

`chronos@localhost / $ ip route get 1 | awk -F 'src ' '{ split($2,a," ");print a[1];exit}'
192.168.2.103`

i'll update the patch.  thanks!

zguithues:
OK, i've updated the command, and switched to using a variable so i could easily keep the line length down.

The only issue i still have with this is clearing the rule after kodi closes.  I'm not sure the best way to do this, but it's not really that big of a deal...

zguithues:
ok, previously this would fail if no network exists, i changed the test, removed the `-n`.  This seems to work from my end.  please let me know if you see any other issues with the code @DennisLfromGA  @divx118 

thanks a bunch for your help guys!

host-bin/startkodi
```diff
@@ -17,7 +17,7 @@ Options are directly passed to enter-chroot; run enter-chroot to list them."
 # Forward ports needed to browse smb shares
 MYIP=$(ip route get 1 | awk -F 'src ' '{ split($2,a," ");print a[1];exit}')
 
-if test -n $MYIP; then
+if test $MYIP; then
```
divx118: `if [ -n "$MYIP" ]; then` should work. It also would be more consistent with the syntax elsewhere used in crouton scripts. Also try to always quote strings.

host-bin/startkodi
```diff
@@ -17,7 +17,7 @@ Options are directly passed to enter-chroot; run enter-chroot to list them."
...
 	iptables -I INPUT 1 -p udp \
 	--source $MYIP/255.255.255.0 \
```
divx118: `"$MYIP"/255.255.255.0 \` quotes

zguithues:
thanks @divx118 !

host-bin/startkodi
```diff
@@ -14,5 +14,14 @@ By default, it will log into the primary user on the first chroot found.
 
 Options are directly passed to enter-chroot; run enter-chroot to list them."
 
+# Forward ports needed to browse smb shares
+MYIP=$(ip route get 1 | awk -F 'src ' '{ split($2,a," ");print a[1];exit}')
+
+if [ -n "$MYIP" ]; then
+	iptables -I INPUT 1 -p udp \
```
divx118: Sorry I overlooked this, but indentation is standard 4 spaces (no tabs) in the crouton scripts so it should look like
```
if [ -n "$MYIP" ]; then
    iptables -I INPUT 1 -p udp \
        --source "$MYIP"/255.255.255.0 \
        --dport 1025:65535 -j ACCEPT
fi
```

zguithues: done

DennisLfromGA:
LGTM2 👍

host-bin/startkodi
```diff
@@ -14,5 +14,14 @@ By default, it will log into the primary user on the first chroot found.
 
 Options are directly passed to enter-chroot; run enter-chroot to list them."
 
+# Forward ports needed to browse smb shares
+MYIP=$(ip route get 1 | awk -F 'src ' '{ split($2,a," ");print a[1];exit}')
```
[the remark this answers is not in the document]
zguithues: My limited knowledge indicated 2 places that obviously needed spacing, did i miss anything else?

I think i added the quote properly...

zguithues:
@dnschneid  Thanks for the direction, but unfortunately that's mostly Chinese to me...

can you point me in the direction of something i can build off of?

host-bin/startkodi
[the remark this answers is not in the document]
dnschneid: Yep, that looks better.

dnschneid:
Unfortunately, I don't think there's anything like it at the moment, and this feature has a weird set of requirements

1. root permission, regardless of which user is running kodi
2. needs to be triggered when the chroot unmounts
3. needs to only be triggered when launching kodi

I guess if we accept the caveat that you can only launch one kodi at a time, you can just get rid of the exec keyword in the startkodi script and add a trap on INT/HUP/0 to run the ip command.

zguithues:
i'll poke around and see what i can come up with.

i'd be happy for recommendations from the peanut gallery ;)

@dnschneid do you consider the removal of this iptables rule to be imperative to the merging of this patch?

RUDIVANTORRE:
z

dnschneid:
I agree with @RUDIVANTORRE.  It should be pretty straightforward to make the change, so I think the merge can wait until you have it working.

PR #3507, opened 2017-11-12
Create mate

EnderNightLord-ChromeBook:
Done File This is my best work yet I'm only at intermediate and im finished my first work
BUT im not sure if it will install?


DennisLfromGA:
@EnderNightLord-ChromeBook,

I'm not sure if you're aware but there's already a PR and work in progress for Mate, see https://github.com/dnschneid/crouton/pull/2532

-**DennisL**

dnschneid:
Yep, and as you can see from that PR, there's a bit more involved in adding a target.  Perhaps you should continue where @zxvv left off?

Issue #1410, opened 2015-02-13
Feature Request: enter-chroot should be aware of multiprofiles


beaufortfrancois:
Right now, `~/Downloads` maps the first logged in user Downloads folder which is great except when you use multiple accounts at the same time. See https://support.google.com/chromebook/answer/6088201?hl=fr&source=genius-rts

It would be great if `enter-chroot` could be aware of the profiles we use and map accordingly the `~/Downloads` folder.

DennisLfromGA:
I have multiple accounts setup on my HP Chromebook 14 and I've noticed some interesting 'features' when using them. When switching into each account, if you open the 'Files' app, it shows that account's  Downloads folder and their files. But, if you open a shell (Ctrl+Alt+T) in that account, it uses the first logged in user's account and shell AND ~/Downloads folder. I assume this is intended but I don't know how 'crouton' could get around it this since it seems to use only the first logged in user's shell.

I have a short [screencast](http://goo.gl/dAcX3p) ( Thanx Francois :) showing this behaviour that demonstrates the above - I hope ;) 

-DennisL

tedm:
On the stable channel, ARM, I don't have the ability to bring up multiple accounts yet, but I can understand the issue. In the chroot, ~/Downloads is given ownership of first linux user and their corresponding group (tedm:1001), perms: 710. In a crosh shell at /home/user/chronos, Downloads is owned by chronos:chronos-access with same perms. When additional user desktops are opened on Chromium OS, is the UID:GID shared, or unique? I'm also baffled that ~/Downloads is now a ~/tmp dir with the caution that things may be deleted by the system. We were warned of this lack of permanence earlier on, but traditional desktop users (Mac/Windows), may find this concept disconcerting, as it's moving away from having some safe local storage.

dnschneid:
It looks like multiple users get mounted to /home/chronos/u-\*.  Could someone with multiple login post the output of `ls -ld /home/chronos/u-*`?  Non-logged in users have empty directories owned by root.

It looks like we can get the names of the users via `/home/chronos/Local State`, but then what?  Subfolders of ~/Downloads for each account?  Stick with the current behavior but expose multiple login in the /etc/crouton/shares syntax?

DennisLfromGA:
Here's `ls -ld /home/chronos/u-*`:
```
chronos@localhost ~ $ ls -ld /home/chronos/u-*
[...]

Issue #560, opened 2013-12-30
courtonpowerd not present in cli-extra?


scottharney:
I created a cli-extra chroot environment and croutonpowerd is not present.  Is it only available in the x11 chroots?

PR #3531, opened 2017-11-24
Use Downloads folder of the user who launched the chroot

ibrado:
For multiple logins, this uses the `LastActiveUser` in `/home/chronos/Local State` to determine which ~/Downloads folder should be mapped in the chroot, i.e. that of the user who started the chroot.

Additionally, this creates `/etc/crouton/user` containing the email address of the ~/Downloads "owner".

See also issue #1410

### DIFF
--- a/host-bin/enter-chroot
+++ b/host-bin/enter-chroot
@@ -456,7 +456,9 @@ fi
 # Bind-mount the stuff in $CHROOT/etc/crouton/shares, unless NOLOGIN is set
 if [ -z "$NOLOGIN" -a -f "$shares" ]; then
     if [ "$activeuser" != "Default" ]; then
-        localdownloads="$(jq -r '.profile.info_cache | to_entries[] | select(.value.user_name == "'$activeuser'") | "/home/chronos/\(.key)/Downloads"' '/home/chronos/Local State')"
+        localdownloads="$(jq -r '.profile.info_cache | to_entries[]
+          | select(.value.user_name == "'$activeuser'")
+          | "/home/chronos/\(.key)/Downloads"' '/home/chronos/Local State')"
     fi
     if [ -z "$localdownloads" ]; then
         localdownloads='/home/chronos/user/Downloads'
--- a/host-bin/enter-chroot
+++ b/host-bin/enter-chroot
@@ -338,15 +338,6 @@ tmpfsmount() {
     mount -i -t tmpfs -o "rw${2:+,}$2" tmpfs "$target"
 }
 
-# Determines currently-logged in user
-getactiveuser() {
-    local activeuser="Default"
-    if [ -x "/usr/bin/jq" ]; then
-        activeuser="$(jq -r .LastActiveUser '/home/chronos/Local State')"
-    fi
-    echo $activeuser
-}
-
 # If /var/run isn't mounted, we know the chroot hasn't been started yet.
 if mountpoint -q "`fixabslinks '/var/run'`"; then
     firstrun=''
@@ -446,16 +437,13 @@ downloads ~/Downloads
 EOF
 fi
 
-activeuser=$(getactiveuser)
-if [ -n "$firstrun" ]; then
-    # Store the active user for possible use in other scripts, etc.
-    activeuserfile="$(fixabslinks '/etc/crouton/user')"
-    echo "$activeuser" > $activeuserfile
-fi
-
 # Bind-mount the stuff in $CHROOT/etc/crouton/shares, unless NOLOGIN is set
 if [ -z "$NOLOGIN" -a -f "$shares" ]; then
-    if [ "$activeuser" != "Default" ]; then
+    activeuser=''
+    if [ -x "/usr/bin/jq" ]; then
+        activeuser="$(jq -r .LastActiveUser '/home/chronos/Local State')"
+    fi
+    if [ -n "$activeuser" ]; then
         localdownloads="$(jq -r '.profile.info_cache | to_entries[]
           | select(.value.user_name == "'$activeuser'")
           | "/home/chronos/\(.key)/Downloads"' '/home/chronos/Local State')"
--- a/host-bin/enter-chroot
+++ b/host-bin/enter-chroot
@@ -338,6 +338,15 @@ tmpfsmount() {
     mount -i -t tmpfs -o "rw${2:+,}$2" tmpfs "$target"
 }
 
+# Determines currently-logged in user
+getactiveuser() {
+    local activeuser="Default"
+    if [ -x "/usr/bin/jq" ]; then
+        activeuser=$(jq -r .LastActiveUser /home/chronos/Local\ State)
+    fi
+    echo $activeuser
+}
+
 # If /var/run isn't mounted, we know the chroot hasn't been started yet.
 if mountpoint -q "`fixabslinks '/var/run'`"; then
     firstrun=''
@@ -436,9 +445,21 @@ elif [ ! -f "$shares" ]; then
 downloads ~/Downloads
 EOF
 fi
+
+activeuser=$(getactiveuser)
+if [ -n "$firstrun" ]; then
+    # Store the active user for possible use in other scripts, etc.
+    activeuserfile="$(fixabslinks '/etc/crouton/user')"
+    echo $activeuser > $activeuserfile
+fi
+
 # Bind-mount the stuff in $CHROOT/etc/crouton/shares, unless NOLOGIN is set
 if [ -z "$NOLOGIN" -a -f "$shares" ]; then
-    localdownloads='/home/chronos/user/Downloads'
+    if [ "$activeuser" != "Default" ]; then
+        localdownloads=$(jq -r ".profile.info_cache | to_entries[] | select(.value.user_name == \"$activeuser\") | \"/home/chronos/\(.key)/Downloads\"" /home/chronos/Local\ State)
+    else
+        localdownloads='/home/chronos/user/Downloads'
+    fi
     if [ ! -d "$localdownloads" ]; then
         localdownloads=''
     fi
--- a/host-bin/enter-chroot
+++ b/host-bin/enter-chroot
@@ -457,7 +457,8 @@ fi
 if [ -z "$NOLOGIN" -a -f "$shares" ]; then
     if [ "$activeuser" != "Default" ]; then
         localdownloads=$(jq -r ".profile.info_cache | to_entries[] | select(.value.user_name == \"$activeuser\") | \"/home/chronos/\(.key)/Downloads\"" /home/chronos/Local\ State)
-    else
+    fi
+    if [ -z "$localdownloads" ]; then
         localdownloads='/home/chronos/user/Downloads'
     fi
     if [ ! -d "$localdownloads" ]; then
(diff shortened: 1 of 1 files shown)


ibrado:
I signed it!

host-bin/enter-chroot
```diff
@@ -436,9 +445,21 @@ elif [ ! -f "$shares" ]; then
...
+    # Store the active user for possible use in other scripts, etc.
+    activeuserfile="$(fixabslinks '/etc/crouton/user')"
+    echo $activeuser > $activeuserfile
+fi
+
 # Bind-mount the stuff in $CHROOT/etc/crouton/shares, unless NOLOGIN is set
 if [ -z "$NOLOGIN" -a -f "$shares" ]; then
-    localdownloads='/home/chronos/user/Downloads'
+    if [ "$activeuser" != "Default" ]; then
+        localdownloads=$(jq -r ".profile.info_cache | to_entries[] | select(.value.user_name == \"$activeuser\") | \"/home/chronos/\(.key)/Downloads\"" /home/chronos/Local\ State)
```
DennisLfromGA: Although 'activeuser' is populated and valid, both 'localdownloads' locations appear empty:
```
chronos@localhost ~ $ localdownloads=$(jq -r ".profile.info_cache | to_entries[] | select(.value.user_name == \"$activeuser\") | \"/home/chronos/\(.key)/Downloads\"" /home/chronos/Local\ State)

ls -l $localdownloads
/home/chronos/Default/Downloads:
total 0

/home/chronos/LockScreenAppsProfile/Downloads:
total 0
```
Yet when I list `/home/chronos/user/Downloads` it contains all of my files.

I'm not sure this routine will get you the expected results but I may not understand the full intent.

ibrado: Ah, I'm guessing you haven't logged in to multiple accounts on that Chromebook?

Even so, I expected that the `jq -r` part would not have run (`if [ "$activeuser" != "Default" ]`) and would have used `/home/chronos/user/Downloads` instead.

Would you mind checking the output of `jq -r .LastActiveUser /home/chronos/Local\ State` ? That's what `activeuser` should have been set to.

ibrado: BTW, here's what I'm getting (except for the actual email address and u- id):
```
$ activeuser=$(jq -r .LastActiveUser /home/chronos/Local\ State)
$ echo $activeuser
myemail@example.com

$ localdownloads=$(jq -r ".profile.info_cache | to_entries[] | select(.value.user_name == \"$activeuser\") | \"/home/chronos/\(.key)/Downloads\"" /home/chronos/Local\ State)
$ echo $localdownloads
/home/chronos/u-xxxxxxxe64bcd6851c24f81f6566ef0519e781cba/Downloads

$ ls $localdownloads
...files of myemail@example.com here...
```

DennisLfromGA: You're correct, I only have one account setup on this device.
Here's the output you requested with my email masked:
```
chronos@localhost ~ $ jq -r .LastActiveUser /home/chronos/Local\ State
xxxxxxxx@gmail.com
```
It could be that when I add another user account I'd get the correct $localdownloads location populated but I'm thinking it might be best not to do that yet so we can test this on devices with only one account. When I run the second part beginning with `localdownloads=...` I still get the two empty directories.

Let me know if I'm doing something wrong or if you'd like me to test something else.

DennisLfromGA: I ran the modified `enter-chroot` adding `set -x` at line [#455](https://github.com/ibrado/crouton/blob/232fc0d951e5ad096d2255381003b87f281517f7/host-bin/enter-chroot#L455) and an `exit` at line [#560](https://github.com/ibrado/crouton/blob/232fc0d951e5ad096d2255381003b87f281517f7/host-bin/enter-chroot#L560)  to see what's getting set. I posted the output at pastebin [here](https://pastebin.com/9yP7PQvV).  Hopefully, you'll get something useful out of it.

ibrado: Thanks! I guess those keys aren't set if there's only one, account set up, i.e. `jq -r ".profile.info_cache | to_entries[]" /home/chronos/Local\ State` would probably only have the default entry. Sorry I couldn't test this myself, only have one Chromebook and am hesitant to powerwash, hehe.

Anyway, I've made it so it always checks for an empty `$localdownloads`. Just committed it...

DennisLfromGA: Cool, that works. 👍

DennisLfromGA: BTW, I ran both commits on a Chromebook with multiple users and they both worked so I guess it was good to check it on a single user device.

ibrado: Woohoo! Yeah, it was good you had that around... I was seriously considering a powerwash earlier. :-P

DennisLfromGA: I have four Chromebooks going all the way back to a Cr-48 and now a Pixelbook so I have options. ,-)
I like to run the 'chrx' script, 1st phase only and use partiton 7 as a safe place for crouton so I don't have to worry about powerwashing.

ibrado: Haven't had the guts to try chrx yet, haha! I had 2 Chromebooks before this one, but I lent one and gave the other away...

host-bin/enter-chroot
```diff
@@ -338,6 +338,15 @@ tmpfsmount() {
     mount -i -t tmpfs -o "rw${2:+,}$2" tmpfs "$target"
 }
 
+# Determines currently-logged in user
+getactiveuser() {
+    local activeuser="Default"
+    if [ -x "/usr/bin/jq" ]; then
+        activeuser=$(jq -r .LastActiveUser /home/chronos/Local\ State)
```
dnschneid: quote the path instead of using backslash escapes

dnschneid: quote the output of the command

host-bin/enter-chroot
```diff
@@ -338,6 +338,15 @@ tmpfsmount() {
...
+    fi
+    echo $activeuser
```
dnschneid: quote variable usage

host-bin/enter-chroot
```diff
@@ -436,9 +445,22 @@ elif [ ! -f "$shares" ]; then
 downloads ~/Downloads
 EOF
 fi
+
+activeuser=$(getactiveuser)
```
dnschneid: I don't think this needs to be a separate function

host-bin/enter-chroot
```diff
@@ -436,9 +445,22 @@ elif [ ! -f "$shares" ]; then
...
+if [ -n "$firstrun" ]; then
+    # Store the active user for possible use in other scripts, etc.
+    activeuserfile="$(fixabslinks '/etc/crouton/user')"
+    echo $activeuser > $activeuserfile
```
dnschneid: I would rather not (easily) expose this information to the chroot.  Is there a real use for it?

host-bin/enter-chroot
```diff
@@ -436,9 +445,22 @@ elif [ ! -f "$shares" ]; then
...
+fi
+
 # Bind-mount the stuff in $CHROOT/etc/crouton/shares, unless NOLOGIN is set
 if [ -z "$NOLOGIN" -a -f "$shares" ]; then
-    localdownloads='/home/chronos/user/Downloads'
+    if [ "$activeuser" != "Default" ]; then
+        localdownloads=$(jq -r ".profile.info_cache | to_entries[] | select(.value.user_name == \"$activeuser\") | \"/home/chronos/\(.key)/Downloads\"" /home/chronos/Local\ State)
```
dnschneid: Keep lines to 80 characters, please

dnschneid: Quote the output of the command

dnschneid: Quote the path instead of escaping spaces

dnschneid: The escape before the parentheses...is that meaningful to jq?

host-bin/enter-chroot
```diff
@@ -338,6 +338,15 @@ tmpfsmount() {
     mount -i -t tmpfs -o "rw${2:+,}$2" tmpfs "$target"
 }
 
+# Determines currently-logged in user
+getactiveuser() {
+    local activeuser="Default"
+    if [ -x "/usr/bin/jq" ]; then
```
dnschneid: I like that you check for the existence of jq (in case it disappears in the future).  Is it possible, though, to do this with awk in a reasonable way?

dnschneid (commented):
Thanks for implementing this cleanly!

host-bin/enter-chroot
ibrado: I don't know enough awk to answer this :blush: .. I do think jq will be around for a while since the config file is json.

host-bin/enter-chroot
ibrado: Done!

ibrado: Done!

host-bin/enter-chroot
ibrado: Done!

host-bin/enter-chroot
ibrado: Ok... Since we're not creating /etc/crouton/user anymore (see next comment), there's no reason for it.

host-bin/enter-chroot
ibrado: Just a nice-to-have. Removed...

host-bin/enter-chroot
ibrado: Done!

ibrado: Done!

ibrado: Done!

ibrado: Yes, else you'd get a literal `(.key)` in the result.

ibrado:
@DennisLfromGA Hi.. Would you mind checking the latest commits in your single-user device? Thanks!

host-bin/enter-chroot
```diff
@@ -436,9 +436,21 @@ elif [ ! -f "$shares" ]; then
 downloads ~/Downloads
 EOF
 fi
+
 # Bind-mount the stuff in $CHROOT/etc/crouton/shares, unless NOLOGIN is set
 if [ -z "$NOLOGIN" -a -f "$shares" ]; then
-    localdownloads='/home/chronos/user/Downloads'
+    activeuser=''
+    if [ -x "/usr/bin/jq" ]; then
+        activeuser="$(jq -r .LastActiveUser '/home/chronos/Local State')"
```
dnschneid:
```
activeuser="$(sed -n 's/.*"LastActiveUser":\s*"\([^"]*\)".*/\1/p' '/home/chronos/Local State')"
```

host-bin/enter-chroot
```diff
@@ -436,9 +436,21 @@ elif [ ! -f "$shares" ]; then
...
+    fi
+    if [ -n "$activeuser" ]; then
+        localdownloads="$(jq -r '.profile.info_cache | to_entries[]
+          | select(.value.user_name == "'$activeuser'")
+          | "/home/chronos/\(.key)/Downloads"' '/home/chronos/Local State')"
```
dnschneid: I'm not convinced this is better (also not certain it is the correct way of doing it), but if we want to get rid of the jq dependency:
```
localdownloads="$(grep -l '"account_id":\s*"'"$activeuser"'"' /home/chronos/u-*/Preferences | sed 's/Preferences$/Downloads/;q')"
```

DennisLfromGA:
I looks like something along these lines is being addressed in CrOS:
- [crbug.com/770766 - Allow running containers with multi-user](https://bugs.chromium.org/p/chromium/issues/detail?id=770766)

-**DennisLfromGA**

Issue #3519, opened 2017-11-18
Horizontal scrolling is not "Austrailian"


DavidFrahm:
```
chronos@localhost / $ sudo edit-chroot -all
name: xenial
encrypted: no
Entering /mnt/stateful_partition/crouton/chroots/xenial...
crouton: version 1-20171109145451~master:35a16889
release: xenial
architecture: amd64
xmethod: xiwi
targets: xiwi,touch
host: version 9901.66.0 (Official Build) stable-channel eve 
kernel: Linux localhost 4.4.79-11650-ge987f76b729a #1 SMP PREEMPT Tue Oct 31 22:05:39 PDT 2017 x86_64 x86_64 x86_64 GNU/Linux
freon: yes
```

#### Please describe your issue:

In Chrome OS settings I have set the touchpad to the "Austrailian" (aka "natural") setting.

After installing crouton, the crouton/xiwi windowed apps I run do not scroll horizontally the same way Chrome OS does.

Crouton/xiwi apps do obey that setting for vertical scrolling.

Please advise if there is a crouton config or anything I could try to get natural scrolling horizontally. Alternatively, if there's a different crouton install, target, etc. that should yield better results that would be helpful as well.

#### If known, describe the steps to reproduce the issue:

Chromebook model: Google Pixelbook
Crouton install: `$ sudo sh ~/Downloads/crouton -t xiwi,touch`

dnschneid:
We only process the wheel messages we receive; see https://github.com/dnschneid/crouton/blob/master/host-ext/nacl_src/kiwi.cc#L610

I dunno if we have access to the australian scrolling setting to do anything smarter.  @drinkcat?

ibrado:
I inserted the following as line 125 (right after the `if` of `# Crouton-in-a-tab: Start fbserver and launch display`) in `/usr/local/bin/croutonxinitrc-wrapper`:
```
xmodmap -e "pointer = 1 2 3 4 5 7 6"
```
This works for me for both windowed and full-desktop (KDE) apps. Scroll drection remained consistent between ChromeOS and the xiwi apps even as I toggled between traditional/Australian scrolling.

DavidFrahm:
The solution from @ibrado worked perfectly for me.

Is there something we can do to detect Austrailian and apply this automatically for people?

DavidFrahm:
I'm just using plain xiwi with the extension to run one app (WebStorm IDE) in a window.

My crouton install:

```
[...]

PR #3534, opened 2017-11-27
Fix horizontal scroll direction for xiwi

ibrado:
xiwi apps scroll horizontally in the opposite direction as ChromeOS. This just reverses the direction via xmodmap and makes scrolling consistent across xiwi and ChromeOS.

Tested with Australian/traditional settings, KDE/XFCE desktops, and separate xiwi windows.

See also #3519

### DIFF
--- a/chroot-bin/croutonxinitrc-wrapper
+++ b/chroot-bin/croutonxinitrc-wrapper
@@ -122,6 +122,8 @@ fi
 
 # Crouton-in-a-tab: Start fbserver and launch display
 if [ "$xmethodtype" = 'xiwi' ]; then
+    # Fix horizontal scroll direction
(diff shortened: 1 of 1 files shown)


dnschneid:
If we're applying it unconditionally for xiwi, it would be better just to generate the correct events in the first place: https://github.com/dnschneid/crouton/blob/master/host-ext/nacl_src/kiwi.cc#L624 (and 628).

@drinkcat any reason we're generating the events the way we are?  Did something on the Chrome side change in the last 3 years?

therealbstern:
FYI, this is still a thing on my `falco` Chromebook.

joshuafvr:
I tried installing kali Linux on my Samsung s9 thru termux  it brings this
Screenshot_20220621-155909_Termux

Issue #3623, opened 2018-01-27
issue #3621 (gksu has no install candidate when installing with xfce)


pcs3rd:
```
Please paste the output of the following command here: sudo edit-chroot -
Can't get kali-rolling to start even, nothing to start, chroot doesn't finish set up.
```
(the original issue #3621 )
when trying to install kali-rolling, it gets to installing xfce and manages to choke this out...
```
Building dependency tree       
Reading state information... Done
The following packages were automatically installed and are no longer required:
  libasound2-dev:i386 libc6-dev:i386 libdrm-dev libpthread-stubs0-dev
  libspeex-dev:i386 libspeex1:i386 libspeexdsp-dev:i386 libx11-dev libxau-dev
  libxcb1-dev libxdmcp-dev libxext-dev libxfixes-dev libxi-dev
  linux-libc-dev:i386 x11proto-core-dev x11proto-fixes-dev x11proto-input-dev
  x11proto-kb-dev x11proto-xext-dev xorg-sgml-doctools xtrans-dev
Use 'apt autoremove' to remove them.
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Package gksu is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'gksu' has no installation candidate
Failed to complete chroot setup.
Unmounting /media/removable/crouton/chroots/kali-rolling...
```
[Exact same thing mentioned on stack exchange](https://unix.stackexchange.com/questions/419199/package-gksu-has-no-installation-candidate-on-chromebook-kali-installation) 3 days ago (as of Friday, January 26). I have a feeling this is more of a bug with how crouton sets up kali than a bug with kali itself. Does anyone know where the setup script is in the chroot? I can't find it.

rickybrent:
I ran into the same issue with buster -- it's [targets/gtkextra](https://github.com/dnschneid/crouton/pull/3624/commits/5d905ebf931e523c135c0d3e082ebb486f8686fb).

Seems like debian's [removed it](https://bugs.debian.org/867236).

pcs3rd:
@rickybrent , is there any quickfix. I really want to install kali. And once the fix is released, Will I have to redownload crouton?

rickybrent:
[...]

Issue #3621, opened 2018-01-26
Packages missing for kali-rolling (gksu has no installation candidate) (probably should be under PIE)


pcs3rd:
I can't paste output of edit-chroot all, I don't have a chroot, it is failing to finish installation with apt-get error out
```
Building dependency tree       
Reading state information... Done
The following packages were automatically installed and are no longer required:
  libasound2-dev:i386 libc6-dev:i386 libdrm-dev libpthread-stubs0-dev
  libspeex-dev:i386 libspeex1:i386 libspeexdsp-dev:i386 libx11-dev libxau-dev
  libxcb1-dev libxdmcp-dev libxext-dev libxfixes-dev libxi-dev
  linux-libc-dev:i386 x11proto-core-dev x11proto-fixes-dev x11proto-input-dev
  x11proto-kb-dev x11proto-xext-dev xorg-sgml-doctools xtrans-dev
Use 'apt autoremove' to remove them.
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Package gksu is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'gksu' has no installation candidate
Failed to complete chroot setup.
Unmounting /media/removable/crouton/chroots/kali-rolling...
```
any 'official' fix, or is there a way to specify to brute through errors and just work with an incomplete chroot. If not, I am going to (kindly) make a feature request.

pcs3rd:
This isn't an answer, but for now, users MAY be able to install kali sana (2.0)and use [THIS PAGE](https://forums.kali.org/showthread.php?32697-How-to-upgrade-to-latest-kali-rolling-2016-1) to at least get up to date where rolling should be (hopefully)

hsosa88:
I have same problem, but when I try to install kali sana get this:
```
WARNING: sana has reached upstream end-of-life.
That means there will be no package updates available.
You also have to specify a mirror to crouton (-m) for installation to proceed.
```

pcs3rd:
Yea, I was going to go from sana and upgrade, but there doesn't seem to be a way to. @dnschneid is there a way to do this? what mirrors are there?

pcs3rd:
yet again another potential temporary solution, instead of installing xfce at time of setup, run 'sudo sh -e ~/Downloads/crouton -r kali-rolling -n kali -t xiwi' and then install xfce

[...]

PR #3624, opened 2018-01-27
Remove gksu from gtk-extra target for debian buster and kali-rolling

rickybrent:
Per debian's bug, they've removed gksu (but patched gdebi to use policy kit):
https://bugs.debian.org/867236

Should fix [3623](https://github.com/dnschneid/crouton/issues/3623)/[3621](https://github.com/dnschneid/crouton/issues/3621) and buster, which has the same issue atm.

### DIFF
--- a/targets/gtk-extra
+++ b/targets/gtk-extra
@@ -3,12 +3,15 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 REQUIRES='x11'
-DESCRIPTION='GTK-based tools including gdebi, gksu, and a simple browser.'
+DESCRIPTION='GTK-based tools including gdebi and a simple browser.'
 . "${TARGETSDIR:="$PWD"}/common"
 
 ### Append to prepare.sh:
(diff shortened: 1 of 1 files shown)


targets/gtk-extra
```diff
@@ -3,12 +3,15 @@
...
 REQUIRES='x11'
-DESCRIPTION='GTK-based tools including gdebi, gksu, and a simple browser.'
+DESCRIPTION='GTK-based tools including gdebi and a simple browser.'
 . "${TARGETSDIR:="$PWD"}/common"
 
 ### Append to prepare.sh:
 install_dummy network-manager network-manager-gnome
-install gdebi gksu
+install gdebi
+if release -lt buster -lt kali-rolling; then
```
dnschneid: This whitelist results in nothing on Ubuntu.  So you probably want to add `-lt artful`

rickybrent: Done! 

Hmm. I imagine it still worked when I tested Xenial because gdebi requires gksu before debian ported it to policy kit. Would simply removing it from all targets be a better approach?

PR #3662, opened 2018-02-23
Adding root password requirement to README.md

MezNyx:
crouton displays that it needs the root password
set for encryption to work when run. This will
update the README to reflect that requirement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,14 @@ start\* commands. Ta-da! That was easy.
 
   1. Add the `-e` parameter when you run crouton to create an encrypted chroot
      or encrypt a non-encrypted chroot.
-  2. You can get some extra protection on your chroot by storing the decryption
+  2. If you want to use encryption, you must have the root password set by using
+     `sudo chromeos-setdevpasswd`. You can tell the root password is not set if
+     you see the following in your install:
+     ```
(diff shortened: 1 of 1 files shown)


MezNyx:
I signed it!

DennisLfromGA:
@MezNyx,

It's true you need to set a dev passwd before encrypting a chroot but the '[mount-chroot](https://github.com/dnschneid/crouton/blob/master/host-bin/mount-chroot#L141)' script invokes `chromeos-setdevpasswd` if one has not yet been set. 

This way the user doesn't have to do it or even know to do it before they use crouton's '-e' option since it'll do it for them if needed.

*Hope this helps,*
-**DennisLfromGA**

MezNyx:
Ok thanks for letting me know. I will close this PR. There may be a bug though as when I ran it, I didn't get an invocation of chromeos-setdevpasswd, but instead just got continuous spamming of the openssl: symbol lookup error: openssl: undefined symbol: RC5_32_set_key till I finally had to hit ctrl+c. I will do further testing.

DennisLfromGA:
@MezNyx,

That's interesting, if it's throwing an error the devs will want to know.
Maybe your testing will give us something to look at.

*Thanx,*
-**DennisLfromGA**

PR #3668, opened 2018-02-27
add i3

errietta:
This is my first time trying to touch crouton, but I made a quick target for the i3 window manager.
It appears to work. Please review if you want to 🐱


errietta:
I signed it!

PR #3714, opened 2018-03-30
ubuntu/boostrap: Pin debootstrap to 1.0.92~bpo9+1

### LINKED ISSUE #3704: Kali-rolling and sid Debian both run into this problem"W: Cannot check Release signature; keyring file not available /usr/share/keyrings/debian-archive-keyring.gpg" (For Debian same for kali as well)
```
sudo sh -e ~/Downloads/crouton -r kali-rolling -t xfce
sudo sh -e ~/Downloads/crouton -r sid -t xfce 
"W: Cannot check Release signature; keyring file not available /usr/share/keyrings/debian-archive-keyring.gpg"
```

#### Please describe your issue:
As I run both of these, they get the same output. I tried again with an Ubuntu install and it installed perfectly! I have looked up the issue and have seen it popping up over and over again and cannot seem to find a solution to the problem. I powerwashed my chromebook last night because of the issue, but it seems it did not work. 

Thanks

#### If known, describe the steps to reproduce the issue:
sudo sh -e ~/Downloads/crouton -r kali-rolling -t xfce
or
sudo sh -e ~/Downloads/crouton -r sid -t xfce

drinkcat:
Later versions fail, for some reason.

Pinning the version is also generally a good idea, so that we can
uprev it at our own pace, when we need it for new releases, instead
of having to react quickly when HEAD is broken.

Fix #3704.

### DIFF
--- a/installer/ubuntu/bootstrap
+++ b/installer/ubuntu/bootstrap
@@ -8,28 +8,22 @@
 # directory), $INSTALLERDIR/$DISTRO, $RELEASE, $BOOTSTRAP_RELEASE (if different
 # from $RELEASE), $ARCH, and $MIRROR.
 
-# Grab the latest release of debootstrap
-echo 'Downloading latest debootstrap...' 1>&2
-d='https://anonscm.debian.org/gitweb/?p=d-i/debootstrap.git;a=snapshot;h=HEAD;sf=tgz'
-if ! curl -f -# -L --connect-timeout 60 --retry 2 "$d"  \
+# Grab a known good release of debootstrap
+# TODO(#3713): Figure out why >= 1.0.92~bpo9+1 is broken
+debootstrap_version="1.0.92~bpo9+1"
+
+echo "Downloading debootstrap ${debootstrap_version}..." 1>&2
+if [ "${debootstrap_version}" = "HEAD" ]; then
+    d='https://anonscm.debian.org/gitweb/?p=d-i/debootstrap.git;a=snapshot;h=HEAD;sf=tgz'
+else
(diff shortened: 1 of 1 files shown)


installer/ubuntu/bootstrap
```diff
@@ -8,28 +8,21 @@
...
+
+echo "Downloading debootstrap ${debootstrap_version}..." 1>&2
+if [ "${debootstrap_version}" = "HEAD" ]; then
+    d='https://anonscm.debian.org/gitweb/?p=d-i/debootstrap.git;a=snapshot;h=HEAD;sf=tgz'
+else
+    base='http://httpredir.debian.org/debian/pool/main/d/debootstrap'
+    d="${base}/debootstrap_${debootstrap_version}.tar.gz"
+fi
+
+if ! curl -f -# -L --connect-timeout 60 --retry 2 "$d$f" \
```
dnschneid: what is `$f`

installer/ubuntu/bootstrap
```diff
@@ -8,28 +8,21 @@
...
-Check your internet connection or proxy settings and try again.'
-    fi
-    v="${f#*_}"
-    v="${v%.tar.gz}"
-    echo "Downloading debootstrap version $v..." 1>&2
-    if ! curl -f -# -L --connect-timeout 60 --retry 2 "$d$f" \
-            | tar -C "$tmp" --strip-components=1 -zx 2>/dev/null; then
-        error 1 'Failed to download debootstrap.'
-    fi
+    error 1 'Failed to download debootstrap.'
```
dnschneid: `Check your internet connection or proxy settings and try again.'`?

PR #3715, opened 2018-03-30
debootstrap: Print debootstrap.log on error

drinkcat:
Makes debugging a tiny bit easier.

### DIFF
--- a/installer/ubuntu/bootstrap
+++ b/installer/ubuntu/bootstrap
@@ -71,7 +71,11 @@ fi
 
 # Grab the release and drop it into the subdirectory
 echo 'Downloading bootstrap files...' 1>&2
-PATH="$newpath" DEBOOTSTRAP_DIR="$tmp" $FAKEROOT \
-    "$tmp/debootstrap" --foreign --extractor='ar' --arch="$ARCH" \
-    "$RELEASE" "$tmp/$subdir" "$MIRROR" 1>&2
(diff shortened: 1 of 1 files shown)


installer/ubuntu/bootstrap
```diff
@@ -71,7 +71,11 @@ fi
...
 # Grab the release and drop it into the subdirectory
 echo 'Downloading bootstrap files...' 1>&2
-PATH="$newpath" DEBOOTSTRAP_DIR="$tmp" $FAKEROOT \
-    "$tmp/debootstrap" --foreign --extractor='ar' --arch="$ARCH" \
-    "$RELEASE" "$tmp/$subdir" "$MIRROR" 1>&2
+if ! PATH="$newpath" DEBOOTSTRAP_DIR="$tmp" $FAKEROOT \
+        "$tmp/debootstrap" --foreign --extractor='ar' --arch="$ARCH" \
+        "$RELEASE" "$tmp/$subdir" "$MIRROR" 1>&2; then
+    echo "debootstrap error log:" 1>&2
+    cat "$tmp/$subdir/debootstrap/debootstrap.log" 1>&2 || true
```
dnschneid: is this going to be crazy long if debootstrap fails towards the end of bootstrapping due to a network event?

drinkcat: Yeah, you're right... Maybe just tailing the last 3 lines would be better than nothing...

PR #3717, opened 2018-03-31
bootstrap: Do not install apt-transport-https on >=bionic

drinkcat:
bionic onwards does not require installing apt-transport-https (apt provides it). In theory, there is a transitional package, but debootstrap does not find it.

Somewhat relevant bug:
   https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=879755

`2018-03-31_01-51-56_drinkcat_chroagh_unbreak-bionic_10` is quite green.

### DIFF
--- a/installer/ubuntu/bootstrap
+++ b/installer/ubuntu/bootstrap
@@ -8,68 +8,85 @@
 # directory), $INSTALLERDIR/$DISTRO, $RELEASE, $BOOTSTRAP_RELEASE (if different
 # from $RELEASE), $ARCH, and $MIRROR.
 
-# Grab the latest version from gitweb
+# This script is also sourced with nothing but $tmp set, as part of the bundle
+# build process. Any pre-downloading of tools should happen and be stored in
+# $tmp; the contents of which will be moved into $INSTALLERDIR/.. in the bundle
+
+# Grab debootstrap
 d="https://salsa.debian.org/installer-team/debootstrap/-/archive/1.0.101/debootstrap-1.0.101.tar.gz"
 
-if ! curl -f -# -L --connect-timeout 60 --retry 2 "$d" \
-        | tar -C "$tmp" --strip-components=1 -zx 2>/dev/null; then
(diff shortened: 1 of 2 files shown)


installer/ubuntu/bootstrap
```diff
@@ -46,6 +46,14 @@ else
     echo "Unable to patch debootstrap, moving on..." 1>&2
 fi
 
+# bionic onwards does not require installing apt-transport-https (apt provides
+# it). In theory, there is a transitional package, but debootstrap does not find
+# it. Somewhat relevant bug:
+#   https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=879755
+if release -ge bionic; then
```
dnschneid: Seems to also affect buster and sid?  And probably kali-rolling?

drinkcat: For some reason debootstrap does find the transitional package on buster and sid. Unsure about kali-rolling (our tests are using an http server...).

We could file a bug for bionic...

dnschneid: Maybe the transitional package was added late in bionic and ended up in the wrong repo.  But it sounds like this package is history and we should remove it if `! release -le artful -le stretch -le kali` or something?

drinkcat: Well, hopefully debootstrap will end up fixing this...

But, yeah, I can make the rule more general and try to clean that up in one of the future uprevs.

drinkcat:
V-1, breaks kali-rolling...

`2018-04-08_03-43-08_drinkcat_chroagh_unbreak-bionic_10/status.html`

dnschneid:
> # There is no bootstrap script for some distros derived from Debian. Thus we use
> # the scripts for matching upstream distros to bootstrap the derived distros.

So either patch $BOOTSTRAP_RELEASE or do the patching after the symlink is created.  I think patching after the symlink is created is the way to go.

dnschneid:
Any update on this?

PR #3718, opened 2018-04-04
For clueless non-gamers like me, wikify EXP

nealmcb:
I.e. just change README.md to add a link to the Wikipedia article for https://en.wikipedia.org/wiki/Experience_point

Note: the term "eat pie" is cute but also confusing. If there was a "description" on the "pie" label, that would help. I guess I don't have the rights to edit labels for this repo.  cf. https://help.github.com/articles/creating-a-label/

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,7 +271,8 @@ some suggestions:
   * Discovered a bug lurking within the scripts, or a papercut that bothers you
     just enough to make you want to actually do something about it? You guessed
     it: fork crouton, fix everything, and create a pull request.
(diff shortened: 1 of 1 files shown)


DennisLfromGA:
@nealmcb,

Probably not a bad idea to add a reference to EXP, I didn't know what it meant but I'm not a gamer.

I think the 'pie' label is a reference to 'easy as pie', it must be since even I have taken on some of the feature requests tagged with 'pie'.

-**DennisLfromGA**

PR #3742, opened 2018-04-23
A crouton target for Plex Media Server

wstewartii:
This is a crouton target for plex media server, a popular alternative to kodi due to it's support on a wide variety of devices such as the roku tv and the ps4.


dnschneid:
Sorry for the delay.  crouton generally tries to avoid individual apps as targets, as it's a bit of a support and testing nightmare to keep everything up-to-date.

As for an alternative to Kodi, if this were an all-in-one UI I would be a little more amenable to the idea, but in this case it's just the server backend which puts it in the same category as LAMP stacks as far as I'm concerned (unless I'm misunderstanding something).  And Plex has a perfectly functional web UI, so it doesn't really make sense to include a target for a front-end either...

PR #2145, opened 2015-09-29
Support enter-chroot in a chroot (nested enter-chroot)

AurelienLourot:
This PR implements a support for entering a crouton chroot from another chroot. 

Use case: I have two full self-contained chroots able to run their own X and desktop environments, but **sometimes** I just want to "start" `myGreatChroot1` and from there quickly enter `myGreatChroot2` and run a command using `myGreatChroot1`'s X server (i.e. without starting `myGreatChroot2`'s X server).

Here is how it now works, based on this PR.

Install this version of `enter-chroot` on the host (temporary hack until this PR gets merged):

```
localhost# cd installer
localhost# export INSTALLERDIR=`pwd`
localhost# . functions
localhost# cd ..
localhost# installscript host-bin/enter-chroot /usr/local/bin/
localhost# chmod 755 /usr/local/bin/enter-chroot
```

Create two chroots:

```
chronos@localhost$ sudo sh crouton -t xfce -n myGreatChroot1
chronos@localhost$ sudo sh crouton -t xfce -n myGreatChroot2
```

"Start" the first chroot:

```
chronos@localhost$ sudo startxfce4 -n myGreatChroot1
```

Install crouton in the first chroot:

```
(myGreatChroot1)user@localhost$ sudo apt-get install curl
(myGreatChroot1)user@localhost$ sudo sh crouton -b
(myGreatChroot1)user@localhost$ sudo mkdir /mnt/stateful_partition
(myGreatChroot1)user@localhost$ sudo mount /dev/mmcblk0p1 /mnt/stateful_partition
(myGreatChroot1)user@localhost$ sudo ln -s /mnt/stateful_partition/crouton/chroots /usr/local/chroots
(myGreatChroot1)user@localhost$ sudo ln -s /usr/sbin/chroot /usr/bin/chroot
```

Install this version of `enter-chroot` in the first chroot (temporary hack until this PR gets merged):

```
(myGreatChroot1)root@localhost# cd installer
(myGreatChroot1)root@localhost# export INSTALLERDIR=`pwd`
(myGreatChroot1)root@localhost# . functions
(myGreatChroot1)root@localhost# cd ..
(myGreatChroot1)root@localhost# installscript host-bin/enter-chroot /usr/local/bin/
(myGreatChroot1)root@localhost# chmod 755 /usr/local/bin/enter-chroot
```

From the first chroot, enter the second one:

```
(myGreatChroot1)user@localhost$ sudo enter-chroot -n myGreatChroot2
[...]

PR #3816, opened 2018-07-05
ghuser.io metadata.

AurelienLourot:
(I have signed the CLA.)

@dnschneid @drinkcat we know each other from [here](https://github.com/dnschneid/crouton/pull/2145).

We're trying to build **complete** GitHub profiles like:
* https://ghuser.io/dnschneid
* https://ghuser.io/drinkcat

You can help us by accepting this repo-metadata. Thanks!

### DIFF
--- a/.ghuser.io.json
+++ b/.ghuser.io.json
@@ -1,0 +1,4 @@
+{
+  "_comment": "Repo metadata for ghuser.io. See https://github.com/AurelienLourot/ghuser.io/blob/master/docs/repo-settings.md",
+  "avatar_url": "https://rawgit.com/AurelienLourot/crouton-articles/master/thirdparty/crouton_icon.png"
+}


drinkcat:
Can you store the database in one of your repo instead?

Adding a metadata file to every single github repo does not seem very scalable (nor easy to manage for you), and I can already tell you that most projects will not accept this.

AurelienLourot:
Thanks for pointing this out :) The idea is that each time a user wants to define an icon for a repo (in order to have it displayed on their profile and the profiles of other contributors), they can define it this way. So here I'm not acting as the maintainer of ghuser.io but as a particular user.

So I think it scales because each user will take care of adding this file to a few repos. And they don't have to, it's up to them. And also it's not necessary for repos belonging to an organization as they can get the logo of the organization. 

From your side as a maintainer of crouton, you benefit from it for your own profile as well. And also for you it's the same kind of file as a `.travis.yml` file, or a `.github/` folder, or a linter definition for example. It's just a settings file exposed by your project for a consumer.

But you're indeed right and I'll have to implement a "plan B" in case a repo owner refuses ;)

dnschneid:
I agree with @drinkcat; it doesn't really make sense for a new project to base its functionality around every project on github adding a vanity metadata file.

In this case, since the settings file doesn't improve the crouton development process, I think we'll pass.

PR #3839, opened 2018-07-16
Bundle debootstrap with the installer

dnschneid:
Let's bundle the debootstrap scripts with the installer so we don't get hid by surprise issues (except when trying to release).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,8 @@ EXTPEXE = host-ext/crouton/kiwi.pexe
 EXTPEXESOURCES = $(wildcard host-ext/nacl_src/*.h) \
 				 $(wildcard host-ext/nacl_src/*.cc)
 EXTSOURCES = $(wildcard host-ext/crouton/*)
+BOOTSTRAPS := $(wildcard installer/*/bootstrap)
+BUILDDIR = crouton.build
 GENVERSION = build/genversion.sh
 CONTRIBUTORSSED = build/CONTRIBUTORS.sed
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -47,16 +49,26 @@ else
...
 	{ \
 		sed -e "s/\$$TARPARAMS/$(TARPARAMS)/" \
 			-e "s/VERSION=.*/VERSION='$(shell $(GENVERSION) $(VERSION))'/" \
 			$(WRAPPER) \
-		&& tar --owner=root --group=root -c $(TARPARAMS) $(SCRIPTS) \
+		&& (cd $(BUILDDIR) && tar --owner=root --group=root -c $(TARPARAMS) *)\
 		&& chmod +x /dev/stdout \
 	;} > $(TARGET) || ! rm -f $(TARGET)
 
+$(BUILDDIR): $(SCRIPTS) Makefile
```
drinkcat: Missing $(BOOTSTRAPS) as dep?

drinkcat:
LGTM, just one nit.

Makefile
dnschneid: it's included in $(SCRIPTS)

drinkcat: Ack.

dbaynard:
Hello,

I've been testing an arch linux fork (based on @drinkcat's work!) and this PR makes crouton a 200MB download (as opposed to 100kB). Why should crouton bundle the bootstrap files in its final output, rather than just make sure they actually work?

PR #3840, opened 2018-07-17
Override arm64 ARCH on wheezy if not explicitly chosen

dnschneid:
Wheezy doesn't have arm64 builds, so use armhf instead.


drinkcat:
LGTM.

Issue #3847, opened 2018-07-21
Feature request: backup to stdout


shanegibbs:
I would ❤️ to be able to backup to stdout.

Today I can backup by

```
$ time sudo edit-chroot -b bionic
Unmounting /mnt/stateful_partition/crouton/chroots/bionic...
  Backing up /mnt/stateful_partition/crouton/chroots/bionic to /home/chronos/u-5011e24640c6e14822f5c585da4c6e48797d2d1a/DownloFinished backing up /mnt/stateful_partition/crouton/chroots/bionic to bionic-20180721-1148.tar.gz

real    5m0.736s
user    4m22.649s
sys     0m29.752s
$ time scp bionic-20180721-1148.tar.gz server:
bionic-20180721-1148.tar.gz                                                       100% 2092MB  22.8MB/s  24.8MB/s   01:31    

real    1m31.993s
user    0m10.866s
sys     0m16.751s
```

This compresses and writes the backup file localy. I then copy it off.

Being able to backup to stdout would enable me to backup directly to another computer. I could then avoid the CPU and IO load of compressing and writing locally.

The closest I have got so far is (I haven't looked into exactly why it fails, but having to specify `/dev/stdout` seems hacky)

```
$ sudo edit-chroot -b bionic -f /dev/stdout |ssh server 'cat > /dev/null'
Unmounting /mnt/stateful_partition/crouton/chroots/bionic...
tar: /dev/stdout: Wrote only 8192 of 10240 bytesoots/bionic to /dev/stdout...
tar: Error is not recoverable: exiting now
Unable to backup /mnt/stateful_partition/crouton/chroots/bionic.
Deleting partial archive.
```

I had a quick look at the scripts, I believe I can manually do something like this which is a lot faster in my case.

```
$ sudo unmount-chroot bionic
Unmounting /mnt/stateful_partition/crouton/chroots/bionic...
$ time bash -c "sudo tar -c --one-file-system -C /mnt/stateful_partition/crouton/chroots bionic 
| ssh server 'gzip > backup.tar.gz'"

real    2m46.778s
user    0m22.249s
sys     0m36.510s
```

Perhaps being able to specify the target file as `-` might be nice?

```
$ sudo edit-chroot -b bionic -f-
```

PR #3849, opened 2018-07-21
host-bin/edit-chroot: backup to stdout

shanegibbs:
Implementation of https://github.com/dnschneid/crouton/issues/3847

* The `spinner` writes to `stdout` so I have removed it from `stdout` backups

Question: can/do we want to handle multiple chroots?

Could we put all specified chroots in a single archive when writing to `stdout`? That would be different from existing behavior. See:

https://github.com/dnschneid/crouton/blob/5d4c301447b989c62afd03712e2dfda327bb2d3b/host-bin/edit-chroot#L45

### DIFF
--- a/host-bin/edit-chroot
+++ b/host-bin/edit-chroot
@@ -43,6 +43,7 @@ Options:
                 When used with -r, specifies the tarball to restore from.
                 If TARBALL is a directory, automatic naming is still used.
                 If multiple chroots are specified, TARBALL must be a directory.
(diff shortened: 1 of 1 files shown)


dnschneid:
Thanks for the patch.  I don't think multiple chroots should be allowed; the output will not be very useful.

Is it possible to fix spinner to output to stderr?  I'd prefer for the stdout mode to use the same codepath somehow...

PR #3862, opened 2018-08-07
Avoid letting crouton run inside crostini

dnschneid:
Reviewers: please also comment on the README and error message wording.

### DIFF
--- a/installer/main.sh
+++ b/installer/main.sh
@@ -138,7 +138,11 @@ done
 shift "$((OPTIND-1))"
 
 # Check against the minimum version of Chromium OS
-if ! awk -F= '/_RELEASE_BUILD_NUMBER=/ { exit int($2) < '"${CROS_MIN_VERS:-0}"' }' \
+if [ -n "$DOWNLOADONLY" ]; then
+    :
+elif ! grep -q '_RELEASE_BUILD_NUMBER=' /etc/lsb-release 2>/dev/null; then
+    error 2 "$APPLICATION can only be installed in the Chromium OS dev mode shell."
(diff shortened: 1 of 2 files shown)


DennisLfromGA:
LGTM 👍  ( *not that it matters of course* )
And sound great to me too. 😂

installer/main.sh
```diff
@@ -138,12 +138,16 @@ done
 shift "$((OPTIND-1))"
 
 # Check against the minimum version of Chromium OS
-if ! awk -F= '/_RELEASE_BUILD_NUMBER=/ { exit int($2) < '"${CROS_MIN_VERS:-0}"' }' \
+if [ -n "$DOWNLOADONLY" ]; then
+    :
+elif ! grep -q '_RELEASE_BUILD_NUMBER=' /etc/lsb-release 2>/dev/null; then
```
DennisLfromGA: Just curious, is this enough or should it be even more qualified with 'CHROMEOS_RELEASE_BUILD_NUMBER' here and on line 145?

dnschneid: I think that would mess up Chromium OS

DennisLfromGA: Well, I thought so too at first but after checking it seems all the Chromium OS derivatives use the 'CHROMEOS_' prefix. Here's the `lsb-release` files for CrOS, Chromium OS, Cloudready, & Flint OS (yes, I have all of them)

- [lsb-release.all.txt](https://transfer.sh/vwTCd/lsb-release.all.txt)

But I guess things can change so whatever you decide is fine with me.

Issue #2921, opened 2016-12-02
unity xenial locale issue


jaimejimbo:
when installing xenial, gnome locales aren't installed by the script. It makes gnome-terminal not to work. It is fixed when installing language-pack-gnome-en language-pack-en packages.

DennisLfromGA:
@jaimejimbo,

Do you know if this only pertains to the 'xenial' release?

-**DennisL**

jaimejimbo:
I didn't have that issue with trusty, but I haven't tested the rest of releases.

xxcriticxx:
i installed `language-pack-gnome-en` terminal still wont open i am forced to use `xterm` since day one

jaimejimbo:
if you install all packages I listed before, it should work. you can try sudo dpkg-reconfigure gnome-terminal or something like that.

xxcriticxx:
my terminal opens icon blinks gray purple gray purple and application auto close

i got few application that just blink and close by them self

jaimejimbo:
oh, my terminal didn't even open. So perhaps is a video driver issue.

xxcriticxx:
@jaimejimbo what chromebook you on? i will try installing drivers later?

alexluecke:
@jaimejimbo I think this is purely a locale issue. I am almost certain I fixed the problem by making locales were installed:

    sudo apt-get install locales
    sudo dpkg-reconfigure locales

jaimejimbo:
@xxcriticxx , @alexluecke i had to reinstall and now it doesn't work. I'm trying to fix it...

jaimejimbo:
localectl set-locale LANG="en_US.UTF-8"

That fixed it

xxcriticxx:
@jaimejimbo my still does the same blinks and auto close after language selection

jaimejimbo:
have you tried running it from xterm? you can search for error message on google, that's how i fixed my issue.

octopusnz:
I have the same issue after installing the xenial release.
You need to manually configure a default locale else gnome-terminal won't work.

I actually think it may be an issue with the default locale not being set by the installer in the xenial release (not just the gnome locale).
And my hypothesis is this is because the default locale set in the crouton script is not available when the script tries to complete this step.

My steps:

Install using something like: -r xenial -t unity
Run locale 
All default locale values appear empty
Run locale-gen - I can't be 100% sure unfortunately but I think US.UTF-8 was not installed at this stage.

To fix:

[...]

Issue #3283, opened 2017-06-16
[SOLVED] Unable to launch Gnome Terminal, Error constructing proxy for org.gnome.Terminal


trusktr:
```
chronos@localhost / $ sudo edit-chroot -all
name: xenial
encrypted: no
Entering /mnt/stateful_partition/crouton/chroots/xenial...
crouton: version 1-20170315143304~master:95589555
release: xenial
architecture: amd64
xmethod: xorg
targets: gnome
host: version 9334.72.0 (Official Build) stable-channel caroline 
kernel: Linux localhost 3.18.0-14453-g2356e6e #1 SMP PREEMPT Wed May 17 18:05:37 PDT 2017 x86_64 x86_64 x86_64 GNU/Linux
freon: yes
Not unmounting /mnt/stateful_partition/crouton/chroots/xenial as another instance is using it.
```

#### Please describe your issue:
I'm getting the same error as here: https://github.com/dnschneid/crouton/issues/1266#issuecomment-67756004

```
Error constructing proxy for org.gnome.Terminal:/org/gnome/Terminal/Factory0: Error calling StartServiceByName for org.gnome.Terminal: GDBus.Error:org.freedesktop.DBus.Error.Spawn.ChildExited: Process org.gnome.Terminal exited with status 8
```

#### If known, describe the steps to reproduce the issue:

```
sudo sh ~/Downloads/crouton -t gnome
sudo startgnome
```

then run xterm, then

```
sudo apt-get install gnome-terminal
```

then try to open Gnome Terminal but the cursor hangs loading.

Then I try running `gnome-terminal` in xterm, and it gives that error

trusktr:
Ah, oh well, startcli inside a crosh shell works nicely, I can paste into that. I was just having a hard time with the default xterm.

It might be nice to at least include a good terminal with the desktop targets. For example `gnome-terminal` with the `gnome` target, etc.

trusktr:
Looks like the problem was with locales. I was able to fix it based on this comment: https://github.com/dnschneid/crouton/issues/2921#issuecomment-266349209

I was having trouble running `mongod` (in `meteor`) which depends on having locales set up, then I thought this would probably fix gnome-terminal too and it did.

Should locales be configured by default, or maybe run dpkg-reconfigure during target installs?

DennisLfromGA:
@trusktr,

> Should locales be configured by default, or maybe run dpkg-reconfigure during target installs?

[...]

PR #3952, opened 2018-12-01
Fix locale for Ubuntu Xenial

motevets:
Previously, core checked for the absence of /etc/default/locale to
determine if a locale needed to be generated an set as the default.
Unfortunately, in Xenial /etc/default/locale exists by default in
a clean installation with a comment that reads:

    # File generated by update-locale

This commit enhances the test for /etc/default/locale to check that
the file contains `LANG=`, ensuring that a default locale is set.

Fixes: #2921 and #3283

### DIFF
--- a/targets/core
+++ b/targets/core
@@ -113,7 +113,7 @@ fixkeyboardmode
 install --minimal sudo wget ca-certificates apt-transport-https
 
 # Generate and set default locale
-if [ ! -f '/etc/default/locale' ] && hash locale-gen 2>/dev/null; then
+if ! grep "LANG=" /etc/default/locale && hash locale-gen 2>/dev/null; then
     echo 'LANG=en_US.UTF-8' > '/etc/default/locale'
     locale-gen --lang en_US.UTF-8
     if [ "${DISTROAKA:-"$DISTRO"}" = 'debian' ]; then
--- a/targets/core
+++ b/targets/core
@@ -113,7 +113,7 @@ fixkeyboardmode
 install --minimal sudo wget ca-certificates apt-transport-https
 
 # Generate and set default locale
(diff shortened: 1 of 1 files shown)


motevets:
I signed it!

targets/core
```diff
@@ -113,7 +113,7 @@ fixkeyboardmode
 install --minimal sudo wget ca-certificates apt-transport-https
 
 # Generate and set default locale
-if [ ! -f '/etc/default/locale' ] && hash locale-gen 2>/dev/null; then
+if ! grep "LANG=" /etc/default/locale && hash locale-gen 2>/dev/null; then
```
dnschneid: thanks for the patch!  grep should be used with `-q` to silence its output, and also should be used with `2>/dev/null` to avoid output in the case the file doesn't exist.

Finally, you probably want to make sure the `LANG=` line isn't commented out (e.g., if in the future the skeleton file includes an example).

Something like (untested):

`if ! grep -q '^[^#]*LANG=' /etc/default/locale 2>/dev/null && ...`

motevets:
Thank you for the suggestion. I put it in and tested it out, and it works great.

motevets:
@dnschneid do you need me to do anything else to make this merge ready?

ndarilek:
Would love to see this merged. I'm having to manually perform these steps in my chroots and it's a bit of a pain.

Thanks!

ndarilek:
FWIW, under Bionic at least I had to also run `apt-get install -y language-pack-en` before this worked. Not sure if that's also true under latest Xenial, but with that addition, I'd call this ready to merge, and probably essential since I couldn't get anything significant running under GNOME without it.

dnschneid:
@ndarilek give it a shot and if the latest crouton still needs language-pack-en, open a new issue.

Issue #3243, opened 2017-05-24
Turn off gnome


razmikb:
```
chronos@localhost / $ sudo edit-chroot -all
name: xenial
encrypted: no
Entering /mnt/stateful_partition/crouton/chroots/xenial...
crouton: version 1-20170315143304~master:95589555
release: xenial
architecture: amd64
xmethod: xorg
targets: gnome
host: version 9334.72.0 (Official Build) stable-channel auron_yuna 
kernel: Linux localhost 3.14.0 #1 SMP PREEMPT Wed May 17 17:58:32 PDT 2017 x86_64 x86_64 x86_64 GNU/Linux
freon: yes
Not unmounting /mnt/stateful_partition/crouton/chroots/xenial as another instance is using it.
```

Hi, how should I disable gnome? It doesn't have log out function (as far as I know) like other desktops.

DennisLfromGA:
@razmikb,

I've run into this a couple of times at least, most recently with an elementary desktop and found one or both of the following to work:
```
exec gnome-session-quit --no-prompt
exec gnome-session-quit --logout --force
```
I actually use both in a `logout.sh` script that simply contains the above two lines.
You have to run it from the command line or I guess you could make a `.desktop` file for it and luanch it in a gui.

*Hope this helps,*
-**DennisL**

Issue #3930, opened 2018-11-07
/prepare.sh cannot create as Directory is nonexistent


fujiyama17:
```
Please paste the output of the following command here: sudo edit-chroot -all
```
name: kali-rolling
encrypted: no
Entering /mnt/stateful_partition/crouton/chroots/kali-rolling...
crouton: version 1-20181001133934~master:38012fdf
release: kali-rolling
architecture: amd64
targets: core,audio,xfce
host: version 11151.11.0 (Official Build) dev-channel eve 
kernel: Linux localhost 4.4.159-15334-g388c74685442 #1 SMP PREEMPT Wed Oct 24 02:07:35 PDT 2018 x86_64 GNU/Linux
freon: yes
Unmounting /mnt/stateful_partition/crouton/chroots/kali-rolling...

#### Please describe your issue:

I get the following error:

Compiling CRAS (native)...
Installing CRAS...
/prepare.sh: 1166: /prepare.sh: cannot create /usr/share/alsa/alsa.conf.d/10-cras.conf: Directory nonexistent
Failed to complete chroot setup.
Unmounting /mnt/stateful_partition/crouton/chroots/kali-rolling...

#### If known, describe the steps to reproduce the issue:

I just ran:

sudo sh -e ~/Downloads/crouton -r kali-rolling -u -t core,audio,xfc

---

I have tried other targets, even just core, and this keeps coming up.  Can you assist?  Thank you!

bitozoid:
Me too.

```
$ sudo edit-chroot -all
name: kali
encrypted: no
Entering /mnt/stateful_partition/crouton/chroots/kali...
crouton: version 1-20181001133934~master:38012fdf
release: kali-rolling
architecture: amd64
targets: xiwi,cli-extra,keyboard,xfce,extension
host: version 11021.56.0 (Official Build) stable-channel peppy 
kernel: Linux localhost 3.8.11 #1 SMP Mon Oct 22 23:38:25 PDT 2018 x86_64 GNU/Linux
freon: yes
Unmounting /mnt/stateful_partition/crouton/chroots/kali...
```

I did
```
sudo sh ~/Downloads/crouton -t xiwi,cli-extra,keyboard,xfce,extension -n kali -r kali-rolling
```

and got
```
...
Installing CRAS...
/prepare.sh: 1166: /prepare.sh: cannot create /usr/share/alsa/alsa.conf.d/10-cras.conf: Directory nonexistent
Failed to complete chroot setup.
```

[...]

PR #3953, opened 2018-12-01
Always show the logout menu in gnome

motevets:
Gnome by default does not display the user section which contains the
logout option in the system menu when there is only one user. This is
a problem because there is normally only one user in a crouton chroot,
and there is no other way for the user to exit the chroot from the UI
(though it can be done using the CLI.) In this commit I install
dconf-tools and specifies a new system wide default to always show the
logout option.

I also incidentally separated the gnome packages in new lines and
alphabetized them for readability.

Fixes: #3243

### DIFF
--- a/targets/gnome
+++ b/targets/gnome
@@ -15,10 +15,32 @@ else
 fi
 
 ### Append to prepare.sh:
-install --minimal evolution-data-server gnome-control-center \
-        gnome-screensaver gnome-session $legacy_session_package \
-        gnome-shell gnome-themes-standard gvfs-backends nautilus \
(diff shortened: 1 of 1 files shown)


dnschneid:
Wow, this is obscure.  Thanks!
What releases have you tested against?

motevets:
I've tested this with Xenial, Trusty, stretch, and buster.

Xenial and stretch worked great. Trusty successfully installed, but the logout option still didn't display.  I think this is because that the `org/gnome/shell.always-show-log-outd` `dconf` setting didn't exist in older Gnome versions, but I don't think adding it hurts anything. I couldn't get buster to install because of #3930.

Issue #3947, opened 2018-11-23
decryption of newly encrypted chroot reliably fails


TimFH:
Please paste the output of the following command here: sudo edit-chroot -all
chronos@localhost ~/Downloads $ sudo edit-chroot -all
name: trusty
encrypted: yes, locked
Unmounting /mnt/stateful_partition/crouton/chroots/trusty...
name: xenial
encrypted: yes, locked
Unmounting /mnt/stateful_partition/crouton/chroots/xenial...

#### Please describe your issue:
Every time I try to encrypt my new chroot and then enter it, it says "Failed to decrypt xenial." as if I typed the passphrase wrong. Doesn't matter if I encrypt it when I create it with crouton or later with edit-chroot. Decryption does still work on old chroots. To make sure I'm not losing my mind, I cut the passphrase down to a single character that I absolutely could not possibly be typing incorrectly, and it still happens. Am I jinxed or is everyone having this problem?

TimFH:
Postscript: The following tail of /var/log/messages following an attempt to enter the encrypted chroot appear relevant:

2018-11-23T14:39:22.104987-08:00 ERR ecryptfs-unwrap-passphrase[26324]: Failed to detect wrapped passphrase version: Invalid argument
2018-11-23T14:39:22.355093-08:00 ERR ecryptfs-unwrap-passphrase[26329]: Incorrect wrapping key for file [/tmp/tmp.NqxCVTiKmm]
2018-11-23T14:39:57.813827-08:00 ERR ecryptfs-unwrap-passphrase[26363]: Failed to detect wrapped passphrase version: Invalid argument
2018-11-23T14:39:58.054423-08:00 ERR ecryptfs-unwrap-passphrase[26368]: Incorrect wrapping key for file [/tmp/tmp.nHtpdC2pvE]
2018-11-23T14:40:56.193503-08:00 ERR ecryptfs-rewrap-passphrase[26748]: Failed to detect wrapped passphrase version: Invalid argument
2018-11-23T14:41:19.673955-08:00 ERR cras_server[1557]: Unable to find the best channel map
2018-11-23T14:42:45.596298-08:00 ERR cras_server[1557]: message repeated 2 times: [ Unable to find the best channel map]
2018-11-23T14:44:11.763348-08:00 ERR ecryptfs-unwrap-passphrase[27231]: Failed to detect wrapped passphrase version: Invalid argument
[...]

PR #3978, opened 2019-01-02
Fix encrypted mount support for ecryptfs v2 wrapping (Issue #3947)

justinfrankel:


host-bin/mount-chroot
```diff
@@ -237,8 +237,10 @@ $passphrase" | ecryptfs-wrap-passphrase "$wrappedfnek" -
             addtrap "rm -f '$wrappedkey' '$wrappedfnek'"
 
             # Extract wrapped keys from keyfile
-            tail -c 160 "$KEYFILE" | head -c 80 > "$wrappedkey"
-            tail -c 80 "$KEYFILE" > "$wrappedfnek"
+            wrappedtotal="`expr $(cat "$KEYFILE" | wc -c) - $(head -n 1 "$KEYFILE" | wc -c)`"
+            wrappedsize="`expr $wrappedtotal / 2`"
+            tail -c $wrappedtotal "$KEYFILE" | head -c $wrappedsize > "$wrappedkey"
```
dnschneid: `"$wrappedtotal"`, `"$wrappedsize"`, etc

host-bin/mount-chroot
```diff
@@ -237,8 +237,10 @@ $passphrase" | ecryptfs-wrap-passphrase "$wrappedfnek" -
             addtrap "rm -f '$wrappedkey' '$wrappedfnek'"
 
             # Extract wrapped keys from keyfile
-            tail -c 160 "$KEYFILE" | head -c 80 > "$wrappedkey"
-            tail -c 80 "$KEYFILE" > "$wrappedfnek"
+            wrappedtotal="`expr $(cat "$KEYFILE" | wc -c) - $(head -n 1 "$KEYFILE" | wc -c)`"
+            wrappedsize="`expr $wrappedtotal / 2`"
```
dnschneid: `"$((wrappedtotal / 2))"`

host-bin/mount-chroot
```diff
@@ -237,8 +237,10 @@ $passphrase" | ecryptfs-wrap-passphrase "$wrappedfnek" -
             addtrap "rm -f '$wrappedkey' '$wrappedfnek'"
 
             # Extract wrapped keys from keyfile
-            tail -c 160 "$KEYFILE" | head -c 80 > "$wrappedkey"
-            tail -c 80 "$KEYFILE" > "$wrappedfnek"
+            wrappedtotal="`expr $(cat "$KEYFILE" | wc -c) - $(head -n 1 "$KEYFILE" | wc -c)`"
```
dnschneid: `$((...))` is preferred over ``` `expr ...` ```
You can also skip the cat line without causing wc to output the file name by redirecting stdin.
Combined:
`"$(($(wc -c < "$KEYFILE") - ...))`"

dnschneid:
Thanks for submitting a fix!

I wonder if it would be better to use that first line to store the size of the keys, so that we don't have to try to guess it.  We can default to "80 80" if nothing is specified.

The downside is that people who have encrypted new chroots will have to re-create them (this is ok; they never worked), but people who recently encrypted unencrypted chroots would still be up a creek without a paddle (this is not ok).

Also I'm not actually sure what people do with the first line in KEYFILE currently.  In theory, if you're storing keyfiles separately, one could use the line to add whatever comments you want, which could be useful.  Using it for metadata could mess things up.

Really what I should have done from the get-go is store the wrapped keys base64-encoded on separate lines (maybe even with variable names), instead of dumping the binary directly into the file, but too late for that.

All considered, I think the current approach is fine, and we should merge it once the syntax is fixed.

justinfrankel:
Cool, I've pushed those style changes. I think supporting existing .ecryptfs files is important -- I first encountered this when adding encryption to an existing chroot, so having it work following an update would be nice.

dnschneid:
Thanks; now we just need to run the encryption test suite on the change.

dnschneid:
Apologies for the delay.  Thanks for the fix!

Issue #3991, opened 2019-01-24
Unable to locate package dconf-tools


guidoscientia:
```
Please paste the output of the following command here: sudo edit-chroot -all
(made a mistake and deleted it but i can share install command tho;
 
sudo sh ~/Downloads/crouton -e -r sid -t core,audio,extension,gnome,keyboard,xorg,xiwi -p '/media/removable/USB Drive/Crostini'

``'

#### Please describe your issue:

Installing target gnome...
Reading package lists... Done
Building dependency tree       
Reading state information... Done
E: Unable to locate package dconf-tools
Failed to complete chroot setup.

Well that's all i can show

Asternitix:
I’m having this problem too need fix.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Asternitix:
Seems like a xiwi issue

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Asternitix:
Found a fix do an install core only and then go from there.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Asternitix:
Looks like it’s a dummy package.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Asternitix:
``

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Asternitix:
Found it. Fetch the deb file for stretch and install that. It should fix your problem. Either way xiwi will work the same.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #4033, opened 2019-04-07
Replacing 'dconf-tools' with 'dconf-cli' in targets/gnome

llzes:
Likewise as described in issue #3991, my install originally was unable to be completed when it failed to locate `dconf-tools` while installing Debian Buster. Starting with Buster `dconf-tools` was removed due to being a transitional package for downloading `dconf-cli` and `dconf-editor`, the graphical editor. From what I could tell, crouton should only need `dconf-cli`.

Until then this is how I've gone around the issue:

```
chronos@localhost / $ sudo enter-chroot -u0
Entering /mnt/stateful_partition/crouton/chroots/llzes...
(llzes)root@localhost:~# cat /etc/apt/sources.list
deb http://httpredir.debian.org/debian/ buster main non-free contrib
deb-src http://httpredir.debian.org/debian/ buster main non-free contrib
deb http://deb.debian.org/debian stretch main # Added to install dconf-tools
deb-src http://deb.debian.org/debian stretch main # Added to install dconf-tools
(llzes)root@localhost:~# sudo apt-get update
[...]
(llzes)root@localhost:~# sudo apt-get install dconf-tools
```

### DIFF
--- a/targets/gnome
+++ b/targets/gnome
@@ -17,7 +17,7 @@ fi
 ### Append to prepare.sh:
 install --minimal \
   $legacy_session_package \
-  dconf-tools \
+  dconf-cli \
   evolution-data-server \
   gnome-control-center \
   gnome-screensaver \


dnschneid:
Looks like dconf-cli doesn't exist in trusty, so that will need to have a custom case: `trusty=dconf-tools,dconf-cli` if I remember the syntax correctly

dnschneid:
I take that back. Trusty was EOL'd this past month.

dnschneid:
Thanks!

PR #4055, opened 2019-05-08
Update enter-chroot

dochench:
change from Downloads to MyFiles by default, following ChromeOS change.


dochench:
CLA signed.

dnschneid:
For compatibility and migration purposes, I think I'd rather enter-chroot just magically pick the correct location of the Downloads folder when specifying downloads share.  So stick with the downloads prefix in the config and pick the correct directory to bind mount based on what is present on the user's system.

PR #4060, opened 2019-05-10
Add ARC Sdcard support + Changed how Alsaconf handles missing configuration directory.

Hacksawfred3232:
+ Added new target "arc" + Modified enter-chroot and post-common to integrate this new target.
This target enables support for accessing the virtual sdcard used by the App Run-time for Chrome. By default, most of the files that the files apps interacts with is stored in here.

+ Changed how alsaconf deals with "non-existing" directories
Due to a bug that i can not determine the origin of, ALSA may drop the 10-cras.conf file into /tmp as a incorrectly triggered fail-safe.
A through analysis shows the following:
  - The directories are present, but the installer logic fails to detect it on the first run of that chroot's setup process.
  - Additionally, while /etc/alsa/conf.d is always first to be checked, the actual format is that all conf files are symlinked to their respective counterparts
    in /usr/share/alsa/alsa.conf.d
I have made the following change in accordance:
  - I have changed the default action for no directories found. Instead of changing the alsaconf var to /tmp, the directories are forcefully made and the default
    structure is installed. However, if the checks pass (Most likely due to the chroot being updated, which means the directories already exist), then the previous
    functionality is unaltered.

Additional post-commit notes:
Apologies for the terrible spelling in my original commit message. As it turns out, working on a Chromebook with the android apps is a bit wonky. And apparently CTRL + C and CTRL + V does not work. So hooray for that. At least the code works, so there is that.
Anyway, with that out of the way, does anyone know why the installer is bugging out around the alsaconf bit? As I referenced in my commit, I could not determine what was causing the problem of the installer not detecting the ALSA configuration directory. Maybe someone with more knowledge of bash might know, cause I don't.

### DIFF
--- a/host-bin/enter-chroot
+++ b/host-bin/enter-chroot
@@ -361,6 +361,11 @@ for m in /lib/modules/*; do
     fi
 done
 
+if grep -q "^arc$" "$CHROOTS/$NAME/.crouton-targets" 2>/dev/null; then #If we have the ARC target installed, bindmount the virtual SDCard.
+    bindmount /run/arc/sdcard/write/emulated/0 /var/host/sdcard/write
+    bindmount /run/arc/sdcard/read/emulated/0  /var/host/sdcard/read
+fi
+
 # Add a shm symlink to our new /var/run
 ln -sfT /dev/shm "`fixabslinks '/var/run'`/shm"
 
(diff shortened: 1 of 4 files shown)


Hacksawfred3232:
CLA has been signed.

Hacksawfred3232:
I signed it!

Hacksawfred3232:
Nevermind, i think i broke the commit. Marking this as closed and modifying title to reflect this.

PR #4061, opened 2019-05-10
Added "Arc" target + Changed how Alsaconf deals with "non-existing" directories.

Hacksawfred3232:
- Added new target "arc" + Modified enter-chroot and post-common to integrate this new target.
This target enables support for accessing the virtual sdcard used by the App Run-time for Chrome.
By default, most of the files that the files apps interacts with is stored in here.

- Changed how alsaconf deals with "non-existing" directories
Due to a bug that i can not determine the origin of, ALSA may drop the 10-cras.conf file into /tmp as a incorrectly triggered fail-safe.
A through analysis shows the following:
  - The directories are present, but the installer logic fails to detect it on the first run of that chroot's setup process.
  - Additionally, while /etc/alsa/conf.d is always first to be checked, the actual format is that all conf files are symlinked to their respective counterparts
   in /usr/share/alsa/alsa.conf.d

I have made the following change in accordance:
  - I have changed the default action for no directories found. Instead of changing the alsaconf var to /tmp, the directories are forcefully made and the default
   structure is installed. However, if the checks pass (Most likely due to the chroot being updated, which means the directories already exist), then the config
   file is written to the existing directory.

Additional post-commit notes:
Does anyone know why the installer is bugging out around the alsaconf bit? As I referenced in my commit, I could not determine what was causing the problem of the installer not detecting the ALSA configuration directory. Maybe someone with more knowledge of bash might know, cause I don't.

### DIFF
--- a/host-bin/enter-chroot
+++ b/host-bin/enter-chroot
@@ -361,6 +361,11 @@ for m in /lib/modules/*; do
     fi
 done
 
+if grep -q "^arc$" "$CHROOTS/$NAME/.crouton-targets" 2>/dev/null; then #If we have the ARC target installed, bindmount the virtual SDCard.
+    bindmount /run/arc/sdcard/write/emulated/0 /var/host/sdcard/write
+    bindmount /run/arc/sdcard/read/emulated/0  /var/host/sdcard/read
+fi
+
 # Add a shm symlink to our new /var/run
 ln -sfT /dev/shm "`fixabslinks '/var/run'`/shm"
 
--- a/targets/arc
+++ b/targets/arc
@@ -1,0 +1,22 @@
+#!/bin/sh -e
+# Copyright (c) 2016 The Chromium OS Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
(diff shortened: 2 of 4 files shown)


DennisLfromGA:
I would suggest submitting the arc target and alsaconf fix as separate pull requests, I think they'd have a better chance of getting merged that way.

Hacksawfred3232:
Concerning that i have no real idea on how to do that, i will pass. Unless you know how to split commits into two different pull requests?

Hacksawfred3232:
@DennisLFromGA I guess dnschnied is currently tied up with other things? If so, I can make a release branch on my end for my personal use for now while you guys sort things out.

DennisLfromGA:
@Hacksawfred3232,

Yeah, that might be best in the interim at least.
I'm just a contributor, not a real crouton dev so can't help much here, sorry.

-**DennisLfromGA**

Hacksawfred3232:
Late response (Been busy with exams), but i have gone ahead and updated my version of the releases branch. I guess nothing else can be done for now, as dnschnied seems to be busy with other projects.

PR #4114, opened 2019-08-02
We actually need to create the metric directory ourselves

dnschneid:
https://bugs.chromium.org/p/chromium/issues/detail?id=989219

@hardikgoyal


drinkcat:
CR+1, would be good to have Hardik confirm.

PR #4170, opened 2019-11-25
removed unnecessary "-m 755"

cubetronic:
755 are the permissions by default, so it is unnecessary to write this.
man install:
```
        -m, --mode=MODE
              set permission mode (as in chmod), instead of rwxr-xr-x
```


cubetronic:
install command behavior based on
```
install --version
install (GNU coreutils) 8.31
```
which is a new version. you may want to check it for older versions

dnschneid:
I don't recall it being consistent on old versions...and at this point it's too late to change the internet again :)

Issue #4051, opened 2019-05-03
Downloads folder on chrome os side is now MyFiles, how to mount in crouton?


dochench:
```
Please paste the output of the following command here: sudo edit-chroot -all
```
name: bionic
encrypted: no
Entering /mnt/stateful_partition/crouton/chroots/bionic...
crouton: version 1-20190403182822~master:174af0eb
release: bionic
architecture: amd64
xmethod: xorg
targets: audio,xorg,extension,xiwi,gnome-desktop,touch
host: version 12105.25.0 (Official Build) dev-channel eve 
kernel: Linux localhost 4.4.178-16492-g4245c5e16521 #1 SMP PREEMPT Thu May 2 00:43:09 PDT 2019 x86_64 x86_64 x86_64 GNU/Linux
freon: yes
Unmounting /mnt/stateful_partition/crouton/chroots/bionic...

#### Please describe your issue:
Inside the chroot in /etc/crouton/shares there are only three options for automounting.
# The only directories that can be mapped from outside of the chroot are
# subdirectories of the following locations:
#  downloads: ~/Downloads
#  shared: /mnt/stateful_partition/crouton/shared
#  encrypted: ~/crouton/shared

It looks as if the ~/Downloads folder has been relocated to ~/MyFiles/Downloads on the chrome OS side of things.

How should this now be accessed? Symlinking?

DennisLfromGA:
@dochench,

On my Pixelbook using version 74.0.3729.125 (Official Build) beta (64-bit) with support for new files and folders under the “My files” local root I have both (identical) folders listed below:

`/home/chronos/user/Downloads`
`/home/chronos/user/MyFiles/Downloads`

So you don't need to symlink Downloads since crouton still bind mounts it in `~/Downloads` but any *other* folders besides Downloads that you create under `/home/chronos/user/MyFiles` will not be accessible in your chroot IIUC. 
I believe that the `enter-chroot` script will have to be updated to include 'MyFiles' for those other folders to be shared in your chroot. The crouton developers will have to decide if this is important enough to warrant changes.

I don't understand the reason to create folders under MyFiles and outside of Downloads since they are still a part of the user's profile and subject to getting wiped with a powerwash, recovery, or profile corruption but I guess there are some valid reasons since they went to great lengths to add this feature. Maybe I'll understand it better after using it more.

*Hope this helps,*
-**DennisLfromGA**

dochench:
Fixed. :)

[...]

PR #4312, opened 2020-07-18
Add "MyFiles" mapping to enter-chroot

stevedesmond-ca:
This commit adds a mapping to the "MyFiles" directory from the host user's home directory, and includes it by default in the chroot's `/etc/crouton/shares` file.

Adapted from https://github.com/dnschneid/crouton/issues/4051#issuecomment-602163403 by @levinit

### DIFF
--- a/host-bin/enter-chroot
+++ b/host-bin/enter-chroot
@@ -412,6 +412,7 @@ elif [ ! -f "$shares" ]; then
 # subdirectories of those mounts.
 # The only directories that can be mapped from outside of the chroot are
 # subdirectories of the following locations:
+#  myfiles: ~/MyFiles
 #  downloads: ~/Downloads
 #  shared: /mnt/stateful_partition/crouton/shared
 #  encrypted: ~/crouton/shared
@@ -434,10 +435,16 @@ elif [ ! -f "$shares" ]; then
 
 # Share the downloads folder of the current user's profile
 downloads ~/Downloads
+# Share the "My Files" folder of the current user's profile
+myfiles ~/MyFiles
 EOF
 fi
 # Bind-mount the stuff in $CHROOT/etc/crouton/shares, unless NOLOGIN is set
 if [ -z "$NOLOGIN" -a -f "$shares" ]; then
+    localmyfiles='/home/chronos/user/MyFiles'
+    if [ ! -d "$localmyfiles" ]; then
+        localmyfiles=''
+    fi
     localdownloads='/home/chronos/user/Downloads'
     if [ ! -d "$localdownloads" ]; then
         localdownloads=''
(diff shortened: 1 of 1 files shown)


stevedesmond-ca:
(since #4055 seems abandoned)

DennisLfromGA:
@stevedesmond-ca,

I would not call #4055 *abandoned*, the crouton author stated his view of the issue here:
> ***For compatibility and migration purposes***, I think I'd rather enter-chroot just magically pick the correct location of the Downloads folder when specifying downloads share. So stick with the downloads prefix in the config and pick the correct directory to bind mount based on what is present on the user's system.

I guess we'll see if that view has changed...

-**DennisLfromGA**

stevedesmond-ca:
I more meant the PR hasn't been touched in a year since that comment.

This PR goes beyond just downloads, e.g. I have a "code" directory in MyFiles, and this way I can access it from Chrome OS, termina/penguin, and now crouton too.

DennisLfromGA:
@stevedesmond-ca,

> I more meant the PR hasn't been touched in a year since that comment.

Yes, PR #4055 was *closed* after that comment so it normally would no longer be *touched*.

Granted, this is a little different approach, instead of replacing Downloads you're adding MyFiles.
With Crostini in the mix now it does seem to change things a bit so, I guess we shall see what we shall see.

-**DennisLfromGA**

stevedesmond-ca:
FWIW, if access from the host Chrome OS is not required, e.g. just shared between crouton and crostini, the crostini home directory can be accessed from within crouton via `/var/host/media/fuse/crostini_.../` -- I think this alternative is acceptable for my current use case, should this PR be rejected.

host-bin/enter-chroot
```diff
@@ -434,10 +435,16 @@ elif [ ! -f "$shares" ]; then
 
 # Share the downloads folder of the current user's profile
 downloads ~/Downloads
+# Share the "My Files" folder of the current user's profile
+myfiles ~/MyFiles
```
dnschneid, line 439: I don't think we should be adding this by default, especially on older systems that don't have the directory.

stevedesmond-ca: Makes sense -- added another commit removing the default, feel free to squash them if preferred!

DennisLfromGA:
@stevedesmond-ca 

> This PR goes beyond just downloads, e.g. I have a "code" directory in MyFiles, and this way I can access it from Chrome OS, termina/penguin, and now crouton too.

I just updated and added the myfiles share.
The crouton chroot can access some of the folders in MyFiles but not all. Missing are 'Linux files' & 'Play files'
The penguin container can access Downloads only since that's all that's mounted under `/mnt/chromeos/MyFiles`
See below:

Screenshot 2020-07-28 at 7 13 12 PM

So while it's an improvement it's not giving access to everything exposed in CrOS - for me at least.

*Hope this helps,*
-**DennisLfromGA**

Issue #4363, opened 2020-10-25
Scripts like `xiwi` and `startxiwi` to use `sommelier` instead of `xiwi`


kapilhp:
For a while now `xiwi` windows seem to "disappear" (not entirely!) when the system sleeps. Morever, `sommelier` windows seem to work better.

I have added some scripts to the [sommelier page on the wiki](https://github.com/dnschneid/crouton/wiki/Sommelier-(A-more-native-alternative-to-xiwi)) which can serve to use `sommelier` instead of `xiwi`.

I have found them useful. Perhaps others will too. So with appropriate modifications they could be added to the installation.

[runsommelier.gz](https://github.com/dnschneid/crouton/files/5434886/runsommelier.gz)
[startsommelier.gz](https://github.com/dnschneid/crouton/files/5434887/startsommelier.gz)

satmandu:
FYI the current version of sommelier works fine too, and also gives you graphics acceleration.

See https://github.com/skycocker/chromebrew/pull/4450

```
chronos@localhost ~ $ glxinfo -B
name of display: :0
display: :0  screen: 0
direct rendering: Yes
Extended renderer info (GLX_MESA_query_renderer):
    Vendor: Intel Open Source Technology Center (0x8086)
    Device: Mesa DRI Intel(R) UHD Graphics 615 (AML-KBL) (0x591c)
    Version: 20.2.1
    Accelerated: yes
    Video memory: 3058MB
    Unified memory: yes
    Preferred profile: core (0x1)
    Max core profile version: 4.6
    Max compat profile version: 3.0
    Max GLES1 profile version: 1.1
    Max GLES[23] profile version: 3.2
OpenGL vendor string: Intel Open Source Technology Center
OpenGL renderer string: Mesa DRI Intel(R) UHD Graphics 615 (AML-KBL)
OpenGL core profile version string: 4.6 (Core Profile) Mesa 20.2.1
OpenGL core profile shading language version string: 4.60
OpenGL core profile context flags: (none)
OpenGL core profile profile mask: core profile

OpenGL version string: 3.0 Mesa 20.2.1
OpenGL shading language version string: 1.30
OpenGL context flags: (none)

OpenGL ES profile version string: OpenGL ES 3.2 Mesa 20.2.1
OpenGL ES profile shading language version string: OpenGL ES GLSL ES 3.20

chronos@localhost ~ $ 
```

kapilhp:
[...]

PR #4390, opened 2021-01-04
Add a new target `somm` to address #4363

kapilhp:
Hope this is the right way to do things!

I created a new target `somm`.
1. Compiles and installs binary executable `/usr/local/bin/sommelier`
2. Installs a script `/usr/local/bin/runsommelier` in the chroot which sets up environment variables to run `sommelier`.
3. Installs a script `/usr/local/bin/startsommelier` in ChromeOS which is the alternative to `startxiwi`.
4. Installs `/usr/local/etc/sommelierrc` which is required by `sommelier`.

This uses `sommelier` to interact with ChromeOS Wayland compositor and start graphical programs as windows within ChromeOS without using the `crouton` extension.

As of now it works for me. The output of `glxinfo` suggests that it supports some acceleration. On the other hand the output of `vainfo` is not encouraging!

All of this is based on work done by `satmandu` over in the `chromebrew` repository.

Comments and suggestions welcome.

### DIFF
--- a/chroot-bin/runsommelier
+++ b/chroot-bin/runsommelier
@@ -17,6 +17,7 @@ if [ -z "$SOMM" ]; then
        --data-driver=noop --display=$WAYLAND_DISPLAY \
        --xwayland-path=/usr/bin/Xwayland \
        --accelerators=$SOMMELIER_ACCELERATORS \
+       --xwayland-gl-driver-path=/usr/lib/x86_64-linux-gnu/dri/ \
        --no-exit-with-child \
        /bin/sh -c "/usr/local/etc/sommelierrc" > /tmp/sommelier.log 2>&1 &
     sleep 3
--- a/chroot-bin/runsommelier
+++ b/chroot-bin/runsommelier
@@ -1,0 +1,37 @@
+#!/bin/sh
+GDK_BACKEND=x11
+CLUTTER_BACKEND=wayland
+XDG_RUNTIME_DIR=/var/run/chrome
+WAYLAND_DISPLAY=wayland-0
+DISPLAY=:0
+SCALE=1
+export GDK_BACKEND CLUTTER_BACKEND XDG_RUNTIME_DIR WAYLAND_DISPLAY DISPLAY SCALE
+
+SOMMELIER_ACCELERATORS="Super_L,<Control>space,<Alt>space,<Alt>bracketleft,<Alt>bracketright,<Alt>minus,<Alt>equal"
+
+SOMM=$(pidof sommelier 2> /dev/null)
+if [ -z "$SOMM" ]; then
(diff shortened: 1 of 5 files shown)


supechicken:
@kapilhp Can you change `startsommelier` to `startsomm`? (It may overwrite the Chromebrew one if it installed)

kapilhp:
@supechicken  There is a new branch `startsomm` on my forked repository with the relevant changes.

However, I am not sure that it is required. I am not using `chromebrew` so I cannot check. I thought only one running sommelier is required provided that within `crouton` you have access to `/var/run/chrome/wayland-0` and the X socket.

supechicken:
> @supechicken There is a new branch `startsomm` on my forked repository with the relevant changes.
> 
> However, I am not sure that it is required. I am not using `chromebrew` so I cannot check. I thought only one running sommelier is required provided that within `crouton` you have access to `/var/run/chrome/wayland-0` and the X socket.

You may also need to stop the chromebrew sommelier server if it started (you also can use the chromebrew one if it installed)

The `sommelier` required the `WAYLAND_DISPLAY` and `XDG_RUNTIME_DIR` env variable only :)

supechicken:
In `targets/somm, line 40`, Does `weston` is a dependency of `sommelier` ?

kapilhp:
@supechicken Thanks for pointing this out. I put it in to pull in some dependencies of `sommelier`.

I am still not entirely sure how to ensure that dependencies of `sommelier` are kept installed so that it does not break. Perhaps the best way would be to build a dummy package with all the dependencies.

supechicken:
In `targets/somm, line 59`: These three lines can replace by `/usr/bin/install -Dt /usr/local/bin/ -m 755 build/sommelier`

supechicken:
> In `targets/somm, line 59`: These three lines can replace by `/usr/bin/install -Dt /usr/local/bin/ -m 755 build/sommelier`

Why not install it with `ninja` ? 
```shell
ninja install -C build
```

supechicken:
> @supechicken Thanks for pointing this out. I put it in to pull in some dependencies of `sommelier`.
> 
> I am still not entirely sure how to ensure that dependencies of `sommelier` are kept installed so that it does not break. Perhaps the best way would be to build a dummy package with all the dependencies.

You can take a look of  the `depends_on` part in `skycocker/chromebrew/packages/sommelier.rb`

kapilhp:
The target is currently broken as upstream `sommelier` changed!

1. We should tie the target to a specific git hash of the upstream repository.
2. Need to get the newer version of sommelier working with `crouton`.

kapilhp:
The above commit should fix the problem by using git-versioned `sommelier.tar.gz` and `virtwl.h`.

kapilhp:
Renamed the branch so that the current HEAD of my fork can be used from there!

PR #4410, opened 2021-02-07
Add a new target `sommelier` (changed from `somm`)

kapilhp:
I created a new target somm.

1. Donwloads a suitable git version of sommelier.tar.gz and virtwl.h from Chromium sources for use with crouton. (This utility was meant for crostini.)
1. Compiles and installs binary executable /usr/local/bin/sommelier
1. Installs a script /usr/local/bin/runsommelier in the chroot which sets up environment variables to run sommelier.
1. Installs a script /usr/local/bin/startsommelier in ChromeOS which is the alternative to startxiwi.
1. Installs /usr/local/etc/sommelierrc which is required by sommelier.

This uses sommelier to interact with ChromeOS Wayland compositor and start graphical programs as windows within ChromeOS without using the crouton extension.

As of now it works for me. The output of glxinfo suggests that it supports some acceleration. On the other hand the output of vainfo is not encouraging!

All of this is based on work done by satmandu over in the chromebrew repository.

Comments and suggestions welcome.


chroot-bin/runsommelier
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh
```
dnschneid, line 1: I'm tempted to call this just `sommelier` (and then provide the full path to the actual sommelier executable inside).  Any opinions?

targets/somm
```diff
@@ -0,0 +1,69 @@
...
+### Append to prepare.sh:
+XMETHOD="${XMETHOD:-sommelier}"
+
+# Download the latest sommelier
+urlbase="https://chromium.googlesource.com/chromiumos/platform2/+archive/fef7f012df0e5ec865f94255c10503ec24d9a370/vm_tools"
+
+# Download the corresponding virtwl.h
+urlvirtwl="https://chromium.googlesource.com/chromiumos/third_party/kernel/+/7f7629cc680bf3c22d547fb5333e521af58c6c9e/include/uapi/linux/virtwl.h?format=TEXT" 
+
+DUMMYBUILDTMP="`mktemp -d crouton-cras.XXXXXX --tmpdir=/tmp`"
```
dnschneid, line 22: `crouton-sommelier`

targets/somm
```diff
@@ -0,0 +1,69 @@
...
+
+addtrap "rm -rf --one-file-system '$DUMMYBUILDTMP'"
+
+echo "Download sommelier..." 1>&2
```
dnschneid, line 26: `Downloading`

targets/somm
```diff
@@ -0,0 +1,69 @@
...
+
+superinsecure=''
+if release -le wheezy; then
+     # Wheezy doesn't have the right CA in ca-certificates that validates x.org
```
dnschneid, line 30: is the `superinsecure` hack necessary for chromium?

targets/somm
dnschneid, line 22: variable should be SOMMELIERBUILDTMP ("dummy" was for the dummy xserver)

targets/somm
```diff
@@ -0,0 +1,69 @@
...
+     superinsecure='--no-check-certificate'
+fi
+wget $superinsecure -O "$DUMMYBUILDTMP/dummy.tar.gz" \
```
dnschneid, line 33: sommelier.tar.gz

targets/somm
```diff
@@ -0,0 +1,69 @@
...
+     "$urlbase/sommelier.tar.gz"
+
+mkdir $DUMMYBUILDTMP/linux
```
dnschneid, line 36: quote

targets/somm
```diff
@@ -0,0 +1,69 @@
...
+wget $superinsecure -O- "$urlvirtwl" \
+    | base64 --decode> $DUMMYBUILDTMP/linux/virtwl.h
```
dnschneid, line 38: space around >, and quote

targets/somm
```diff
@@ -0,0 +1,69 @@
...
+
+install xwayland weston xorg
+
+install --minimal --asdeps pkg-config libwayland-dev libgbm-dev gcc libx11-xcb-dev libsystemd-dev libxcb-composite0-dev libxkbcommon-dev libxrender-dev libxtst-dev libpixman-1-dev ninja-build meson cmake libdrm-dev build-essential
```
dnschneid, line 42: wrap long lines

targets/somm
```diff
@@ -0,0 +1,69 @@
+#!/bin/sh -e -v -x
+# Copyright (c) 2016 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+REQUIRES='core'
+PROVIDES='x11'
+DESCRIPTION='Wayland backend running in a Chromium OS window.'
+HOSTBIN='startsommelier'
+CHROOTBIN='runsommelier'
+CHROOTETC='sommelierrc'
```
dnschneid, line 10: are there any platforms where the wayland server isn't available?

dnschneid (changes requested):
Thank you soooo much for your continued work on this!

All of the X11 methods in crouton allow for starting with a desktop environment -- i.e., they are able to create a root window that the environment can use as its display, and then chain on its own window manager, etc.  Is there a way to do this using sommelier?

If so, then with that implemented, the x11 target should be updated to select sommelier when it's available, instead of xiwi.

targets/somm
```diff
@@ -0,0 +1,69 @@
+#!/bin/sh -e -v -x
```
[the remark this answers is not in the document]
kapilhp: > All of the X11 methods in crouton allow for starting with a desktop environment -- i.e., they are able to create a root window that the environment can use as its display, and then chain on its own window manager, etc. Is there a way to do this using sommelier?

The `Xwayland` that is started by sommelier is a "rootless" server. So window management seems to be under "ash". One *can* start *another* `Xwayland` (as DISPLAY=:1) and run a window manager under it. This occupies the whole screen. I was able to run `i3` this way. Hardware acceleration was not checked.

I tried running `sway` under sommelier. However, it complains about the wayland version that is supported by sommelier. (On another system running Debian, I was able to run `sway` under `sway`.) So perhaps one needs to see if a newer version of sommelier can do that.

targets/somm
```diff
@@ -0,0 +1,69 @@
...
+HOSTBIN='startsommelier'
+CHROOTBIN='runsommelier'
+CHROOTETC='sommelierrc'
+. "${TARGETSDIR:="$PWD"}/common"
+
+### Append to prepare.sh:
+XMETHOD="${XMETHOD:-sommelier}"
+
+# Download the latest sommelier
+urlbase="https://chromium.googlesource.com/chromiumos/platform2/+archive/fef7f012df0e5ec865f94255c10503ec24d9a370/vm_tools"
```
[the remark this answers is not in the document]
kapilhp: Will do.

targets/somm
```diff
@@ -0,0 +1,69 @@
+#!/bin/sh -e -v -x
+# Copyright (c) 2016 The crouton Authors. All rights reserved.
```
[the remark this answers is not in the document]
kapilhp: OK! I wasn't sure, so I changed all of them back from 2020!

targets/somm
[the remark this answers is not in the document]
kapilhp: Regarding the target name, @supechicken wanted the name to be shortened, so I did that! Will do.

kapilhp:
@dnschneid I have made many of the changes suggested. Unfortunately, not all as explained below.

1. The file `sommelierrc`  is taken from a [ChromiumOS source](https://chromium.googlesource.com/chromiumos/containers/sommelier/+/refs/heads/master/sommelierrc), so I am not sure how much to change it! In particular, there does not seem to be any reference to `.xwlrc` except in this file.
2. I did not manage to figure out how to change the `runsommelier` script avoid the 'sleep 3'.

Regarding your questions:

1. I don't know of releases supported by `crouton` where wayland support is absent.
2. As mentioned in [my response](https://github.com/dnschneid/crouton/issues/4363#issuecomment-776554767) to @DennisLfromGA 's [suggestion](https://github.com/dnschneid/crouton/issues/4363#issuecomment-776286752) in #4363, it *is* possible to start a different Xwayland server under sommelier and run a window manager on that server.

Please do let me know if I have missed anything!

cvmiller:
Thanks for all the good work you guys are doing.

Will we (the Crouton user) be able to just update our crouton with the new target 'somm' (instead of xiwi) or should be try this out on a new chroot?

chroot-bin/runsommelier
```diff
@@ -0,0 +1,55 @@
...
+# SOMMELIER_ACCELERATORS and GL_DRIVER_PATH
+
+DEFAULT_SOMMELIER_ACCELERATORS="Super_L,<Control>space,<Alt>space,<Alt>bracketleft,<Alt>bracketright,<Alt>minus,<Alt>equal"
+SOMMELIER_ACCELERATORS="${SOMMELIER_ACCELERATORS:=$DEFAULT_SOMMELIER_ACCELERATORS}"
+
+# Find the architecture specific libgl1-mesa-dri library location
+FOUND_DRIVER_PATH="$(grep '\/dri$' /var/lib/dpkg/info/libgl1-mesa-dri*list | head -1 )"
+GL_DRIVER_PATH="${GL_DRIVER_PATH:=$FOUND_DRIVER_PATH}"
+
+SOMMELIER="$(pidof sommelier 2> /dev/null)" || SOMMELIER=""
```
dnschneid, line 29: `SOMMELIER="$(pidof sommelier 2> /dev/null || true)"

chroot-bin/runsommelier
```diff
@@ -0,0 +1,41 @@
...
+    sommelier -X --x-display=$DISPLAY --scale=$SCALE \
+       --glamor --drm-device=/dev/dri/renderD128  \
+       --virtwl-device=/dev/null --shm-driver=noop \
+       --data-driver=noop --display=$WAYLAND_DISPLAY \
+       --xwayland-path=/usr/bin/Xwayland \
+       --accelerators=$SOMMELIER_ACCELERATORS \
+       --xwayland-gl-driver-path=$GL_DRIVER_PATH \
+       --no-exit-with-child \
+       /bin/sh -c "/usr/local/etc/sommelierrc" > /tmp/sommelier.log 2>&1 &
+    sleep 3
```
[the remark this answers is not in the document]
dnschneid: > I did not manage to figure out how to change the runsommelier script avoid the 'sleep 3'.

Does /tmp/crouton-sommelier.log have a characteristic string that gets printed when sommelier is successfully launched?

If so, you can use something like `timeout 10 tail -f /tmp/crouton-sommelier.log | grep -q 'the string'` and use the exit code to determine success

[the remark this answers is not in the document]
dnschneid: and once that's figured out, use $! to update SOMMELIER rather than running pidof again

chroot-etc/sommelierrc
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh -e
+# Copyright (c) 2021 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# Based on https://chromium.googlesource.com/chromiumos/containers/sommelier/+/refs/heads/master/sommelierrc
+
+if hash xsetroot > /dev/null 2>&1; then
```
dnschneid, line 8: just `hash xsetroot 2>/dev/null`

chroot-bin/runsommelier
```diff
@@ -0,0 +1,55 @@
+#!/bin/sh -eu
+# Copyright (c) 2021 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+CLUTTER_BACKEND=wayland
+DISPLAY=:0
+GDK_BACKEND=x11
+SCALE=1
```
dnschneid, line 9: should this also be user-overridable?

chroot-bin/runsommelier
```diff
@@ -0,0 +1,55 @@
+#!/bin/sh -eu
+# Copyright (c) 2021 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+CLUTTER_BACKEND=wayland
+DISPLAY=:0
```
dnschneid, line 7: other x targets have logic to select the first available DISPLAY, iirc; should copy that, or use a structure similar to `xiwi`

chroot-bin/runsommelier
```diff
@@ -0,0 +1,55 @@
+#!/bin/sh -eu
```
dnschneid, line 1: Is there any reason someone would want to run sommelier directly?  seems like we should be installing sommelier to /usr/bin and this script to /usr/local/bin as just "sommelier", and then putting the absolute path in the chained execution below

dnschneid (changes requested):
Added some more comments; there are also some unresolved comments still from the previous round.

I think we need to figure out how to use this as a generic, rooted X server (similar to how xiwi has two functions) and properly integrate into the xinit script so that all usages are covered.

targets/sommelier
```diff
@@ -0,0 +1,74 @@
...
+# Copyright (c) 2021 The crouton Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+REQUIRES='core'
+PROVIDES='x11'
+DESCRIPTION='Wayland backend running in a Chromium OS window.'
+HOSTBIN='startsommelier'
+CHROOTBIN='runsommelier'
+CHROOTETC='sommelierrc'
+. "${TARGETSDIR:="$PWD"}/common"
```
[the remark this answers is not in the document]
kapilhp: According to [this commit path history](https://chromium.googlesource.com/chromiumos/platform2/+log/refs/heads/release-R49-7834.B/libchromeos-ui) it seems that wayland support was available in R49-7834.B but not earlier.

The date of the commit is "Fri Mar 18 22:03:53 2016" which also seems to jive with the introduction of `host-wayland` in `crouton` later that year by @dreveman.

How does one add the check for Chromium OS version?

targets/sommelier
```diff
@@ -0,0 +1,74 @@
...
+install --minimal --asdeps pkg-config libwayland-dev libgbm-dev gcc \
+    libx11-xcb-dev libsystemd-dev libxcb-composite0-dev \
+    libxkbcommon-dev libxrender-dev libxtst-dev libpixman-1-dev \
+    ninja-build meson cmake libdrm-dev build-essential
+
+# Compilation subshell
+(
+    cd "$SOMMELIERBUILDTMP"
+    # -m prevents "time stamp is in the future" spam
+    tar -xzmf sommelier.tar.gz
```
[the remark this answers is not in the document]
kapilhp: When we start `Xwayland` without `-rootless` it is a fullscreen window in the screen that currently has the focus. In principle, this should be possible since when I connect an external monitor which is differently shaped, the window was not fullscreen any more.  However, I did not manage resize it or specify a size. I have inadequate wayland-foo and the window is borderless :-(

I can try to create the argument processing patch.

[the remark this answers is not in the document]
kapilhp: Regarding `sommelierrc`, I suppose it should then be OK to pull it from git-versioned google sources instead of creating a duplicate.

In that case, does it go in `/usr/local/etc/sommelierrc` or `/etc/crouton/sommelierc`?

chroot-bin/runsommelier
kapilhp: We could follow the approach taken by the vm-tools in chromiumos/Debian and install the binary as `sommelier.real` and the script as `sommelier`. If the former is installed in `/usr/local/lib` then it would not interfere with command completion.

Anyway, I agree that calling the runnable script `sommelier` instead of `runsommelier` makes more sense.

targets/sommelier
```diff
@@ -0,0 +1,74 @@
...
+
+# Download the versioned sommelier
+sommelier_version="fef7f012df0e5ec865f94255c10503ec24d9a370"
+urlbase="https://chromium.googlesource.com/chromiumos/platform2/+archive/$sommelier_version/vm_tools"
+
+# Download the corresponding virtwl.h
+kernel_version="7f7629cc680bf3c22d547fb5333e521af58c6c9e"
+urlvirtwl="https://chromium.googlesource.com/chromiumos/third_party/kernel/+/$kernel_version/include/uapi/linux/virtwl.h?format=TEXT" 
+
+DUMMYBUILDTMP="`mktemp -d crouton-sommelier.XXXXXX --tmpdir=/tmp`"
```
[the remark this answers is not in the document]
kapilhp: commit 0aff4d1533fc61c4945c7357c19b28efc4f1c42d

targets/sommelier
```diff
@@ -0,0 +1,74 @@
...
+
+addtrap "rm -rf --one-file-system '$DUMMYBUILDTMP'"
```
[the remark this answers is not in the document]
kapilhp: commit 0aff4d1533fc61c4945c7357c19b28efc4f1c42d

chroot-bin/runsommelier
```diff
@@ -0,0 +1,55 @@
...
+# found in the LICENSE file.
+
+CLUTTER_BACKEND=wayland
+DISPLAY=:0
+GDK_BACKEND=x11
+SCALE=1
+WAYLAND_DISPLAY=wayland-0
+XDG_RUNTIME_DIR=/var/run/chrome
+
+export CLUTTER_BACKEND
```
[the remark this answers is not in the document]
kapilhp: commit 0aff4d1533fc61c4945c7357c19b28efc4f1c42d

chroot-bin/runsommelier
kapilhp: commit aa3b116135e4e667a8aa5b07d14f4be6bf4ef6b3

targets/sommelier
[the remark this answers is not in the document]
kapilhp: commit 41dfd2e7fed7f2e006c7c95d29a5cd0f09e18b90 
installs ChromiumOS's `sommelierrc`

chroot-etc/sommelierrc
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+set -e
+
+if which xsetroot > /dev/null 2>&1; then
+  xsetroot -cursor_name left_ptr
+fi
+
+if [ -f ~/.xwlrc ]; then
```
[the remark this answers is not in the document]
kapilhp: commit 41dfd2e7fed7f2e006c7c95d29a5cd0f09e18b90

chroot-bin/runsommelier
dnschneid: That sounds good to me.

targets/sommelier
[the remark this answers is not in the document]
dnschneid: Probably just copy the one-liner here: https://github.com/dnschneid/crouton/blob/08dac97006c8ae268bb92e503edf5710dd03c1db/installer/main.sh#L145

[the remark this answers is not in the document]
kapilhp: Resolved in commit 3c41cad2593f66165e969274123b8493ed5f0ed1.

chroot-bin/runsommelier
kapilhp: Resolved with two separate commits so as to allow alternate solutions with the same renaming of `runsommelier` to `sommelier`:

- commit 98d12b021f778896b8d198887833326fae2ba94e
- commit 7fb307182c49aea6a8adddb09e7106bb0411491e

chroot-bin/runsommelier
```diff
@@ -0,0 +1,55 @@
...
+export DISPLAY SCALE
+export GDK_BACKEND
+export WAYLAND_DISPLAY
+export XDG_RUNTIME_DIR
+
+# Allow user to define their own variables
+# SOMMELIER_ACCELERATORS and GL_DRIVER_PATH
+
+DEFAULT_SOMMELIER_ACCELERATORS="Super_L,<Control>space,<Alt>space,<Alt>bracketleft,<Alt>bracketright,<Alt>minus,<Alt>equal"
+SOMMELIER_ACCELERATORS="${SOMMELIER_ACCELERATORS:=$DEFAULT_SOMMELIER_ACCELERATORS}"
```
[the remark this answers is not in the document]
kapilhp: commit 0b040a4c706d90293497c9011e77fcc41b92ae59

Not sure if the method used to quote is acceptable.

chroot-bin/runsommelier
kapilhp: commit 0b040a4c706d90293497c9011e77fcc41b92ae59

chroot-etc/sommelierrc
kapilhp: commit 41dfd2e7fed7f2e006c7c95d29a5cd0f09e18b90

`sommelierrc` is now pulled from upstream.

kapilhp:
> Should only copy this in if it doesn't exist, I'd think.

commit 0cb4e0b18f7129d8560f17739b40ec70d8df3a10 creates a backup copy of `/etc/crouton/sommelierrc` if it already exists.

cvmiller:
If you need a tester, I am happy to create a new chroot, and give the target 'somm' a try when you think it is ready.

kapilhp:
Just a quick note to say that I have been preoccupied with work, so I haven't managed to work on this for a while. Will try to get back to it soon!

targets/somm
kapilhp: commit 347db0916f0fbe92420a48dddc25d8571b8f41ce

chroot-bin/runsommelier
kapilhp: commit 7fb307182c49aea6a8adddb09e7106bb0411491e

chroot-bin/runsommelier
kapilhp: commit ea8f7f9923780f99ffd4d4004ef5f499bad91920

chroot-bin/runsommelier
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh
+GDK_BACKEND=x11
+CLUTTER_BACKEND=wayland
+XDG_RUNTIME_DIR=/var/run/chrome
+WAYLAND_DISPLAY=wayland-0
+DISPLAY=:0
+SCALE=1
+export GDK_BACKEND CLUTTER_BACKEND XDG_RUNTIME_DIR WAYLAND_DISPLAY DISPLAY SCALE
```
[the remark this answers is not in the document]
kapilhp: commit 86d547d4b8065b36d9f3306ed70066fab31acc92

chroot-bin/runsommelier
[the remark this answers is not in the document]
kapilhp: commit f741f7aa7678cae6775f5a0ebeb657e3901a4978

I am not sure if this is an acceptable solution. However, the logic is that we need the 'Xwayland' server to run our graphical applications. So we wait for it. The WAITFORX variable is configurable.

chroot-bin/runsommelier
```diff
@@ -0,0 +1,41 @@
...
+       --no-exit-with-child \
+       /bin/sh -c "/usr/local/etc/sommelierrc" > /tmp/sommelier.log 2>&1 &
+    sleep 3
+fi
+
+SOMM=$(pidof sommelier 2> /dev/null)
+if [ ! -z "$SOMM" ]; then
+  echo "sommelier process $SOMM is running"
+else
+  echo "sommelier failed to start"
```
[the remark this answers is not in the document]
kapilhp: commit f741f7aa7678cae6775f5a0ebeb657e3901a4978

chroot-bin/runsommelier
```diff
@@ -0,0 +1,41 @@
...
+  exit 1
+fi
+
+if [ "$1" ]; then
```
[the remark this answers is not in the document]
kapilhp: commit 932c64bd7badb30bc17b684be033dcd8514d7660

chroot-bin/runsommelier
```diff
@@ -0,0 +1,41 @@
...
+    exec "$@" >> /tmp/sommelier.log 2>&1
```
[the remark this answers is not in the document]
kapilhp: commit 932c64bd7badb30bc17b684be033dcd8514d7660

I am not sure how we want to log messages from `sommelier` and its child processes. This is a mechanism.

targets/somm
kapilhp: commit fcf9cafa02bacfc316df012609838af9ca82cae1

kapilhp: commit fcf9cafa02bacfc316df012609838af9ca82cae1

targets/somm
kapilhp: commit e9f9710994196c06f21f29bb708ec194df618433

chroot-bin/runsommelier
```diff
@@ -0,0 +1,55 @@
...
+export XDG_RUNTIME_DIR
+
+# Allow user to define their own variables
+# SOMMELIER_ACCELERATORS and GL_DRIVER_PATH
+
+DEFAULT_SOMMELIER_ACCELERATORS="Super_L,<Control>space,<Alt>space,<Alt>bracketleft,<Alt>bracketright,<Alt>minus,<Alt>equal"
+SOMMELIER_ACCELERATORS="${SOMMELIER_ACCELERATORS:=$DEFAULT_SOMMELIER_ACCELERATORS}"
+
+# Find the architecture specific libgl1-mesa-dri library location
+FOUND_DRIVER_PATH="$(grep '\/dri$' /var/lib/dpkg/info/libgl1-mesa-dri*list | head -1 )"
```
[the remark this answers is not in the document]
kapilhp: commit 71e3420337e7bfbba3f40aea181a7de33ee67acd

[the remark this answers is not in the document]
kapilhp: commit 71e3420337e7bfbba3f40aea181a7de33ee67acd

kapilhp:
As far as I can see only unresolved issue is setting things up to start a rooted 'Xwayland'.

I think that a better solution is to have separate script that does that. The idea would be as follows:

1. Use `sommelier` as an alternative to `xiwi`.
2. Directly use some script with `host-wayland` and `Xwayland` as an alternative to starting the X server on `/dev/tty1` or some other.

The issue with (2) that needs to be resolved is that `ash` grabs some keystrokes (like `Search+[`) from `Xwayland` and that may be confusing.

supechicken:
> As far as I can see only unresolved issue is setting things up to start a rooted 'Xwayland'.
> 
> I think that a better solution is to have separate script that does that. The idea would be as follows:
> 
> 1. Use `sommelier` as an alternative to `xiwi`.
> 2. Directly use some script with `host-wayland` and `Xwayland` as an alternative to starting the X server on `/dev/tty1` or some other.
> 
> The issue with (2) that needs to be resolved is that `ash` grabs some keystrokes (like `Search+[`) from `Xwayland` and that may be confusing.

Do you mean running `Xwayland` directly on the ChromeOS exo server?

kapilhp:
This is a bit OT here perhaps. The following command:
```
XDG_RUNTIME_DIR=/run/chrome /usr/bin/xinit /usr/bin/i3 -- /usr/bin/Xwayland :2
```
will start a fullscreen X server in which one can *somewhat* use the `i3` window manager.

This is not problem-free:

1. `ash` grabs some keystrokes from this X server. For example `Search+[` etc.
2. Windows launched from ChromeOS can appear on *top* of this since `ash` thinks of this as a fullscreen window under it.

In any case, this is a 'rooted` X server.

cvmiller:
This is great news. Just curious, does the sommelier target survive a CB sleep/wake cycle? If not, how does one get the rooted window back?

kapilhp:
> This is great news. Just curious, does the sommelier target survive a CB sleep/wake cycle? If not, how does one get the rooted window back?

The windows survive. If they have been iconified, they can be re-opened from the "overview" or using `Alt+Tab`.

Also,  on my chromebook, I see some "penguin" icons associated with the windows. However, this may be because I had once enabled "Linux (beta)" (since disabled).

chroot-bin/runsommelier
```diff
@@ -0,0 +1,41 @@
...
+       --xwayland-path=/usr/bin/Xwayland \
+       --accelerators=$SOMMELIER_ACCELERATORS \
+       --xwayland-gl-driver-path=$GL_DRIVER_PATH \
+       --no-exit-with-child \
+       /bin/sh -c "/usr/local/etc/sommelierrc" > /tmp/sommelier.log 2>&1 &
+    sleep 3
+fi
+
+SOMM=$(pidof sommelier 2> /dev/null)
+if [ ! -z "$SOMM" ]; then
```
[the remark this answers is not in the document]
kapilhp: This part has been removed entirely.

kapilhp:
> Thank you soooo much for your continued work on this!
> 
> All of the X11 methods in crouton allow for starting with a desktop environment -- i.e., they are able to create a root window that the environment can use as its display, and then chain on its own window manager, etc. Is there a way to do this using sommelier?
> 
> If so, then with that implemented, the x11 target should be updated to select sommelier when it's available, instead of xiwi.

The sommelier approach uses the Aura shell directly as its window manager. The Xwayland runs rootless. In some ways this is analogous to xiwi, in that the windows appear alongside ChromeOS. However, in other ways it is different, in that there is no separate window manager. This also means that the windows do not have decorations.

Running Xwayland as a Wayland *window* under Aura shell is *also* possible. However, I was unable to figure out a way to define its dimensions or give it resize handles etc. It just occupied the full screen. As suggested elsewhere if this can be done, it may be a simpler approach than sommelier.

cvmiller:
> > This is great news. Just curious, does the sommelier target survive a CB sleep/wake cycle? If not, how does one get the rooted window back?
> 
> The windows survive. If they have been iconified, they can be re-opened from the "overview" or using `Alt+Tab`.
> 
This is great news. I tested the Linux Beta to see if those Xapp windows survive a sleep/wake, and they do. I wonder what they are doing that is different.

kapilhp:
@cvmiller You said:

> If you need a tester, I am happy to create a new chroot, and give the target 'somm' a try when you think it is ready.

I don't quite know how to get the sommelier version of the installer to download via /usr/local/bin/crouton. (I really should be able to!) However, the following steps worked for me:

1. Run `sudo crouton -h`. This will download the older version of the installer in `/tmp/crouton-installer-cache/crouton`.
2. Replace that with the sommelier version by
```
curl -# -L --retry 2 'https://github.com/kapilhp/crouton/releases/download/1-alpha/crouton' -o /tmp/crouton-installer-cache/crouton
```
3. Now you can test the creation of a `test` chroot with:
```
sudo crouton -r buster -n test -t sommelier
```
4. This will take a while as the build requires a lot of downloads. After this you should be able to run:
```
sudo startsommelier -n test -b xterm
```
to get a floating `xterm` in your new chroot.

Regarding updating an existing chroot, I would make a backup before trying it!

Thanks in advance for any feedback.

cvmiller:
> sudo startsommelier -n test -b xterm

Alas, it was all going well until the last command. the `sudo startsommelier ...` command says it is entering the chroot, but it doesn't actually. Here's my output:
```
$ sudo startsommelier -n test -b xterm
Entering /mnt/stateful_partition/crouton/chroots/test...
chronos@localhost / $ 
chronos@localhost / $ 
chronos@localhost / $ cd
chronos@localhost ~ $ id
uid=1000(chronos) gid=1000(chronos) groups=1000(chronos),7(lp),18(audio),27(video),208(pkcs11),222(input),240(brltty),303(policy-readers),304(fuse-drivefs),415(suzy-q),420(crash-user-access),600(cras),601(wayland),602(arc-bridge),603(arc-camera),608(midis),1001(chronos-access),20162(traced-producer),20164(traced-consumer),665357(android-everybody) context=u:r:cros_browser:s0
```
As you can see I haven't changed to my chroot user (cvmiller) and I am not in the choot. So therefore, the xterm never appears.

kapilhp:
@cvmiller Thanks for trying it out!
If you haven't restarted yet, then have a look at the following:
1. Check `/tmp/crouton-sommelier.log` for error messages.
2. Look at the directory `/tmp/.X11-unix`. Does it contain a socket like `X0`.
3. Run `pidof sommelier.real` and `pidof Xwayland` do they give an output?

My guess is that my sommelier script (which is quite hacky!) is a bit buggy. I expect that (3) should give output, but let's see.

cvmiller:
> 1. Check `/tmp/crouton-sommelier.log` for error messages.

Thanks for the pointer to the log. I see the problem, the log reports:
```
chronos@localhost ~ $ cd /tmp/
chronos@localhost /tmp $ 
chronos@localhost /tmp $ more crouton-sommelier.log 
error: could not open /dev/dri/renderD128 (No such file or directory)
chronos@localhost /tmp $ 
```
Looking at the /dev/dri directory, I see:

```
chronos@localhost /tmp $ ls -l /dev/dri/
total 0
crw-rw----. 1 root video 226,   0 Feb 19 08:51 card0
crw-rw----. 1 root video 226,   1 Feb 19 08:51 card1
crw-rw----. 1 root video 226,  64 Feb 19 08:51 controlD64
crw-rw----. 1 root video 226, 128 Feb 19 08:51 renderD128
crw-rw----. 1 root video 226, 129 Feb 19 08:51 renderD129
lrwxrwxrwx. 1 root root        10 Feb 19 08:51 vgem -> renderD128

```

Not sure why it is complaining that it can't find the device, as I am starting the command with `sudo`

To clarify some of the other questions:
* This is a fresh crouton install with _only_ the sommelier target
* I am running from Crosh/shell, and keep it open for the entire time
* I am running your command `sudo startsommelier -n test -b xterm` from an alias (defined in .bashrc)
* I am running on a Asus C101PA (ARM) CB

kapilhp:
@cvmiller Thanks once again for the quick response!

> error: could not open /dev/dri/renderD128 (No such file or directory)

That is certainly a big problem!

> Looking at the /dev/dri directory, I see:
> 
> ```
> chronos@localhost /tmp $ ls -l /dev/dri/
> total 0
> crw-rw----. 1 root video 226,   0 Feb 19 08:51 card0
> crw-rw----. 1 root video 226,   1 Feb 19 08:51 card1
> crw-rw----. 1 root video 226,  64 Feb 19 08:51 controlD64
> crw-rw----. 1 root video 226, 128 Feb 19 08:51 renderD128
> crw-rw----. 1 root video 226, 129 Feb 19 08:51 renderD129
> lrwxrwxrwx. 1 root root        10 Feb 19 08:51 vgem -> renderD128
> ```

I see far fewer devices on my machine (HPx360 12b):
```
(bullseye)kapil@localhost:~$ ls -l /dev/dri
total 0
crw-rw----. 1 root video 226,   0 Mar  5 21:44 card0
crw-rw----. 1 root video 226, 128 Mar  5 21:44 renderD128
```
However, I did manage to run `sommelier` on my C100PA when it was running ChromeOS.

I don't think that permissions are the problem here. Something has already "blocked" that particular device.

Since the Asus 101 *can* run "Linux (Beta)" are you running it? I haven't tried to run sommelier alongside "Linux (beta)". The latter would run its own sommelier and that might block the framebuffer device.

cvmiller:
> @cvmiller Thanks once again for the quick response!
> 
> > error: could not open /dev/dri/renderD128 (No such file or directory)
> 
> That is certainly a big problem!
> 
> > Looking at the /dev/dri directory, I see:
> > ```
> > chronos@localhost /tmp $ ls -l /dev/dri/
> > total 0
> > crw-rw----. 1 root video 226,   0 Feb 19 08:51 card0
> > crw-rw----. 1 root video 226,   1 Feb 19 08:51 card1
> > crw-rw----. 1 root video 226,  64 Feb 19 08:51 controlD64
> > crw-rw----. 1 root video 226, 128 Feb 19 08:51 renderD128
> > crw-rw----. 1 root video 226, 129 Feb 19 08:51 renderD129
> > lrwxrwxrwx. 1 root root        10 Feb 19 08:51 vgem -> renderD128
> > ```
> 
> I see far fewer devices on my machine (HPx360 12b):
> 
> ```
> (bullseye)kapil@localhost:~$ ls -l /dev/dri
> total 0
> crw-rw----. 1 root video 226,   0 Mar  5 21:44 card0
> crw-rw----. 1 root video 226, 128 Mar  5 21:44 renderD128
> ```
> 
> However, I did manage to run `sommelier` on my C100PA when it was running ChromeOS.
> 
> I don't think that permissions are the problem here. Something has already "blocked" that particular device.
> 
> Since the Asus 101 _can_ run "Linux (Beta)" are you running it? I haven't tried to run sommelier alongside "Linux (beta)". The latter would run its own sommelier and that might block the framebuffer device.
> 
> _Edit_: I am also mystified by `/dev/dri/card1`. Are there two display cards on the C101? Perhaps the script should check for this and try `/dev/renderD129` as well.
> 
> _Edit2_: I just edited the script `/usr/local/bin/sommelier` on my system and removed the entire line that referred to `/dev/dri/renderD128`. Perhaps you can try that as `sommelier.real` may be picking up the first available device correctly.

A key difference, between your /dev/dri output and mine, is that mine is NOT in the choot, since the choot won't start. Mine is just listed from Crosh.

I have enabled Google's Linux Beta, but have shut down the VM (termina) with `vmc stop termina`. Perhaps there is more to it than that.

I suspect that there are 2 video "cards" in the 101, as a projector video out is performed by USB-C and not a dedicated HDMI port. But I could be wrong, about how USB-C video is handled.

I don't have a `sommelier` script in /usr/local/bin. Here's that I have:
```
chronos@localhost / $ cd /usr/local/bin
chronos@localhost /usr/local/bin $ ls      
crash_reporter_wrapper  delete-chroot  mount-chroot    startxfce4
crouton                 edit-chroot    nospaces.sh     startxiwi
croutonFB_restart.sh    enter-chroot   startsommelier  unmount-chroot
```
Did a debug run (with -x) on the startsommelier script, and I see it dies here:
```
...
+ getopts bc:k:n:t:u:FTf OPT
+ shift 3
+ [ 1 = 0 ]
+ eval exec sh -e "$ENTERCHROOT"  -n 'test' -b     exec sommelier "$@"
+ exec sh -e /usr/local/bin/enter-chroot -n test -b exec sommelier xterm
Entering /mnt/stateful_partition/crouton/chroots/test...
```
Is there supposed to be a binary of sommelier in /usr/local/bin ?

kapilhp:
@cvmiller Sorry I was not clear.
> Is there supposed to be a binary of sommelier in /usr/local/bin ?

This is inside your `test` chroot which you can enter with `sudo enter-chroot -n test`.

After that you should find the script `/usr/local/bin/sommelier`.

cvmiller:
Adding from the previous comment. I see _a_ problem. I have entered the chroot -n test, and I don't see the renderD128, and only see renderD129.

```
chronos@localhost /usr/local/bin $ sudo enter-chroot -n test
Entering /mnt/stateful_partition/crouton/chroots/test...
(test)cvmiller@localhost:~$ 
(test)cvmiller@localhost:~$ 
(test)cvmiller@localhost:~$ 
(test)cvmiller@localhost:~$ ls -l /dev/dri
total 0
crw-rw----. 1 root video 226,   1 Feb 19 08:51 card0
crw-rw----. 1 root video 226,  64 Feb 19 08:51 controlD64
crw-rw----. 1 root video 226, 129 Feb 19 08:51 renderD129
lrwxrwxrwx. 1 root root        10 Feb 19 08:51 vgem -> renderD128
```
BTW, I have 2 chroots right now, the test one, and my old one running Xiwi (I don't run them both at the same time during this testing). I note that the Xiwi /dev/dri only lists renderD129 as well (no 128).

Perhaps the architecture is different on the 101, and it needs to use 129 vs 128?

kapilhp:
> Perhaps the architecture is different on the 101, and it needs to use 129 vs 128?

That sounds right. Which is why I think you should try to remove the explicit reference to /dev/dri/renderD128 from the script `/usr/local/bin/sommelier` in the `test` chroot and try again.

cvmiller:
> > Perhaps the architecture is different on the 101, and it needs to use 129 vs 128?
> 
> That sounds right. Which is why I think you should try to remove the explicit reference to /dev/dri/renderD128 from the script `/usr/local/bin/sommelier` in the `test` chroot and try again.

Sorry, I didn't realize the script was in the chroot. 

I have removed the reference from the script and now xterm works!!! That's the good news.

Unfortunately, I lose the xterm upon sleep/wake up, even if I minimize it. I know that Google has figured out how to keep their Xapps running in Linux Beta through a sleep/wake cycle, so there must be a way.

kapilhp:
> Unfortunately, I lose the xterm upon sleep/wake up, even if I minimize it.

1. Is there a hint of why this happens in `/tmp/crouton-sommelier.log`?
2. After sleep/wake-up is `pidof sommelier.real` still giving a running process? What about `pidof Xwayland`?
3. If those two processes are running then it would be mysterious if `xterm` has died.

On my system here, all the windows under sommelier survive after sleep/wake-up. That did not work for `xiwi`, and was my main (only?) motivation to try sommelier!

cvmiller:
> > Unfortunately, I lose the xterm upon sleep/wake up, even if I minimize it.
> 
> 1. Is there a hint of why this happens in `/tmp/crouton-sommelier.log`?
> 2. After sleep/wake-up is `pidof sommelier.real` still giving a running process? What about `pidof Xwayland`?
> 3. If those two processes are running then it would be mysterious if `xterm` has died.
> 
> On my system here, all the windows under sommelier survive after sleep/wake-up. That did not work for `xiwi`, and was my main (only?) motivation to try sommelier!

1. Yes, the log is reporting a broken pipe to wayland (after sleep/wake)
```
chronos@localhost / $ cat  /tmp/crouton-sommelier.log 
_XSERVTransmkdir: Owner of /tmp/.X11-unix should be set to root
The XKEYBOARD keymap compiler (xkbcomp) reports:
> Warning:          Unsupported maximum keycode 569, clipping.
>                   X11 cannot support keycodes above 255.
> Error:            Key <CAPS> added to map for multiple modifiers
>                   Using Mod3, ignoring Lock.
Errors from xkbcomp are not fatal to the X server
(EE) failed to read Wayland events: Broken pipe
xterm: fatal IO error 11 (Resource temporarily unavailable) or KillClient on X server ":0"
```
2. Neigther of the processes are running after sleep/wake
```
chronos@localhost / $ pidof sommelier.real
chronos@localhost / $ pidof Xwayland
chronos@localhost / $ 
```

dnschneid:
Keep in mind (and sorry I'm slow in reviewing this round) that the `-b` in `sudo startsommelier -n test -b xterm` tells the script to fork after launching, so your shell will end up back in the crosh shell (not the chroot) as xterm runs in the background.  If you drop the `-b` you should end up with an xterm but no usable shell (since xterm is foregrounded).

cvmiller:
> Keep in mind (and sorry I'm slow in reviewing this round) that the `-b` in `sudo startsommelier -n test -b xterm` tells the script to fork after launching, so your shell will end up back in the crosh shell (not the chroot) as xterm runs in the background. If you drop the `-b` you should end up with an xterm but no usable shell (since xterm is foregrounded).

Thanks, I tried it without the `-b` hoping to see more errors, but I don't. Only the umounting of the chroot. Still have to go to the log files to see why the xterm goes away after sleep/wake

kapilhp:
> Keep in mind (and sorry I'm slow in reviewing this round) that the `-b` in `sudo startsommelier -n test -b xterm` tells the script to fork after launching, so your shell will end up back in the crosh shell (not the chroot) as xterm runs in the background. If you drop the `-b` you should end up with an xterm but no usable shell (since xterm is foregrounded).

I am a little confused between the roles of the `-b` and the `-f` arguments. One is handled by `enter-chroot` and the other by `xiwi` in `startxiwi`. I suppose I should try to replicate that behaviour.

cvmiller:
> > Yes, the log is reporting a broken pipe to wayland (after sleep/wake)
> 
> ```
> host-wayland weston --backend=wayland-backend.so
> ```
> 
> This will create a weston window (which may flash a bit).
> 
> Now, do a sleep/wake-up cycle. If the weston window survives, then the problem is with `sommelier`.

Good news, the weston window survives a sleep/wake cycle on my Asus C101PA. 

The terminal output is kind of long. Let me know if you want to see it.

supechicken:
About the sleep/wake issue, it seems like `sommelier` terminated after received a `SIGHUP` signal when laptop lid is closed, maybe we need to wrap `sommelier` with `nohup` to tell sommelier ignore that signal and detach it from nassh (the HTML-based terminal Chrome OS using) using `setsid`.

supechicken:
Also, seems @kapilhp is no longer working on this PR, @dnschneid can I open a new PR about this? (btw I have got the latest commit of sommelier working with some patches, I can add it to the new PR)

TheTechRobo:
Well @supechicken , crouton has entered maintanence only mode, but I might pass up the urge to switch back to Crostini if you can get this working in a fork :D 

I would def appreciate it!

Actually @dnschneid I don't think this woudl make it into the maintainence only aspect, would it?

dnschneid:
Correct, this wouldn't be considered a maintenance PR.  A sommelier-optimized crouton would make for a great fork, though!

kapilhp:
Sorry, folks! I have stopped working on this PR.

I look forward to @cvmiller 's work on this. However, I am currently on Crostini.

PR #4449, opened 2021-03-26
Update chromium docs

Zopolis4:
Makes a minor update to the chromium docs links, changing them to the up to date site.


DennisLfromGA:
Just to add a little context, it appears the document has indeed moved.

crouton-docs

PR #4524, opened 2021-06-20
Update main.sh

lothrond:
Fixes:
    !!!
    !!! PLEASE USE 'dev_boot_altfw' INSTEAD OF 'dev_boot_legacy'
    !!!

### DIFF
--- a/installer/main.sh
+++ b/installer/main.sh
@@ -571,7 +571,7 @@ if [ -z "$DOWNLOADONLY" ] && ! vboot_is_safe; then
     echo "If this is a surprise to you, you should do a full system recovery via USB." 1>&2
     sleep 5
 elif [ -z "$DOWNLOADONLY" ] && \
-    boot="`crossystem dev_boot_usb dev_boot_legacy dev_boot_signed_only`"; then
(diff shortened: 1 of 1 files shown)


lothrond:
@googlebot I signed it!

dnschneid:
The `dev_boot_altfw` variable was only added [very recently](https://chromium.googlesource.com/chromiumos/platform/vboot_reference/+/5b1e2e75b48db4f056dc952616c9a9fecf3535c1%5E%21/utility/crossystem.c), so this code will break most platforms (including old platforms that no longer get updates but we still support according to [CROS_MIN_VERS](https://github.com/dnschneid/crouton/blob/master/build/wrapper.sh#L42)).

This PR (thanks, by the way!) unfortunately will need to detect which is the correct variable to use, and then (preferably) use that.  Something like:

```
altfw=dev_boot_altfw
if crossystem "$altfw" 2>&1 >&- | grep -q Invalid; then
    altfw=dev_boot_legacy
fi
```

lothrond:
Yes. I will get busy on this.
(No worries.)

dnschneid (approved):

PR #4527, opened 2021-06-27
Cause crouton to fail if not executed with `sudo` or similar

saltedcoffii:
Crouton, when not executed as root, fails, because it cannot use `mount --bind`. This prevents that error and causes crouton to only run as root.


dnschneid:
It's possible to run crouton without root to download bootstrap tarballs.  Rather than preventing crouton from being run not as root, maybe just supplement the "Unable to change permissions" message with "you may need to run with sudo".

I think that logic could also be improved to detect the case where you run it on a machine without the noexec restrictions, so it doesn't even try to remount the cache dir as exec...

Also, please be careful not to remove the null character towards the end of the file.  It's a hack to fix the MIME type (to octet_stream), although at the moment I forget what part of the download/install chain breaks without it.

saltedcoffii:
I think most of this is over my head. I'll do some more passive research in my freetime and maybe come back to this project in a bit.

PR #4540, opened 2021-07-08
Wrap uname to enable glibc installs for kernel revisions >= 255

### LINKED ISSUE #4429: crouton on x86_64 *and* arm, kernel version issue with libc updating in chroot
Hello!

I'm having an issue updating my crouton installs.

They fail at updating libc with this error:

ERROR: Your kernel version indicates a revision number
of 255 or greater.  Glibc has a number of built in
assumptions that this revision number is less than 255.
If you\'ve built your own kernel, please make sure that any
custom version numbers are appended to the upstream

I realize this is _not_ a crouton script issue, but an issue with the chrome os version number.

Does anyone know of a way to work around this?

Been stuck with this issue for a while, so I'd thought I'd ask here.

Any help appreciated!

uname -a output:

Linux localhost 4.4.257-19930-ga3a8dfd6a994 #1 SMP PREEMPT Wed Mar 3 21:02:41 PST 2021 aarch64 GNU/Linux

Thanks!

kernel number with a dash or some other delimiter.

```
Please paste the output of the following command here: sudo edit-chroot -all
```

dnschneid:
Fixes #4429, although we're not yet patching uname syscalls.
The key issue is that glibc's version arithmetic can result it in
interpreting revision>=255 as minor+1, which could in turn result in
glibc trying to use kernel features that don't actually exist.
If we find that applications are crashing with SIGSYS then we'll also
need an LD_PRELOAD hack like the one here:
https://gist.github.com/dnschneid/5c88a1802868c1be74d4debb5627c8c7

### DIFF
--- a/installer/ubuntu/prepare
+++ b/installer/ubuntu/prepare
@@ -225,6 +225,24 @@ export DEBIAN_FRONTEND=noninteractive
 echo exit 101 > /usr/sbin/policy-rc.d
 chmod +x /usr/sbin/policy-rc.d
 
+# Debian glibc won't install on kernel revisions >=255, so add a wrapper script
+# to clamp the revision number. Without also patching the uname syscalls,
+# glibc's version math will overflow, resulting in an increased minor version
+# number and possibly in glibc trying to use unavailable syscalls or features.
+if [ ! -e /bin/uname.bin -a ! -e /usr/bin/uname.bin ]; then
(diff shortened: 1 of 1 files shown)


smibarber (approved):
gross, but probably effective and better IMO than LD_PRELOADing if it can be avoided.

PR #4542, opened 2021-07-10
Remove kde-runtime for bullseye

furlongm:


furlongm:
```
Installing target x11...
Installing target kde...
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
E: Unable to locate package kde-runtime
Unmounting /mnt/stateful_partition/crouton/chroots/bullseye...
```

dnschneid (changes requested):
Looks like it still exists in ubuntu and most debian releases, so we'll need to make the installation of this package conditional based on the release.

PR #4608, opened 2021-09-20
Update enter-chroot

ty1118:


ty1118:
In the chroot, many operations cannot be completed because the systemd isn't booted up with the system.
This, with a few tweaks may just fix that.

ty1118:
@googlebot I signed it!

PR #4610, opened 2021-09-20
systemd support with -s option

ty1118:
In the chroot, systemd is not enable rendering certain things useless to the user or certain operations that cannot be done by the user. Please take this into consideration and help me to make crouton even better than you have already done so monumentally already.

### DIFF
--- a/host-bin/enter-chroot
+++ b/host-bin/enter-chroot
@@ -16,6 +16,7 @@ TARGET=''
 USERNAME='1000'
 TMPXMETHOD=''
 NOLOGIN=''
+SYSTEMD=''
 SETUPSCRIPT='/prepare.sh'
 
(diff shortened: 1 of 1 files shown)


ty1118:
@googlebot I signed it!

dnschneid:
This looks to be a dupe of #2474?  At a minimum, if the original author isn't available to continue working on the patch, your PR should add commits on top of their branch, rather than claiming their work as your own.

There are outstanding issues in the first post on #2474 that this PR doesn't seem to address yet.

TomTravis:
can you make this one fresh or close it
i have tried to transfer systemD discussions Github will not let me do this
alll of the issue systemd and no project tasks and some are issues
namespace is 
GC
Video Network Procs

i have contacted google and also signed the agreement if it still active

PR #4669, opened 2022-01-15
Remove outdated package

theblobscp:
The kde-runtime package no longer exists on newer verisons (I tested it on debian 11, not sure about the others). This pr updates the kde install so it doesn't break anymore.


theblobscp:
Just realized this is a duplicate

PR #4796, opened 2022-08-24
Kali rolling needed updated bootstrap

james3212:


dnschneid:
Thanks for submitting a patch!  I don't fully understand it though; MIRROR2 is supposed to be for security updates, but it's now a repo just for debootstrap?  Is it even a valid repo?

If you just need a newer version of debootstrap, that's configured [in this file](https://github.com/dnschneid/crouton/blob/master/installer/ubuntu/bootstrap#L16).

PR #4797, opened 2022-08-24
Update bootstrap

james3212:


dnschneid:
This definitely doesn't make sense.  Just bump the version of debootstrap pulled from the debian server, then please make and actually test that it works before submitting another patch.

PR #4856, opened 2022-11-03
Update xorg -maintenance-

TomTravis:
remove $backport , comment out 2022 for bookworm,sid, kali-rolling and jammy

### DIFF
--- a/targets/xorg
+++ b/targets/xorg
@@ -25,26 +25,25 @@ if [ ! -f "/sys/class/tty/tty0/active" ]; then
 fi
 
 # Pull in backported Xorg when possible on precise to support newer hardware
-backport=''
-if release -eq precise; then
-    if [ "${ARCH#arm}" = "$ARCH" ]; then
-        # Note: lts-trusty mesa requires kernel version >=3.6
-        if ! uname -r | grep -q "^3.4."; then
-            backport='-lts-trusty'
-        fi
-    else
-        # ARM only offers quantal backport at the moment
-        backport='-lts-quantal'
-    fi
-fi
-
+#backport=''
+#if release -eq precise; then
+#   if [ "${ARCH#arm}" = "$ARCH" ]; then
+#        # Note: lts-trusty mesa requires kernel version >=3.6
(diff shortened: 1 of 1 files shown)


targets/xorg
```diff
@@ -189,18 +188,18 @@ else
...
-        xterm x11-common xinput
-else
-    install xorg $installvideodrivers -- xserver-xorg-video-all$backport \
+#     install_dummy xorg -- xserver-xorg$backport libgl1-mesa-glx$backport \
+#         libgl1-mesa-dri$backport libglu1-mesa xfonts-base x11-apps \
+#         x11-session-utils x11-utils x11-xfs-utils x11-xkb-utils \
+#         x11-xserver-utils xauth xinit xfonts-utils xkb-data xorg-docs-core \
+#         xterm x11-common xinput
+# else
+    install xorg $installvideodrivers \
```
dnschneid, line 200: the backports logic only affected Ubuntu 12.04.  I guess we could remove it since that release is long gone.

The only thing that actually affects the stuff you're installing is this change here, which makes it so that now xorg can bring in all of the video drivers, even the ones we don't want.

I think what you actually want to do is determine the video driver you need for your system and add to the logic that sets `installvideodrivers` up above (around line 80).

TomTravis:
it appear to me  the video is installed per -r and Kali has been working more then Debian that is for sure

TomTravis:
botth kali-rolling and bookworm install nivdia kms patches this take 30min
Setting up xserver-xorg (1:7.7+23) ...
Setting up xorg (1:7.7+23) ...
Setting up nvidia-tesla-470-kernel-dkms (470.141.03-3) ...
Loading new nvidia-tesla-470-470.141.03 DKMS files...
It is likely that 4.19.254-13653-ga5ef616743fe belongs to a chroot's host
Building for 4.19.255-13653-ga5ef616743fe and 6.0.0-2-arm64
Module build for kernel 4.19.255-13653-ga5ef616743fe was skipped since the
kernel headers for this kernel do not seem to be installed.
Building initial module for 6.0.0-2-arm64

PR #4861, opened 2022-11-04
Kali systemd

TomTravis:
update enter-chroot
Create new file systemD to manage PTS mounts at startup


TomTravis:
https://github.com/dnschneid/crouton/pull/2474
this was 2016
all kernel now support systemD cgroug control "namespace is dead"

the install of systemd now has the need for cgroups 
this is easy to be done copy and adjust

TomTravis:
namespace is meaningless
sysVinit is not working yet systemd is not far behind 

The current idea is to operate in system space via enter-chroot and status process from the host until all is well

TomTravis:
https://github.com/dnschneid/crouton/issues/4869

dnschneid:
Sorry, crouton is maintenance-only and will not be supporting systemd.

TomTravis:
yes how about namespace
this is why the crouton installs are having issues CG
video network

dnschneid:
Nope, not supporting namespaces either; it's too big a change.

TomTravis:
ok so this is what i have found the later release so far bullleye,,bookworm will work with sudo startxiwi as long as you only install xfce
other wm cause error 2 window starts
the startxiwi cmd kicks off at least 6 process most remain when the app are exits
the systemd repo for chroot will be installed some time this week

something like the croutonxinitwrapper to kick of 3 process a Demons seem much better host cmd via enter-chroot take up to many memory for sleep process doing nothing 6 process the dont die on exit

I could really use you insite on the crouton namespace so do  just blow of systemD as a no support idea

Today is the first time i ran low on memory space and got chromeOS warnings killed about 200 procs and it stopped

PR #4901, opened 2022-11-27
Use ramfs instead of tmpfs

drinkcat:
Seems like Chromium OS >=107 blocked exec on tmpfs.

### DIFF
--- a/installer/main.sh
+++ b/installer/main.sh
@@ -625,9 +625,9 @@ if [ -z "$UPDATE$RESTOREBIN" ] && [ -n "$DOWNLOADONLY" -o -z "$TARBALL" ]; then
     subdir="$RELEASE-$ARCH"
     addtrap "rm -rf --one-file-system '$tmp'"
 
-    # Ensure that the temporary directory has exec+dev, or mount a new tmpfs
+    # Ensure that the temporary directory has exec+dev, or mount a new ramfs
     if [ "$NOEXECTMP" = 'y' ]; then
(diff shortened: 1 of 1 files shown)


dnschneid:
Hard to believe the fix was that simple...

https://chromium-review.googlesource.com/c/chromiumos/third_party/kernel/+/3834992 for reference.  I guess tmpfs is backed by shmem and ramfs isn't...

CroutonIsFun:
> Hard to believe the fix was that simple...
> 
> https://chromium-review.googlesource.com/c/chromiumos/third_party/kernel/+/3834992 for reference. I guess tmpfs is backed by shmem and ramfs isn't...

Thank you!!!

DennisLfromGA:
FYI:

finalize: wipe in ramfs (Iad04ce2e) · Gerrit Code Review
https://chromium-review.googlesource.com/c/chromiumos/platform/factory/+/3935787

drinkcat:
> RAMFS does not swap or is in usr/local as requested by chromeOS many years ago https://chromium.googlesource.com/chromiumos/docs/+/HEAD/security/noexec_shell_scripts.md

This issue on hand here is unrelated to the documentation above: we used to mount our `tmpfs` without `noexec` to avoid that. The problem on hand is a new thing in Chromium OS >=107.

ramfs works, and avoids polluting `/usr/local` in case of problems, so let's stick with that unless we hit other issues. As Dennis pointed out, factory relies on that too now, so we're not alone.

Issue #4923, opened 2023-01-15
Crouton CRAS fails to compile with gcc: error: unrecognized command line option '-Wundef-prefix=HAVE_,CRAS_'


DerenCinar:
```
Please paste the output of the following command here: sudo edit-chroot -all

chronos@localhost /usr/local $ sudo edit-chroot -all
name: xenial
encrypted: no
Entering /mnt/stateful_partition/crouton/chroots/xenial...
/usr/local/bin/crash_reporter_wrapper: 119: exec: /var/run/crw/crw: Permission denied
WARNING: Unable to register core dump handler.
crouton: version git
release: xenial
architecture: amd64
xmethod: xorg
targets: gnome
host: version 15312.0.0 (Official Build) canary-channel hatch 
kernel: Linux localhost 5.15.87-15831-g3504cc8bdad1 #1 SMP PREEMPT Fri Jan 13 04:45:38 PST 2023 x86_64 x86_64 x86_64 GNU/Linux
freon: yes
Unmounting /mnt/stateful_partition/crouton/chroots/xenial...
```

#### Please describe your issue:
I'm doing a fresh installation of crouton. I was able to successfully install Xenial before just fine, but not anymore.

```
$ sudo crouton -r xenial -t gnome

...
patch is already the newest version (2.7.5-1ubuntu0.16.04.2).
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
Compiling CRAS (native)...
gcc: error:  unrecognized command line option '-Wundef-prefix=HAVE_,CRAS_'
```

#### If known, describe the steps to reproduce the issue:

Happens everytime now on ChromeOS Canary channel 111.0.5535.0, with the developer mode turned on.

#### Temporary fix

I was able to fix the issue, at least for now, by changing the **targets/audio** file to remove anything with **undef-prefix**:

```
Line 286:
        sed -i -e 's/\-Wundef\-prefix=HAVE_,CRAS_//g' Makefile.am                                                                                                                                         
        sed -i -e 's/\-Werror=undef-prefix//g' Makefile.am   
```

Nico359:
Where do I find the targets/audio file you are speaking of? I am sorry if I missed something in the text.

NBT22:
It seems as though this is a problem with ChromeOS 111. I had this same problem and was able to fix it by rolling back to the stable channel (109 in my case).

grekpg:
Any new fixes ?

cicciocanestro:
[...]

PR #4957, opened 2023-04-02
Update prepare.sh to resolve gcc error during CRAS compile

mkststrahilov:
Fix for #4923

CRAS fails to compile due to gcc error, a temporary fix was proposed via **targets/audio**  to omit the wrong commands but could not get it to work really... instead implemented the fix directly to convert_automake and it complied without issue.

### DIFF
--- a/installer/prepare.sh
+++ b/installer/prepare.sh
@@ -223,6 +223,11 @@ convert_automake() {
         # Concatenate lines ending in \
         : start; /\\$/{N; b start}
         s/ *\\\n[ \t]*/ /g
+        # Fix for #4932
+        s/\-Wundef\-prefix=HAVE_,CRAS_//g
+        s/-Wundef\-prefix=HAVE_,CRAS_//g
+        s/\-Werror=undef-prefix//g
+        s/-Werror=undef-prefix//g
         # Convert automake to shell
         s/^[^ ]*:/#\0/
         s/^\t/#\0/


drinkcat (commented):
Instead of this, can you edit `Makefile.am` directly?

The idea is to remove this line (https://chromium.googlesource.com/chromiumos/third_party/adhd/+/c2ece09f69565afcbcf511b1693f3fe735bde511/cras/src/Makefile.am#50):
```
COMMON_CPPFLAGS += -Werror=undef-prefix -Wundef-prefix=HAVE_,CRAS_
```

So something like this will do, around here (https://github.com/dnschneid/crouton/blob/master/targets/audio#L165):
```
	sed -i -e 's/^.*error=undef-prefix.*$//'  Makefile.am
```
(it's a bit less strict that necessary but that's ok for now, and we have bigger problems incoming..., see #4958)

drinkcat:
Or as @dnschneid mentioned:
```
	sed -i -e '/error=undef-prefix/d'  Makefile.am
```

mkststrahilov:
> Instead of this, can you edit `Makefile.am` directly?
> 
> The idea is to remove this line (https://chromium.googlesource.com/chromiumos/third_party/adhd/+/c2ece09f69565afcbcf511b1693f3fe735bde511/cras/src/Makefile.am#50):
> 
> ```
> COMMON_CPPFLAGS += -Werror=undef-prefix -Wundef-prefix=HAVE_,CRAS_
> ```
> 
> So something like this will do, around here (https://github.com/dnschneid/crouton/blob/master/targets/audio#L165):
> 
> ```
> 	sed -i -e 's/^.*error=undef-prefix.*$//'  Makefile.am
> ```
> 
> (it's a bit less strict that necessary but that's ok for now, and we have bigger problems incoming..., see #4958)

I can try to play around with it tomorrow and see if it works fine. Just learned about sed yesterday, been reading documentation to figure what to do, cause I wanted to try Crouton out of curiosity. But hey, learning something new ain't bad hehe. 

Just checked #4958... oh boy.

PR #5094, opened 2024-08-26
Extension: Manifest v3

### LINKED ISSUE #5035: crouton extension and manifest v3
Got the notification that manifest v2 extensions might be disabled starting June 2024.  I'd take that date with a grain of salt, but as the crouton extension currently relies on having a background DOM to do copy/paste sync, we'll need to figure out a way to migrate to manifest v3 or defeature clipboard sync entirely.  Hopefully no other features (*looks at xiwi*) are impacted by the switch...

drinkcat:
Should fix #5035.

Manifest V3 introduces a number of changes, in particular, background pages need to be replaced by a service worker, and that service worker does not have access to the DOM of the popup page, so we need to switch everything to message passing.

Also, clipboard handling needs to done in a separate offscreen page now (with similar message passing constraints).

Only tested on a Linux laptop, with a hacked together websocket server. This needs to be tested on a Chromebook. kiwi also totally untested.

### DIFF
--- a/host-ext/crouton/background.html
+++ b/host-ext/crouton/background.html
@@ -7,7 +7,6 @@
   <head>
     <meta http-equiv="Content-Type" content="text/html;charset=utf-8" />
     <title></title>
-    <script src="background.js" type="text/javascript"></script>
   </head>
   <body>
     <div><textarea id="clipboardholder" rows="2" cols="20" /></div>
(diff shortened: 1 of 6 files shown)


paankooza:
my version 105.34

ant573:
Hello

